PR #3, opened 2012-12-19
Trend manual page

rbirnie:


_includes/manual.md
```diff
@@ -40,7 +40,7 @@
         * [4.4.2 Installation Media](#)
         * [4.4.3 Provisioning Templates](#)
         * [4.4.4 Partition Tables](#)
-* ###[5. Advanced Foreman](/manuals/1.0/5_advanced_foreman.html)
+* ###[5. Advanced Foreman](#)
```
skottler: Can we just completely remove this since the page doesn't exist anymore?

skottler:
One minor comment, but other than that this is ready to get merged. Thanks as always, @rbirnie!

_includes/manual.md
rbirnie: That would be fine with me. I just had it as a section heading page, we have similar ones for '4. General Foreman' and '3. Installing Foreman' so I had this one to keep with the pattern.

skottler: What do you think about removing all the links that don't have related content? It will be confusing to users since they will just think it's a dead link.

rbirnie: That sounds good. I'll do that and re-push. Maybe I can just comment these lines out so people who look at source can get an idea what content is still missing but they won't clog up the site.

skottler: That sounds like a good plan @rbirnie.

PR #4, opened 2012-12-19
syntax-highlights

rbirnie:
added some syntax highlights. 

use `sudo easy_install Pygments` to get them working.


skottler:
Thanks so much @rbirnie!

PR #5, opened 2013-01-14
manual internal links fixed and othere ui fixes

abenari:


skottler:
Merged, thanks @abenari!

PR #13, opened 2013-01-29
Database config doc tweaks

richleland:
Updated the database config section to let users know about the foreman-\* packages available for each database type. Also updated the README to note the --auto flag to the jekyll command.

db-docs-update

### DIFF
--- a/_includes/manuals/1.1/3.5.3_database_setup.md
+++ b/_includes/manuals/1.1/3.5.3_database_setup.md
@@ -1,11 +1,27 @@
 
-By default, Foreman will use sqlite3 as a database, its configuration can be found at 
-<pre>config/database.yml</pre>
-By default, the  database can be found at the db subdirectory.
-Foreman is a rails application, therefor, anything that is supported under RAILS (sqlite, mysql, postgresql, ...) can be used.
-At this time, Oracle DB is known to not work. Patches are welcome!
+Foreman is a rails application. Therefore, anything that is supported under RAILS (sqlite, mysql, postgresql, ...) can be used. See <a href="#3.3InstallFromPackages">3.3 Install From Packages</a> for a list of packages for connecting foreman to the databse of your choice. At this time, Oracle DB is known to not work. Patches are welcome!
+
+The database configuration file can be found at:
+
+<pre>/etc/foreman/config/database.yml</pre>
+
+#### SQLite (default)
+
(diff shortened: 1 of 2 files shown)


ohadlevy:
@richleland awesome ! thanks

_includes/manuals/1.1/3.5.3_database_setup.md
```diff
@@ -1,12 +1,28 @@
...
+production:
+  adapter: sqlite3
+  database: db/production.sqlite3
+  pool: 5
+  timeout: 5000
+{% endhighlight %}
 
 #### MySQL
 
+<div class="alert alert-info">When using MySQL, you can install the foreman-mysql package. See <a href="#3.3InstallFromPackages">3.3 Install From Packages</a>.</div>
```
domcleal, line 24: It'd be nice to imply it adds something, as it sort of implies to me that it replaces the configuration.  Perhaps, "When using MySQL, install the foreman-mysql package for support.  See ..."

_includes/manuals/1.1/3.5.3_database_setup.md
```diff
@@ -1,12 +1,28 @@
...
+
+<pre>/etc/foreman/config/database.yml</pre>
+
+#### SQLite (default)
+
+<div class="alert alert-info">When using SQLite, you can install the foreman-sqlite package. See <a href="#3.3InstallFromPackages">3.3 Install From Packages</a>.</div>
+
+By default, the database will be created in the db subdirectory.
+
+{% highlight ruby %}
```
ohadlevy, line 14: this is actually YAML, I wonder if highlight works with yaml syntax too?

richleland, line 14: Yep YAML works, although it doesn't look much different! :smile: I've updated that and the blocks below to be bash where appropriate. Will submit in another PR.

_includes/manuals/1.1/3.5.3_database_setup.md
richleland, line 24: I wasn't sure if it's required or optional to use the "official" package. If it's required, what about "PostgreSQL support requires the foreman-postgresql package. See..."?

domcleal, line 24: True, it's only really optional when running from source and I'd expect those users to have found out how it works there by now (separate bundler.d files and groups).

PR #17, opened 2013-02-07
Baseurl and release fixes

frimik:
For Amos - prepared code to make it easier for hosting in non-root installations (for example personal clones on github). Also added example release version splash to header. Pretty much the only way I could make it fit without ruining the layout.

I don't necessarily like having the versions hardcoded in that html file like that, but wanted to submit for review anyway.

### DIFF
--- a/_includes/releasesplash.html
+++ b/_includes/releasesplash.html
@@ -1,0 +1,3 @@
+<!-- releasesplash.html -->
+  <a class="btn btn-success btn-small" title="1.1 release notes" href="http://projects.theforeman.org/wiki/foreman/ReleaseNotes#11-stable">Stable: 1.1</a>
+  <a class="btn btn-primary btn-small" title="1.1RC5 release notes" href="http://projects.theforeman.org/wiki/foreman/ReleaseNotes#Release-notes-for-11RC5">Unstable: 1.1RC5</a>
--- a/_config.yml
+++ b/_config.yml
@@ -6,6 +6,7 @@ description: Through deep integration with configuration management, DHCP, DNS, 
 
 production_url : http://theforeman.org
 url: http://theforeman.org
+baseurl: /
 
 safe:        false
 auto:        false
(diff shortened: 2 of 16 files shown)


frimik:
This commit is needed for every person wanting to clone and publish their changes for demonstration - for example if they have a pending pull request - https://github.com/frimik/theforeman.org/commit/4c1011f4fa310e70a8938d9a0d28a19e2951f3dc

I've opened a support request with github because I wanted to know if there was any way to make baseurl dynamic based on some sort of internal environment variable etc in the Github Pages jekyll build stage on github side... waiting to hear back.

frimik:
Github support said: "I don't think there's a way to do that currently, and I'm not really sure how that would work. If you come up with a solution, the project is always accepting issues and pull requests.". So, nothing at the moment to get the baseurl automatically fixed for people... will need to advise them to manually update it themselves if they want to "demo" a fully working foreman website on a non-main gh-pages repo.

_includes/releasesplash.html
```diff
@@ -0,0 +1,3 @@
+<!-- releasesplash.html -->
+  <a class="btn btn-success btn-small" title="1.1 release notes" href="http://projects.theforeman.org/wiki/foreman/ReleaseNotes#11-stable">Stable: 1.1</a>
+  <a class="btn btn-primary btn-small" title="1.1RC5 release notes" href="http://projects.theforeman.org/wiki/foreman/ReleaseNotes#Release-notes-for-11RC5">Unstable: 1.1RC5</a>
```
skottler, line 3: Does it make sense to have the RC listed here since 1.1 is already released?

frimik, line 3: nope! I just left it in for demo purposes for commenting on this pullreq.

skottler:
I've merged this patch. I have a PR that I'll submit today to have the latest stable version defined in the settings file with the wiki link. Thanks @frimik!

PR #18, opened 2013-02-07
Modifications on foreman-release rpms locations.

Iouns:
Modified links to allow cut/paste installation of foreman-release (repository).

Iounes

### DIFF
--- a/_includes/manuals/1.1/3.3.1_rpm_packages.md
+++ b/_includes/manuals/1.1/3.3.1_rpm_packages.md
@@ -12,14 +12,27 @@ We maintain a repository for RHEL and Fedora (and clones) - you may add it via:
 
     <pre>http://fedoraproject.org/wiki/EPEL</pre>
 
-2. Install the foreman-release.rpm for your version of RHEL or Fedora
+2. Install the foreman-release.rpm for your version of RHEL or Fedora.
+
+    x86_64 architecture:
     <pre>
-    http://yum.theforeman.org/releases/1.1/el5/foreman-release.rpm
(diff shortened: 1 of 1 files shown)


_includes/manuals/1.1/3.3.1_rpm_packages.md
```diff
@@ -12,15 +12,28 @@ We maintain a repository for RHEL and Fedora (and clones) - you may add it via:
...
+    i386 architecture:
+    <pre>
+    yum localinstall http://yum.theforeman.org/releases/1.1/el5/i386/foreman-release-1.1stable-1.el5.noarch.rpm
+    yum localinstall http://yum.theforeman.org/releases/1.1/el6/i386/foreman-release-1.1stable-1.el6.noarch.rpm
+    yum localinstall http://yum.theforeman.org/releases/1.1/f16/i386/foreman-release-1.1stable-1.fc16.noarch.rpm
+    yum localinstall http://yum.theforeman.org/releases/1.1/f17/i386/foreman-release-1.1stable-1.fc17.noarch.rpm</pre>
 
 3. Install foreman and other foreman-* packages to add functionality:
     <pre>
+    yum install --exclude=foreman-devel,foreman-test foreman-\*</pre>
```
skottler: Most users don't need to all the different packages we have available. For example, users will only need one of the database packages to get up and running. What do you think about leaving the existing list of available packages without the `yum install` command?

domcleal:
+1 to merge as it'll increase accuracy, but an example of yum install foreman-sqlite or similar would be better than everything.

skottler:
Merging - thanks @louns!

PR #19, opened 2013-02-08
Update installer instructions for foreman-installer

domcleal:


_includes/manuals/1.1/2_quickstart_guide.md
```diff
@@ -1,28 +1,41 @@
...
+wget -q http://deb.theforeman.org/foreman.asc -O- | apt-key add -
+apt-get update && apt-get install foreman-installer
+{% endhighlight %}
+
+## Running the installer
+
+**To install Foreman, the Smart Proxy, Puppet, and a puppet master:**
 {% highlight bash %}
-echo include puppet, puppet::server, foreman, foreman_proxy | puppet apply --modulepath /etc/puppet/modules/common
+echo include foreman_installer | puppet apply --modulepath /usr/share/foreman-installer
```
skottler: This should just be `ruby /usr/share/foreman-installer/generate_answers.rb` since the installer will ask users whether they want to run puppet with the choices they've entered

domcleal: Ah yes, fixed!

skottler:
This looks good overall, but the menu changes.

It should look like:

old

but it looks like this:

new

domcleal:
Well-spotted!  I've fixed up a few headings so the levels should be correct now.  Also fixed a broken RPM link.

skottler:
Merged. Thanks @domcleal!

PR #20, opened 2013-02-15
Fixed links to foreman-dev google group in troubleshooting section

vondeetzen:
Fixed broken link; foreman-dev links in troubleshooting sections where pointing to the foreman-users google group before


skottler:
Looks good. Merging - thanks so much!

PR #23, opened 2013-02-21
Update LDAP menu path

supine:
LDAP config is now nested in Users sub-menu

### DIFF
--- a/manuals/1.0/4.1.1_authentication.md
+++ b/manuals/1.0/4.1.1_authentication.md
@@ -21,7 +21,7 @@ and restart Foreman
 
 ## Setting up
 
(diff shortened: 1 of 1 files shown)


domcleal:
The edits need to be in the `manuals/1.1/4.1.1_authentication.md` file rather than 1.0, as it's a Foreman 1.1 change.  Would you mind updating this?

supine:
I think my confusion stems from manuals/1.1/ not having that file. Seems all components have been moved to _includes/manuals/1.1/

skottler:
I committed this change to the 1.1 manual page and gave you git credit, @supine. Thanks so much!

PR #29, opened 2013-03-01
Fixed some styles that looked odd for release notes and add 1.1 release ...

skottler:
...notes to the site

### DIFF
--- a/_includes/manuals/1.1/1.2_release_notes.md
+++ b/_includes/manuals/1.1/1.2_release_notes.md
@@ -1,4 +1,4 @@
-﻿## Release notes for 1.1-1
+## Release notes for 1.1.1
 
 ### Infrastructure
 * Created tmp/ if it doesn't exist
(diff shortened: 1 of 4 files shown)


domcleal:
The green stable release button could be linked to these release notes now.

skottler:
Good catch - added a new commit with that change.

-Sam

_includes/manuals/1.1/1.2_release_notes.md
```diff
@@ -0,0 +1,400 @@
+﻿## Release notes for 1.1-1
```
domcleal: Double check the original file, there's a strange unicode character in here which prevents the heading from working:

```
+<feff>## Release notes for 1.1-1
```

domcleal:
Generally +1, it's an improvement.  I'm only a bit hesitant because there's so much of it.. what do you think about combining all the RCs together so it's just one set of changes?

_includes/manuals/1.1/1.2_release_notes.md
skottler: I'll fix that.

skottler:
Yeah, unfortunately it means there are a lot of additional items in the left menu. My concern about combining all the sections into a single one is that people who are on RC3, for example, will want to know the changes between their version and 1.1.1 stable. What about putting the release notes for each RC under an additional section called "RC notes" or something so they are accessible still?

domcleal:
Indeed, it's just trying to find a good balance - I think that's a good suggestion.  Would you mind doing that?

skottler:
Updated and ready for a final review @domcleal!

domcleal:
:+1: merge!  I like the CSS change too.

PR #32, opened 2013-03-22
Add deep dive videos to media, separate screencasts and talks

domcleal:
Oddly the videos weren't showing in my browser, but I think that's a local issue rather than anything I could have broken in the HTML.  If you could double check, that'd be appreciated...


skottler:
Thank you @domcleal!

PR #39, opened 2013-04-18
Fix Foreman top left link, was partially hidden by navbar div

domcleal:
CSS noob here, feel free to admonish me and improve it!  Tested in FF 19.0 only.

### DIFF
--- a/static/css/style.css
+++ b/static/css/style.css
@@ -17,6 +17,11 @@ th.underlined {
   padding-top: 10px;
 }
 
+.brand {
+  position: relative;
+  z-index: 100;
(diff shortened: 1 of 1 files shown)


ohadlevy:
@abenari ping ?

abenari:
@domcleal is this in the new top bar? what browser do you use?

domcleal:
@abenari this is the web site, not the app.  It's Firefox 20.0.

abenari:
Couldn't reproduce the issue on my browsers, but the css looks good to me.
:+1:

PR #43, opened 2013-04-30
Adding manual pages, also adding manual installation step

lzap:


domcleal:
I think you also need to copy `manuals/1.1` too, otherwise :+1:

lzap:
Fixed, also changed 1.1 strings in there and added Rakefile.

Rakefile
```diff
@@ -0,0 +1,52 @@
...
+          puts "#{referrer.url} contains the following broken links:"
+          pages.each do |page|
+            puts "  HTTP #{page.code} #{page.url}"
+          end
+        end
+      end
+    end
+
+  rescue LoadError
+    abort 'Install anemone gem: gem install anemone'
```
domcleal: Please add this to the Gemfile.

manuals/1.2/index.md
```diff
@@ -0,0 +1,114 @@
+---
+layout: manual
+title: Foreman 1.2 Manual
+---
+
+# 1. Foreman 1.2 Manual
+{% include manuals/1.2/1.2_architecture.md %}
```
domcleal: Search & replace gone wrong?  Should be 1.1_architecture.

domcleal:
I can't get the new check_links rake task to work - any idea what I'm missing?

<pre>
$ rake check_links
iridium - - [01/May/2013:11:26:16 BST] "GET / HTTP/1.1" 200 4746
- -> /
</pre>

manuals/1.2/index.md
lzap: Oh right, found one more like this one. Good catch.

domcleal:
Thanks @lzap, merged!

PR #48, opened 2013-05-17
Fix db instructions

GregSutcliffe:
Saying only 1.0 is confusing, as it means people run the old command for 1.1

### DIFF
--- a/_includes/manuals/1.1/3.5.1_initial_setup.md
+++ b/_includes/manuals/1.1/3.5.1_initial_setup.md
@@ -1,4 +1,3 @@
-
 #### Configuration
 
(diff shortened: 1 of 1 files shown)


domcleal:
a) could we simply remove this?  The manual's versioned, I think it's safe to lose pre-1.0 instructions now.
b) you'll need to make the same change to 1.2 unfortunately as we've cloned it now
c) it won't merge cleanly

domcleal:
d) bundle exec now shouldn't be used on RPM builds of 1.2 or higher

domcleal:
@GregSutcliffe ping?

GregSutcliffe:
Let me review why I wanted to do this in the first place ;)

GregSutcliffe:
I think I'll close this - let's go down the route of creating that wrapper we were talking about and then producing consistent docs from there....

PR #54, opened 2013-06-04
Updated docs

The-Judge:
The path to the database configfile wasn't correct.


domcleal:
Thanks @The-Judge, merged as 6e7338f.

PR #55, opened 2013-06-07
1.2.0-RC2 release notes

domcleal:
@GregSutcliffe I took the liberty of suggesting that Debian and Ubuntu upgrades would be seamless :)


GregSutcliffe:
@domcleal given that implies my packages are awesome, I'll let that slide ;) Reviewing now.

GregSutcliffe:
@domcleal reads fine to me. Merging.

PR #58, opened 2013-06-11
Change specific foreman-release URLs to point to symlinks

domcleal:
https://github.com/theforeman/foreman-infra/pull/43 now creates these automatically, and I've manually created them for all the existing repos.


GregSutcliffe:
:+1:

PR #63, opened 2013-06-26
Clarify smart-proxy DNS update configuration

spacekpe:
I tried to clarify how option dns_key interacts with GSS-TSIG. Also, the procedure how to configure DNS updates with FreeIPA was modified because the old procedure was insecure.

I'm not native English speaker, so the text will require some corrections.

You can contact me on IRC: pspacek @ #freeipa @ FreeNode


domcleal:
That's great, thanks for the contribution.  Committed with only a couple of grammar tweaks.

PR #66, opened 2013-07-01
Revert directory_watcher to 1.4.x to fix regneration

domcleal:
See http://stackoverflow.com/questions/15591000/jekylls-auto-doesnt-work

This seems to improve or fix it for me, it's been terribly unreliable before now on 1.5.


domcleal:
Fixed via #71 as jekyll now requires this older version.

PR #71, opened 2013-07-10
Bump to Jekyll 1.0

skottler:


domcleal:
Can you update the Rakefile too please so it runs the new command?

I think the config file will need updating too, as running `jekyll serve` says this:

```
   Deprecation: Auto-regeneration can no longer be set from your configuration file(s). Use the --watch/-w command-line option instead.
```

skottler:
All set @domcleal.

domcleal:
@skottler missing the Rakefile change?

skottler:
@domcleal nope, the change doesn't need to be in the Rakefile. It's the removed `auto: true` in `_config.yml`.

domcleal:
@skottler doesn't it need to change from --server etc?

Rakefile
```diff
@@ -7,12 +7,12 @@ end
...
 task :build do
-  sh "jekyll"
+  sh "jekyll build"
 end
 
-desc 'Start server with --auto'
+desc 'Start server'
 task :server => :build do
-  sh "jekyll --server --auto"
+  sh "jekyll server"
```
domcleal: This needs -w or --watch to replace --auto.

Rakefile
```diff
@@ -7,12 +7,12 @@ end
 
 desc 'Build site with Jekyll'
 task :build do
-  sh "jekyll"
+  sh "jekyll build"
 end
 
-desc 'Start server with --auto'
+desc 'Start server'
 task :server => :build do
```
domcleal: `jekyll server` appears to regenerate now on the first run, so this :build dependency should be removed I think, to save it running twice.

domcleal:
Thanks @skottler, merged!

PR #72, opened 2013-07-16
Link parameters docs to puppet tutorial about parameters

kleind:
lzap will know (we discussed this on irc)


lzap:
Thanks for the link!

PR #74, opened 2013-07-29
Use the new wiki URL

skottler:


GregSutcliffe:
looks ok to me. are there other urls that need updating?

skottler:
@GregSutcliffe I grepped through the site and there were 5 URL's that had direct wiki references in them. Each of them worked properly, though.

GregSutcliffe:
:+1: then

PR #77, opened 2013-08-04
update jekyll command

shoekstra:
Updated to start Jekyll without deprecation notices:
  Deprecation: The --server command has been replaced by the 'serve' subcommand.
  Deprecation: The switch '--auto' has been replaced with '--watch'.


skottler:
Thanks @shoekstra!

PR #79, opened 2013-08-24
Add section on compute resources and provider-specific notes

domcleal:
Please carefully fact check what I've written, I'm not confident that it's all correct (especially the table).  Any provider specific notes you can think of that we should add, please note too (CC @jmontleon).

### DIFF
--- a/_includes/manuals/1.2/5.2.1_cr_using.md
+++ b/_includes/manuals/1.2/5.2.1_cr_using.md
@@ -1,0 +1,23 @@
+
+The following steps describe how to configure a compute resource and provision new hosts on it.
+
+1. Ensure the necessary package for the provider (from the above table) is installed, e.g. `yum -y install foreman-ovirt`.  Restart the Foreman application to complete installation.
+
+1. Add a compute resource under *More > Provisioning > Compute Resources > New Compute Resource*.  Select the provider type from the menu and appropriate configuration options will be displayed.  Check the notes sections below for any provider-specific setup instructions.
+
+1. Click the *Test Connection* button after entering the configuration.  If no error is displayed, the test was successful.
+
+1. After saving the compute resource, existing virtual machines can be browsed by clicking on the compute resource and the *Virtual Machines* tab.
+
+1. For providers that use images, click on the compute resource, then the *Images* tab, where known images are listed.  To register images that Foreman can use, click *New Image* and enter the details.
+
+1. To provision a new host on this compute resource, from *Hosts*, click *New Host* and select the compute resource from the *Deploy to* menu.
+
+Also note the following features:
+
+1. When viewing a host, power management controls and the console access button are in the top right hand corner of the page.
+
+1. If a host provisioned on a compute resource is deleted, the VM and associated storage on the compute resource will also be deleted.
+
+1. Users in Foreman can have access restricted to hosts present on certain compute resources.  For more information, see Filtering in [4.1.2 Roles and Permissions](/manuals/1.2/index.html#4.1.2RolesandPermissions).
+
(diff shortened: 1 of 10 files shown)


_includes/manuals/1.2/5.2.1_cr_using.md
```diff
@@ -0,0 +1,23 @@
+
+The following steps describe how to configure a compute resource and provision new hosts on it.
+
+1. Ensure the package for the provider support is installed, e.g. `yum -y install foreman-ovirt`.  Restart the Foreman application to complete installation.
```
skottler: Not every compute resource has an associated subpackage. I think it's fine for now since most people will just install the package when the specific compute resource they care about isn't available in the drop-down list, but this section should probably get expanded out to show the packages that each compute resource needs.

_includes/manuals/1.2/5.2.7_cr_vmware.md
```diff
@@ -0,0 +1,46 @@
+
+* Only VMware clusters using vSphere are supported, not standalone ESX or ESXi servers.
```
skottler, line 2: Is that still true? I know of a few people using just ESXi with Foreman and have heard it works fine.

_includes/manuals/1.2/5.2_compute_resources.md
```diff
@@ -0,0 +1,66 @@
+
+Foreman is able to support creating and managing hosts on a number of virtualization and cloud services - referred to as "compute resources" - as well as bare metal hosts.
```
skottler: Nitpick: `Foreman is able to support creating and managing hosts on a number of virtualization` -> `Foreman supports creating and managing hosts on a number of virtualization`

skottler:
I've added a few notes inline, but overall this looks great.

_includes/manuals/1.2/5.2.1_cr_using.md
domcleal: I list the package (which is often just foreman-compute) in a table, which appears above this block of text, so I've clarified this sentence to refer to that.  Unfortunately, libvirt is the only provider that dynamically appears/disappears from the list, the others are hard coded and you get Fog-related errors for them.

_includes/manuals/1.2/5.2.7_cr_vmware.md
domcleal, line 2: Oh, I'm not sure.  I'm pretty certain the free version of ESXi doesn't work as it's lacking API support, but I thought this issue was a blocker for standalone servers: http://projects.theforeman.org/issues/1945

_includes/manuals/1.2/5.2_compute_resources.md
domcleal: Much better, thanks

_includes/manuals/1.2/5.2.4_cr_openstack.md
```diff
@@ -0,0 +1,5 @@
+
+* Supports OpenStack Nova for creating new compute instances.
+
+A full example for configuring image-based provisioning is given on the Foreman blog, also applicable to OpenStack: [EC2 provisioning using Foreman](http://blog.theforeman.org/2012/05/ec2-provisioning-using-foreman.html).
+
```
GregSutcliffe, line 6: As an aside from EC2, it's worth mentioning that Foreman tries to connect to the public IP of the instance for the moment (to SSH in and run the finish script) so you _must_ select a floating IP group for the new instance - if you leave at none it will fail to build correctly.

It's something I'd like to fix (especially if Foreman is on your Openstack cloud and thus has access to the internal IPs) but we've not got there yet.

domcleal, line 6: Thanks for the feedback, updated.

domcleal:
Merged, thanks for the reviews!

PR #81, opened 2013-08-28
Fixing documentation for creating a hashed root password.

jforeman:
Here's the change to the 1.2 docs discussed in IRC.


domcleal:
Thanks for the correction!

PR #86, opened 2013-09-09
Update the API docs from the 1.2-stable branch

skottler:
@inecas is it new for the `apidoc-onepage.html` and `apidoc-plain.html` to no longer get generated?


domcleal:
What happened to the examples?

skottler:
@domcleal not sure why they didn't get generated properly. Any idea @inecas?

iNecas:
I could not reproduce: having fresh installation on ruby1.9.3, removed Gemfile.lock, bundle install, rake apipie:static gave me all the apidoc files, including the examples.

iNecas:
@skottler have you by any chance update the doc/apipie_examples.yml by running tests with APIPIE_RECORD=examples?

skottler:
@iNecas nope, it's a completely fresh instance with its own DB and no special options.

iNecas:
@skottler what db driver to use? (it's unlikely to have any influence, but I'm quite desperate:)

domcleal:
Replaced by #127.

PR #94, opened 2013-10-03
Change generate_answers to foreman-installer, give some more examples

domcleal:


_includes/manuals/1.3/2.1_quickstart_installation.md
```diff
@@ -12,48 +12,18 @@ yum -y install foreman-installer
 For **Debian variants**, run this (replace 'wheezy' with 'precise' if on Ubuntu 12.04, or 'squeeze' for Debian 6):
 
 {% highlight bash %}
-echo "deb http://deb.theforeman.org/ wheezy stable" > /etc/apt/sources.list.d/foreman.list
+echo "deb http://deb.theforeman.org/ wheezy 1.3" > /etc/apt/sources.list.d/foreman.list
```
GregSutcliffe, line 15: I'm not sure which is better here. stable will always link to the latest release, which would mean less things to update when branching the docs for a new release...

_includes/manuals/1.3/2.1_quickstart_installation.md
```diff
@@ -12,48 +12,18 @@ yum -y install foreman-installer
...
 {% endhighlight %}
 
 #### Running the installer
 
-The procedure is can be either fully interactive, allowing changes to the
-configuration and then the option is given to run Puppet to perform the
-installation, or non-interactive.
-
-To run interactive installer do
+The installation run is non-interactive, but the configuration can be customized by supplying any of the options listed in `foreman-installer --help`, or by running `foreman-installer -i` for interactive mode.  More examples are given in the [Installation Options](/manuals/1.3/index.html#3.2.2InstallerOptions) section.  Adding `-v` will disable the progress bar and display all changes.  To run the installer, execute:
```
GregSutcliffe, line 22: Eww, line breaks! It's markdown after all :P `:set tw=80`

domcleal, line 22: Most of the manual doesn't use breaks...

GregSutcliffe, line 22: I know we don't currently have a lot of line-broken text in the manuals yet, but I'm thinking it makes reviews easier, hence the terse comment above - thoughts?

_includes/manuals/1.3/2.1_quickstart_installation.md
domcleal, line 15: We've moved to specifying version numbers on the RPM side, as people want to install old versions.  I think I wrote this a week or two ago when you were moving to versions too, wasn't sure if the stable symlink was going to stay or not... up to you :)

_includes/manuals/1.3/2.1_quickstart_installation.md
domcleal, line 22: I don't care much, but find editing easier without breaks :)

_includes/manuals/1.3/2.1_quickstart_installation.md
GregSutcliffe, line 15: I think consistency with the RPMs wins here. The symlink will stay though :)

_includes/manuals/1.3/2.1_quickstart_installation.md
GregSutcliffe, line 22: You need a better editor :P If you're using vim, try `gqip` while in a paragraph. Otherwise I guess I'm not _that_ fussed :P

manuals/1.3/quickstart_guide.md
```diff
@@ -24,6 +24,7 @@ Or view the quickstart screencasts:
 ### Additional reading...
 
 * [Manual installation and further reading](/manuals/1.3/index.html)
+* [More installation options and scenarios](/manuals/1.3/index.html#3.2.2InstallerOptions)
```
lzap, line 27: Nice one ++

_includes/manuals/1.3/3.2.2_installer_options.md
```diff
@@ -1,29 +1,33 @@
 
-The installer is a collection of Puppet modules that can be operated using an answers file distributed inside the foreman_installer module.  This is located at `/usr/share/foreman-installer/foreman_installer/answers.yaml` in the foreman-installer packages.
+The installer is a collection of Puppet modules, which have a large number of parameters available to customize the configuration.  Parameters can be set by running `foreman-installer` with arguments, e.g. `--foreman-db-type`, by setting up an answers file or changing settings in interactive mode.
```
ares: Maybe just change order somehow. Puppet parameter default values from manifests are default. Answer file is the first that can modify values. Then you can rewrite it by either --parameter or by interactive mode.

EDIT: speaking of what will users use most I'd emphasize --parameters, then interactive mode and then answer file.

ares:
From my point of view (kafo perspective) it all makes sense so :+1:

_includes/manuals/1.3/3.2.2_installer_options.md
domcleal: Updated, does that address it?

ares: It's not the order I prefer but I'm ok with that, thank you.

EDIT: my bad, it's perfect now

domcleal:
Thanks for the comments all.

PR #101, opened 2013-10-18
Adding nice table for 1.3 installer options

lzap:
:)

### DIFF
--- a/_includes/manuals/1.3/3.2.2_installer_options.md
+++ b/_includes/manuals/1.3/3.2.2_installer_options.md
@@ -17,6 +17,612 @@ The installer also provides a text driven interface to customize configuration p
 foreman-installer -i
 {% endhighlight %}
 
+#### Available options
+
+<!-- Generated with:
+/usr/lib/ruby/gems/*/bin/kafo-export-params -c /usr/share/foreman-installer/config/foreman-installer.yaml -f html
+-->
+<div id="installer-options">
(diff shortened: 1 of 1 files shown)


domcleal:
Perhaps wait for http://projects.theforeman.org/issues/3337 and we can then just document the useful ones?

lzap:
Hmmm, if you say that. But note we have no single page on the internet
with all the parameters. Only the --help screen.

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

domcleal:
Yeah, I agree it'd be good.  It just looks long (at least in Markdown) :)

I wonder when @ares might be done...?

lzap:
Well we have MD table output, but it is quite huge in size (lots of

## columns). This looked the best.

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

domcleal:
Ah yes, it doesn't look nearly so bad when you view the rendered version.  I don't mind doing it as-is, but would suggest once 3337's done that we only publish the useful ones.

lzap:
> Ah yes, it doesn't look nearly so bad when you view.

Oh yes, it looks gorgeous.

> I don't mind doing it as-is.

Thank you! :-)

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

domcleal:
Okay, merged as f9c37f40ca, thanks!

PR #106, opened 2013-10-18
[WORK IN PROGRESS] developer documentation for the v2 api standards

komidore64:
included in this discussion are the actual API v2 standards themselves for Foreman and Katello

### DIFF
--- a/_includes/manuals/1.3/5.1.1_api_rabl_overview.md
+++ b/_includes/manuals/1.3/5.1.1_api_rabl_overview.md
@@ -1,9 +1,5 @@
-Three main views for each object
+There are two views for each object that can be accessed via different actions:
 
-1. `base.json.rabl` (small) - minimal fields such as `id`, `name`
+1. `index` (small) - minimal number of fields for the given object (for example: `id` and `name`).
 
-2. `main.json.rabl` (medium) - "important" fields excluding any associated objects (`has_many`)
-
-3. `show.json.rabl` (large) - all fields on an object including some/all associated objects (`has_many`)
-
-Include timestamps in the `main.json.rabl` and `show.json.rabl` partials. If an object is associated with another object (`belongs_to`), include the `_id` and `_name` references (ex. `environment_id`, `environment_name`).
+2. `show` (large) - full list of fields for the given object including references (`_id` and `_name`) to other objects associated with this one.
--- a/_includes/manuals/1.3/5.1.1_api_rabl_overview.md
+++ b/_includes/manuals/1.3/5.1.1_api_rabl_overview.md
@@ -1,0 +1,9 @@
+Three main views for each object
+
+1. `base.json.rabl` (small) - minimal fields such as `id`, `name`
+
+2. `main.json.rabl` (medium) - "important" fields excluding any associated objects (`has_many`)
+
+3. `show.json.rabl` (large) - all fields on an object including some/all associated objects (`has_many`)
+
+Include timestamps in the `main.json.rabl` and `show.json.rabl` partials. If an object is associated with another object (`belongs_to`), include the `_id` and `_name` references (ex. `environment_id`, `environment_name`).
--- a/_includes/manuals/1.3/5.1.2_api_json_collections.md
+++ b/_includes/manuals/1.3/5.1.2_api_json_collections.md
@@ -14,4 +14,4 @@ The collection response should contain the following metadata:
     * `order` - sorted in ascending or descending order
 * `results`(node-name configurable) - list (or collection) of objects; number of objects returned for a special call is equal to `limit`
 
-The fields included in the `main.json.rabl` view will be included for each object in the collection.
+The fields normally included in the `index` action will be included for each object in the collection.
--- a/_includes/manuals/1.3/5.1.2_api_json_collections.md
+++ b/_includes/manuals/1.3/5.1.2_api_json_collections.md
@@ -1,8 +1,6 @@
 Collection responses are used for `index` actions.
 
-The default root node title is configurable, but defaults to `results`.
-
-The collection response contains metadata:
+The collection response should contain the following metadata:
 
 * `total` - total number of objects without any search parameters
 * `subtotal` - number of objects returned with given search parameters (if there is no search, then `subtotal` would be the same as `total`)
@@ -14,6 +12,6 @@ The collection response contains metadata:
 * `sort`
     * `by` - the field that the collection is sorted by
     * `order` - sorted in ascending or descending order
-* `results` - list (or collection) of objects; number of objects returned for a special call is equal to `limit`
+* `results`(node-name configurable) - list (or collection) of objects; number of objects returned for a special call is equal to `limit`
 
 The fields included in the `main.json.rabl` view will be included for each object in the collection.
--- a/_includes/manuals/1.3/5.1.2_api_json_collections.md
+++ b/_includes/manuals/1.3/5.1.2_api_json_collections.md
@@ -1,0 +1,19 @@
+Collection responses are used for `index` actions.
+
+The default root node title is configurable, but defaults to `results`.
+
+The collection response contains metadata:
+
+* `total` - total number of objects without any search parameters
+* `subtotal` - number of objects returned with given search parameters (if there is no search, then `subtotal` would be the same as `total`)
+* `limit` - number of objects return in this API call (for instance, if there are a total of 27 objects, `per_page` is 20, and you're on `page` 2, then only 7 objects would be returned, so the `limit` would be 7)
+* `offset` - index offset from the first item in the collection
+* `page` - page number
+* `per_page` - objects per page
+* `search` - search term used
+* `sort`
+    * `by` - the field that the collection is sorted by
+    * `order` - sorted in ascending or descending order
+* `results` - list (or collection) of objects; number of objects returned for a special call is equal to `limit`
+
+The fields included in the `main.json.rabl` view will be included for each object in the collection.
--- a/_includes/manuals/1.3/5.1.3_api_json_single_objects.md
+++ b/_includes/manuals/1.3/5.1.3_api_json_single_objects.md
@@ -2,4 +2,4 @@ Single object responses are used for `show` actions.
 
 The object's root node name is the object name (ex. `domain`, `environment`, etc) with no metadata since it's only one object.
 
-Use the `show.json.rabl` for the object's fields.
+Uses the `show` action for the object's fields.
--- a/_includes/manuals/1.3/5.1.3_api_json_single_objects.md
+++ b/_includes/manuals/1.3/5.1.3_api_json_single_objects.md
@@ -1,0 +1,5 @@
+Single object responses are used for `show` actions.
+
+The object's root node name is the object name (ex. `domain`, `environment`, etc) with no metadata since it's only one object.
+
+Use the `show.json.rabl` for the object's fields.
--- a/_includes/manuals/1.3/5.1.4_nested_api_routes.md
+++ b/_includes/manuals/1.3/5.1.4_nested_api_routes.md
@@ -1,5 +1,5 @@
-`GET /api/domains/5/subnets` is an example of a nested route. `subnets` would be a nested resource under the `domain` object of id `5`.
-
 Nested routes exist for all associated objects (`has_many`).
 
 All actions are accessible in nested routes.
+
+An example of a nested route would be `GET /api/domains/5/subnets`. In this case, `/api/domains/5/subnets` would give the `subnets` associated with the `domain` that had an `id=5`, whereas `GET /api/subnets` would return all `subnets`.
--- a/_includes/manuals/1.3/5.1.4_nested_api_routes.md
+++ b/_includes/manuals/1.3/5.1.4_nested_api_routes.md
@@ -1,0 +1,5 @@
+`GET /api/domains/5/subnets` is an example of a nested route. `subnets` would be a nested resource under the `domain` object of id `5`.
+
+Nested routes exist for all associated objects (`has_many`).
+
+All actions are accessible in nested routes.
--- a/_includes/manuals/1.3/5.1.5_versioning.md
+++ b/_includes/manuals/1.3/5.1.5_versioning.md
@@ -1,1 +1,3 @@
 To denote the API version, pass `Accept: version=2,application/json` in header.
+
+There is also discussion of having routes-based versioning (i.e - `/api/v2/domains/5/subnets`) - **TBD**
--- a/_includes/manuals/1.3/5.1.5_versioning.md
+++ b/_includes/manuals/1.3/5.1.5_versioning.md
@@ -1,0 +1,1 @@
+To denote the API version, pass `Accept: version=2,application/json` in header.
--- a/_includes/manuals/1.3/5.1_api.md
+++ b/_includes/manuals/1.3/5.1_api.md
@@ -1,0 +1,1 @@
+Below you will find JSON API response standards for the Foreman v2 API and Katello v2 API.
--- a/_includes/manuals/1.3/5.1_api.md
+++ b/_includes/manuals/1.3/5.1_api.md
@@ -1,1 +1,3 @@
 Below you will find JSON API response standards for the Foreman v2 API and Katello v2 API.
+
+**The Foreman/Katello API Standard is a work-in-progress. Notes and documentation found here are subject to change.**
--- a/manuals/1.3/index.md
+++ b/manuals/1.3/index.md
@@ -111,8 +111,19 @@ version: 1.3
 
 #5. Advanced Foreman
 {%include manuals/1.3/missing.html %}
+
 ## 5.1 API
-{%include manuals/1.3/missing.html %}
+{%include manuals/1.3/5.1_api.md %}
+### 5.1.1 Overview of Rabl File Structure/Setup
+{%include manuals/1.3/5.1.1_api_rabl_overview.md %}
+### 5.1.2 JSON Response for Collections
+{%include manuals/1.3/5.1.2_api_json_collections.md %}
+### 5.1.3 JSON Response for Single Objects
+{%include manuals/1.3/5.1.3_api_json_single_objects.md %}
+### 5.1.4 Nested API routes
+{%include manuals/1.3/5.1.4_nested_api_routes.md %}
+### 5.1.5 Versioning
+{%include manuals/1.3/5.1.5_versioning.md %}
 
 ## 5.2 Compute Resources
 {%include manuals/1.3/5.2_compute_resources.md %}


_includes/manuals/1.3/5.1.2_api_json_collections.md
```diff
@@ -0,0 +1,19 @@
...
+* `subtotal` - number of objects contained within this response
+* `limit` - ???
+* `offset` - index offset from the first item in the collection
+* `page` - page number
+* `per_page` - objects per page
+* `search` - search term used
+* `sort`
+    * `by` - the field that the collection is sorted by
+    * `order` - sorted in ascending or descending order
+* `results` - list of objects
```
komidore64: @isratrade are these descriptions accurate?

_includes/manuals/1.3/5.1.5_versioning.md
```diff
@@ -0,0 +1 @@
+To denote the API version, pass `version=2` in a header.
```
domcleal: The `Accept` header?

domcleal:
Looks good, thanks.  I'd try and steer clear of Rails terms long belongs_to if possible, simply as the audience might not know what they mean.  I'd suggest a simpler phase with the Rails term kept, like "associated object (belongs_to)".

The other small issue is that some of this is only in 1.4.. so we might need to hold back or perhaps just add notes saying that this will change between 1.3/1.4, then remove them in the 1.4 manual.  Else maybe we branch the manual for 1.4 soon.

_includes/manuals/1.3/5.1.5_versioning.md
komidore64: ah, yes. it's `Accept: version=2,application/json`

_includes/manuals/1.3/5.1.2_api_json_collections.md
isratrade: ----- Original Message -----

_includes/manuals/1.3/5.1.2_api_json_collections.md
```diff
@@ -0,0 +1,19 @@
+Collection responses are used for `index` actions.
+
+The default root node title is configurable, but defaults to `results`.
+
+The collection response contains metadata:
```
daviddavis: This sentence is kind of ambiguous. I'd recommend saying something like "can contain", "must contain", or "will automatically contain."

ehelms:
I think providing an example that is used throughout this section of documentation would be helpful to avoid any confusion.

_includes/manuals/1.3/5.1.3_api_json_single_objects.md
```diff
@@ -0,0 +1,5 @@
+Single object responses are used for `show` actions.
+
+The object's root node name is the object name (ex. `domain`, `environment`, etc) with no metadata since it's only one object.
```
ehelms, line 3: Did we agree upon using a root node for single objects? I had thought that we decided that was redundant and were shying away from it.

_includes/manuals/1.3/5.1.4_nested_api_routes.md
```diff
@@ -0,0 +1,5 @@
+`GET /api/domains/5/subnets` is an example of a nested route. `subnets` would be a nested resource under the `domain` object of id `5`.
```
daviddavis: So you give an example but you don't really explain when/how nested routes are to be used. IOW, how is `/subnets` different than `/domains/5/subnets` and when do you use one instead of the other?

isratrade: api/domains/5/subnets are the subnets that are associated with domain that has and id = 5

api/subnets are all the subnets

daviddavis: @isratrade I know. I'm wondering if we should add that to the doc.

_includes/manuals/1.3/5.1.3_api_json_single_objects.md
isratrade, line 3: I thought that we decided on having a root node for single objects.  We can open it up for discussion. I also think a node is good if we want to add any metadata in the future, such a

{
    'status': 200,
    'link': localhost/api/hosts/394,
    'host': {
        "name": "fdev2.sat.lab.tlv.redhat.com",
        "id": 394,
        "mac": 
        ....

```
}
```

}

waldenraines, line 3: I have trouble thinking of any metadata related to an object that wouldn't be in the object itself.

The above example is somewhat contrived, I don't think we'll ever want to pass back the http request status or the url the API consumer just requested.

_includes/manuals/1.3/5.1.2_api_json_collections.md
waldenraines: What about "... contains the following metadata:" as I believe that is what was meant.

_includes/manuals/1.3/5.1.2_api_json_collections.md
```diff
@@ -0,0 +1,19 @@
...
+
+* `total` - total number of objects without any search parameters
```
waldenraines, line 5: I think you could just say "total number of objects".

_includes/manuals/1.3/5.1.2_api_json_collections.md
```diff
@@ -0,0 +1,19 @@
...
+* `subtotal` - number of objects returned with given search parameters (if there is no search, then `subtotal` would be the same as `total`)
+* `limit` - number of objects return in this API call (for instance, if there are a total of 27 objects, `per_page` is 20, and you're on `page` 2, then only 7 objects would be returned, so the `limit` would be 7)
```
waldenraines, line 7: I was under the impression that limit was a user provided value and was a method to limit the number of results returned so if you only wanted to display the first 50 results you could do so.  In the example you provide I would expect limit to be null as it wasn't provided by the API consumer.

_includes/manuals/1.3/5.1.2_api_json_collections.md
```diff
@@ -0,0 +1,19 @@
...
+* `offset` - index offset from the first item in the collection
+* `page` - page number
+* `per_page` - objects per page
```
waldenraines, line 10: Can we use camelCase instead of snake_case?  I know ruby uses snake but think that most APIs use camel.

_includes/manuals/1.3/5.1.1_api_rabl_overview.md
```diff
@@ -0,0 +1,9 @@
+Three main views for each object
+
+1. `base.json.rabl` (small) - minimal fields such as `id`, `name`
+
+2. `main.json.rabl` (medium) - "important" fields excluding any associated objects (`has_many`)
+
+3. `show.json.rabl` (large) - all fields on an object including some/all associated objects (`has_many`)
```
waldenraines: I am still of the mind that we should always return all fields of the object that are not sensitive.  What do we gain by introducing this complexity?

_includes/manuals/1.3/5.1.2_api_json_collections.md
daviddavis: That's fine but I am still not sure if the fields are optional, required, etc which is what I was trying to get clarified.

_includes/manuals/1.3/5.1.2_api_json_collections.md
daviddavis, line 5: The current vocabulary `total number of objects without any search parameters` seems clearer/less ambiguous to me.

_includes/manuals/1.3/5.1.3_api_json_single_objects.md
isratrade, line 3: Not contrived.  The example was taken from API hypermedia examples.  Even though we decided not to implement hypermedia at this time, I am in favor of being able to add it later with not breaking things.

_includes/manuals/1.3/5.1.2_api_json_collections.md
isratrade, line 7: If limit is null, then what attributes shows the they there are 7 objects returned if on page=2, per_page=20 and total is 27?

_includes/manuals/1.3/5.1.3_api_json_single_objects.md
ehelms, line 3: Why the inclusion of status in individual object metadata? I have seen the use of a `link` before, but usually on the objects data and included only when showing a collection. In the above example, if I hit resource, don't I know what API path I used? If there is metadata, and in the case of collections, I wonder if we shouldn't have an explicit node that captures all metadata, i.e. { 'metadata': {} }

isratrade, line 3: status was just an example, not saying that we must include it? Regarding showing the link/path, I agree it's redundant which is my I'm not a big fan of Hyermedia (HATEOS).  I just think the option to add metadata at the root level is important without breaking changes.  If you later add  { 'metadata': {} }, we need to add a root node for the object which will break things if there is no root node now.

daviddavis, line 3: +1 to scoping to the obj's name. If you have to submit information to create, update, etc you have to scope it as Rails will automatically add in metadata (controller, action, etc) to the root node so I think scoping the fields to an object is more consistent.

_includes/manuals/1.3/5.1.2_api_json_collections.md
waldenraines, line 7: `results.length` ?

How do I limit my results to 10 if limit is read-only?

_includes/manuals/1.3/5.1.1_api_rabl_overview.md
isratrade: I think it simplifies things, so that for a child :nodes, rabl views can extend one of 3 views rather extending a view with fields or having to manually write attributes :field1, :field2, etc.  If the 'show' view contains relationships, we wouldn't want to include the full 'show' view in the response.  For example

api/hostgroups/123 may show all the puppet classes for the hostgroup.   However, api/hosts/123 may want to show the hostgroup name only (extend base.json.rabl) or more fields (extend main.json.rabl) but probably wont want to extend show.json.rabl and get all the child node information for the hostgroup, but my api call was for show host.

In the future, I want to have customized/partial responses.  Rather than a having a free-for-all of fields=field1,field2,field3, I thought to start by adding the option to fields=all, fields=main, or fields=small/base/etc.  This should cover most people.

_includes/manuals/1.3/5.1.3_api_json_single_objects.md
waldenraines, line 3: @daviddavis is there no way to do this without the redundant root node?  It's nice that rails offers this but I still think the better structure is not to scope it.  We shouldn't structure our API based on rails niceties.

_includes/manuals/1.3/5.1.1_api_rabl_overview.md
waldenraines: :+1: to customized/partial responses and I could really see the benefit of choosing what is in each view based on speed.

What is the criteria for determining whether or not something is "important" though?  I think this part of the documentation could benefit from some examples.

isratrade: The idea was small (id,name), medium, large(all).  You're right, 'medium' is subject to opinion.  There is no clear answer.  It depends on the object.  For simple objects, medium and large may be the small thing.

ohadlevy: @isratrade, we wanted talked about passing a list of attributes you are
interested in, maybe we should only have

 small, large and customized?

isratrade: @ohadlevy, 'main' (medium) is for the index route with no has_many relationships in the view. 
'show' (large) was for the object and may include has_many relationships.  I think it's simply to have a template for attributes in each :index action

isratrade: @ohadlevy, otherwise, we would need to write in the :index action
attributes :field1, :field2, :field3
or 
extend "object/show" which is not always ideal for :index action if it has 'show' view has relationships.

_includes/manuals/1.3/5.1.3_api_json_single_objects.md
komidore64, line 3: we should have it scoped to the object name. it denotes what is part of the object's information and what isn't. we may not have metadata to start with as of now, but it's thinking ahead.

our JSON requests object name scoped, correct? we should do the same with our responses.

daviddavis, line 3: @waldenraines not sure what you're asking but AFAIK there's no way to get Rails not to set controller, action, etc in the root node of params on requests.

@komidore64 V2 in katello now (and thus V1 since V2 extends V1 for the a lot of its actions), some create/update actions use the scoped object name, some don't.

ehelms, line 3: @daviddavis I _think_ the wrapped parameters in Rails can handle this for you and is something Foreman has enabled but we (Katello) do not. I believe wrap_parameters looks at the DB schema, analyzes the request and creates a wrapped object from them to discern what it thinks are the parameters that match that object. The downside is if there are params that don't correspond to the DB schema although there might be a way around this? It would mean we could avoid the use of a root node on single objects but still have what we expect functionality wise on the Rails side.

In V2, we went down the path of no root node because from our perspective there were two things to consider:

1) It just seemed strange to request an object and get: 

```
{'repository': 
    { 
        id: 1,
        name: 'foo'
    }
}
```

and then to put that resource back to the server. I requested a repository, I know what it is.

2) In our angularjs code, this meant doing strange things like:

```
`$scope.repository.repository.name`
```

Where we are creating a resource called "repository" that stores all attributes or the instance of the object coming from the server. When we put back that object, angular would serialize the JavaScript object and make a PUT back to the server. However, we would have to double namespace so that the request is of the form shown above with the root node. We might could get around this with a transform function.

_includes/manuals/1.3/5.1.2_api_json_collections.md
komidore64, line 10: @isratrade what does foreman use?

waldenraines:
@isratrade 

> If limit is null, then what attributes shows the they there are 7 objects returned if on page=2, per_page=20 and total is 27?

`results.length` ?

How do I limit my results to 10 if limit is read-only?

_includes/manuals/1.3/5.1.3_api_json_single_objects.md
daviddavis, line 3: @ehelms so wrap_parameters would allow you to move the keys but they still wouldn't allow you to use those keys.  But you're right, it could be used if we wanted to create a metadata object on incoming requests in order to match your proposal of sending data via show.

One question I am wondering about though--should we open this discussion to a larger audience? Or maybe take a vote in order to resolve this?

_includes/manuals/1.3/5.1.2_api_json_collections.md
isratrade, line 10: @komidore64, @waldenraines, Foreman currently uses snake_case.  afaik there is no standard.  Usually, Javascript devs go for camelCase and Ruby devs go for snake_case.  I prefer snake_case so that the format the server-side (Ruby) expects in POST/PUT request is the same as what it outputs in GET requests.  I realize that client-side(Javascript) prefers things in camelCase.

waldenraines, line 10: @isratrade I would argue that really only ruby and python devs prefer snake_case and that the rest of the world uses camelCase :)

Figured I wouldn't win this one but worth a shot!  I can appreciate that by using snake_case we minimize server side code and that makes sense.

komidore64, line 10: i think we should use `snake_case`

@waldenraines this is a rails application after all :wink:

_includes/manuals/1.3/5.1.3_api_json_single_objects.md
isratrade, line 3: @daviddavis, I am all for opening the discussion to a larger audience by my take is to stick with the "Rails way".  Yes, it's opinionated, but I think Foreman has benefited from this approach until now.  There is no standard convention for JSON API's which is a whole other discussion, but the Rails core team is trying to implement some conventions through ActiveModelSerializers which has the 'opinion' to have an object root name for each :show json response.

_includes/manuals/1.3/5.1.2_api_json_collections.md
daviddavis, line 10: @komidore64 you could easily argue it's a js application too ;)

isratrade, line 10: @waldenraines, good point :)

_includes/manuals/1.3/5.1.4_nested_api_routes.md
```diff
@@ -0,0 +1,5 @@
+Nested routes exist for all associated objects (`has_many`).
+
+All actions are accessible in nested routes.
+
+An example of a nested route would be `GET /api/domains/5/subnets`. In this case, `/api/domains/5/subnets` would give the `subnets` associated with the `domain` that had an `id=5`, whereas `GET /api/subnets` would return all `subnets`.
```
komidore64, line 5: @daviddavis worded  betterly?

daviddavis, line 5: Yes, also kudos for the collection change ("The collection response should contain...").

_includes/manuals/1.3/5.1.3_api_json_single_objects.md
ehelms, line 3: My last comment in this regard, and then I'll shut up (mostly ;)) is to look at what other large projects in our area scope (e.g. Openstack, OpenShift, oVirt) do in this regard. I think it is more important to be consistent with other projects applications/users may interact with alongside our project than to be consistent with the underlying technologies way of doing things.

komidore64, line 3: @ehelms you just volunteered to figure out what they're doing

waldenraines, line 3: @ehelms :+1:

And if you want to look further at other APIs out there in the wild I would surmise that most of them do not scope to a root node.  

If we must include a root node I think it should always be `response` or `data` so we can easily ignore it client side.

daviddavis, line 3: Openstack - ~~not scoped~~ mostly scoped (http://api.openstack.org/api-ref-image.html)

oVirt - not sure. their xml is scoped and their json looks like it might be. seems to be a lack of doc for this (http://www.ovirt.org/images/e/e3/OVirt-API-CLI-SDK-20111102.pdf)

Openshift - attributes are scoped to a node called 'data' (https://www.openshift.com/sites/default/files/documents/OpenShift_Online-2.0-REST_API_Guide-en-US_0.pdf)

ehelms, line 3: Actually, OpenStack is scoped - for example http://docs.openstack.org/api/openstack-block-storage/2.0/content/Show_Volume.html#d6e523

daviddavis, line 3: Look at the [response for "GET v2/images/​{image_id}" though on the link I posted](http://m.davedroid.com/image/2A0A3e070j3r) though. I'm guessing maybe they have multiple apis?

ehelms, line 3: Yea, looks like the imags API doesn't follow the other ones like block and compute. Maybe it's just that different of a kind of API or they are trying to mirror an already existing image API? If you look at http://api.openstack.org/api-ref-blockstorage.html you will see the same as the link I posted.

isratrade, line 3: @ehelms, I'm not stuck on having a root name.  I could go either way.  The question for me is what gives us more flexibility in the future to support the community?  Could we have metadata on single objects in the future? Unfortunately, there are no JSON standards yet.  I see that github.com's API doesn't have metadata for either collections or single objects.  We could also do this and simplify our responses by eliminating metadata, but we didn't so that the Foreman API can support Katello's nutupane.  The Foreman community could possible make similar request changes to support their installation.  The two approaches that I see for API design are either 1) keep it simple or 2) swiss-army knife.  Which one do we want for Foreman / Katello?

waldenraines, line 3: @isratrade I definitely would prefer simple.

Is the only reason we want metadata for nutupane?  We wouldn't necessarily need metadata to support nutupane.  We could do all that calculation client side if we had the entire list.   I would be fine returning entire lists if the APIs were fast.  The one call I know is slow is the list of systems in katello.  The response can be on the order of tens of seconds for just 50 of them.  I would much rather paginate client side if at all possible.

Another option as far as ignoring root node client side goes, as @daviddavis pointed out to me on IRC, is this:  http://ngmodules.org/modules/angularjs-rails-resource

Which accounts for the fact that "Rails prefers JSON be root wrapped".  However, they likely lag behind AngularJS in updates to their own `$resource` object which may be a pain point in the future.

daviddavis, line 3: If we go the no root element route, I recommend turning this on though so we can use strong params and such in the future:

``` ruby
ActiveSupport.on_load(:action_controller) do
  wrap_parameters format: [:json]
end
```

Also, it looks like they have [support for disabling the root element in ActiveModelSerializers](https://github.com/rails-api/active_model_serializers#disabling-the-root-element).

ehelms, line 3: First, I don't think this is just about what the Katello new UI needs (we just seem to be the only consumer of the API chiming in so far). As @waldenraines pointed out, we are in the business of managing everything from small operations to large enterprise operations. Being able to support large scale data sets is important and I think that metadata/paging/search/sort etc. is essential going forward.

I can live with a root node especially given other big projects appear to use them, I just want to make sure we're all set and happy with whatever we choose because APIs are hard to change.

@waldenraines 1.2 $resource includes request/response transforms that would allow us to account for the root node - http://docs.angularjs.org/api/ngResource.$resource

The only question I have left is: generic root node (e.g. "data") or resource specific (e.g. "repository")?

_includes/manuals/1.3/5.1.2_api_json_collections.md
```diff
@@ -0,0 +1,17 @@
+Collection responses are used for `index` actions.
+
+The collection response should contain the following metadata:
+
+* `total` - total number of objects without any search parameters
+* `subtotal` - number of objects returned with given search parameters (if there is no search, then `subtotal` would be the same as `total`)
+* `limit` - number of objects return in this API call (for instance, if there are a total of 27 objects, `per_page` is 20, and you're on `page` 2, then only 7 objects would be returned, so the `limit` would be 7)
```
isratrade, line 7: I suggest to remove calculating limit and offset in foreman controllers based on :page and :per_page.  There is no currently no param[:limit] to pass into Foreman.  A user would use param[:per_page] to limit the # of objects returned in the response.  Thus, I would change the defintion of limit to 

```
* `limit` - number of objects returned as defined by limit parameter
```

isratrade, line 7: I didn't mean to imply above that limit and per_page is the same.
If :limit and :offset are always nil for foreman controllers, then it doesn't make sense to include them. 
Likewise, if katello doesn't use :per_page or :page as params in their controllers, it doesn't make sense to include these in katello responses.

ehelms, line 7: The naming may be different (right now), but they are the same conceptually from what it sounds like. We should just converge on a single convention and adjust our code accordingly.

waldenraines, line 7: Say you have 100 results.  What if the user wants only the first 10 total results, not 10 per page, but 10 total in one page?  

Would they pass offset=0, page_size=10 and page=1?

If so, then I think they are the same and we should go with limit because you may not necessarily be paginating.  I think limit as defined above is easily accomplished client side without a metadata value as `results.length`.

ohadlevy, line 7: it should be just page_size = 10.
page=1 is the default, and offset is not required imho.

isratrade, line 7: The param we currently use is `per_page` rather than `page_size`, `offset` is not a param that currently is used by Foreman.
If there are 100 hosts and a user wants the first 10, then it's just api/hosts?per_page=10.
The default :per_page is defined in Settings::General and is 20 unless changed by the admin.

_includes/manuals/1.3/5.1.5_versioning.md
```diff
@@ -0,0 +1 @@
+To denote the API version, pass `Accept: version=2,application/json` in header.
```
ehelms, line 1: One comment around this convention, this prevents certain applications from taking advantage of the API. The biggest being HTML forms, and while AJAX could be used in place of normal form submitting, for file uploads there is no AJAX upload support yet in browsers. See also - http://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api#versioning

ohadlevy, line 1: Not sure if it block ajax... see
http://stackoverflow.com/questions/3258645/pass-request-headers-in-a-jquery-ajax-get-call

ehelms, line 1: I'm saying, if you use regular HTML forms - you cannot target an API that is based upon setting a header for the version. If the API is header based, you would need to use ajax to submit all your forms for you. However, the only way to upload a file currently is through a standard HTTP form. If you want to do "ajax" file uploads, you have to create an iframe, add a form to the iframe, submit the form in the iframe, and then collect data from inspecting the iframe behind the scenes.

_includes/manuals/1.3/5.1.3_api_json_single_objects.md
komidore64, line 3: **EXECUTIVE DECISION**: collections will have a `results` node (configurable) and single objects will have a `result` object name space (root node) (also configurable).

edit: changed the nodes from being non-configurable to configurable

isratrade, line 3: @komidore64, I think configurable is better even though it strays for the idea of "keep it simple".  Customers have their own dev frameworks, whether open-source or custom in-house, with 'opinionated' conventions. The ability to retrieve data in a certain format (i.e. opinionated) is important.  Imagine if Foreman dictated the root_node as 'data' instead of 'results' and it was non-configurable.  There's a lot of unnecessary work to change  Nutupane to comply with something that should be minor.  It's much easier to change a config file or pass in root_name=results in the URL.  It's a similar idea with partial/custom responses where if a user only wants to retrieve the id/name, the response shouldn't include every attribute.

_includes/manuals/1.3/5.1.5_versioning.md
mccun934, line 1: In general I prefer routes based versioning as a user since it is very clear as a caller what I'm using.  Headers always feel so secondary and require non-obvious ways to utilize them to get the correct version.  I really do like the approach outlined in Eric's link.

komidore64, line 1: I was against it at first, but now i think we'd be good supporting both **a)** a named root api route (i.e. - `/api/v2/blah`) or **b)** non-root-named routes with a v2 header of some sort.

_includes/manuals/1.3/5.1.2_api_json_collections.md
komidore64, line 10: **EXECUTIVE DECISION**: we're going to do `snake_case`.

waldenraines, line 10: @komidore64 I already conceded defeat above.

_includes/manuals/1.3/5.1.5_versioning.md
waldenraines, line 1: You already know my opinion on this.  I also agree with the recommendation in @ehelms' link.

_includes/manuals/1.3/5.1.2_api_json_collections.md
komidore64, line 10: just to make it clear :smile:

_includes/manuals/1.3/5.1.2_api_json_collections.md
isratrade, line 7: @ehelms, can we eliminate :limit and :offset and just have :page and :per_page, at least for Foreman controllers.  I don't see a benefit of including :limit, :offset in the output, especially if the parameters can't be used for input.  Likewise, I don't expect Katello to output :page and :per_page if these params are not used as input.  I don't think that every controller in Foreman/Katello needs the exact same metadata.

ehelms, line 7: @isratrade I didn't realize until you said that that we had proposed trying to use both. The use of `page and per_page` just put more requirement on the application programming against the interface to do more calculation (e.g. where am I at in the collection? Oh, I'm at page \* per_page). If we are OK with that as opposed to explicitly stating that to the user where this data is within the collection then works for me. We should NOT try to support two ways of reporting the same data. We should pick whatever we think makes the most sense from an API standpoint for ourselves and for any application's that may use our API and interact with our server. If that requires changes in Katello or Foreman or both, then so be it.

_includes/manuals/1.3/5.1.1_api_rabl_overview.md
komidore64: i suggest we just have a single view for each object/call to start with. then we can figure out what's slow and what isn't, and make changes accordingly.

I would vote only having two views per object. `index` being small-medium and `show` being everything

isratrade: @komidore64, it was probably my mistake to include names of views/partials in this document.  I suggest to remove it.  The purpose of view/partials is to keep things DRY (Don't Report Yourself).  It is not important for API documentation.  Instead of writing `attributes :id, :name` for an object in a bunch of places, it's DRYer to write ```extend 'api/v2/object_name/base'.  However, the JSON response and documentation is the exact same.  Additionally, API is **not** a full exposure of the database.  Not all db fields need to be exposed in the API, and not all names need to match the db names.

komidore64:
please2merge?

this is definitely not the final set-in-stone standard, but it's a great start. we can continue to make additions and flesh out details as questions are raised.

_includes/manuals/1.3/5.1.2_api_json_collections.md
```diff
@@ -0,0 +1,17 @@
...
+* `offset` - index offset from the first item in the collection
+* `page` - page number
+* `per_page` - objects per page
+* `search` - search term used
+* `sort`
+    * `by` - the field that the collection is sorted by
+    * `order` - sorted in ascending or descending order
+* `results`(node-name configurable) - list (or collection) of objects; number of objects returned for a special call is equal to `limit`
+
+The fields included in the `main.json.rabl` view will be included for each object in the collection.
```
domcleal: should this say the "index" or the "show" view, as defined in 5.1.1?

_includes/manuals/1.3/5.1.3_api_json_single_objects.md
```diff
@@ -0,0 +1,5 @@
...
+
+Use the `show.json.rabl` for the object's fields.
```
domcleal: "Uses"?

_includes/manuals/1.3/5.1.2_api_json_collections.md
isratrade: collection is "index" view and "index" view extends the "main" view for each object.

komidore64: @isratrade that will be transparent to whatever is consuming the API though. To them, they'll either pick `index` or `show`.

_includes/manuals/1.3/5.1.3_api_json_single_objects.md
komidore64: ah yes

domcleal:
Merged as df98980.  Thanks @komidore64 et al for fleshing this out :)

PR #114, opened 2013-11-07
Make the current manual/Foreman version abundantly clear

domcleal:


_layouts/manual.html
```diff
@@ -7,6 +8,12 @@
     <div id="content" class="container">
       <div class="row">
         <div id="toc" class="span3">
+          {% if page.version != page.latestversion %}
+          <div class="alert alert-warn">
+            <b>Newer version available</b><br />
+            This manual is for Foreman {{page.version}}, but the <a href="{{ site.baseurl}}manuals/{{page.latestversion}}/index.html">latest version is {{page.latestversion}}</a>.
```
ehelms: Minor inconsistency with a space after `{{` on the `{{ site.baseurl}}`. Not sure if it blows the styling up, but you may want to wrap this in `<p>` tag for semantics and possibly to avoid having to use the `<br />` above.

domcleal: Thanks, fixed!

domcleal:
Merged as 8d8bec1.

PR #119, opened 2013-11-14
api v2 edits in Section 5

isratrade:


manuals/1.3/index.md
```diff
@@ -126,14 +126,14 @@ version: 1.3
 
 ## 5.1 API
 {%include manuals/1.3/5.1_api.md %}
-### 5.1.1 Overview of Object Views
-{%include manuals/1.3/5.1.1_api_views_overview.md %}
-### 5.1.2 Request Examples
+### 5.1.1 CRUD Request Examples
 {%include manuals/1.3/5.1.2_request_examples.md %}
```
domcleal: Can you keep the filenames in sync with the headings please?

_includes/manuals/1.3/5.1.3_api_json_collections.md
```diff
@@ -1,48 +1,41 @@
-Collection responses are used for `index` actions (i.e. - "Give me a collection of these items").
+Collections are a list of objects (i.e. hosts, domains, etc).  The format for a collection JSON response consists of a `results` root node and metadata fields `total, subtotal, page, per_page`. Note: for Katello objects, the metadata includes `limit, offset` instead of `page, per_page`.
```
domcleal: Why are Katello/Foreman different?  I thought we were standardising.

_includes/manuals/1.3/5.1.4_api_json_single_objects.md
```diff
@@ -1,18 +1,13 @@
-Single object responses are used for `show` actions (i.e. - "Show me this particular item").
+Single object JSON responses are used to show a single object.  The object's unique identifier `:id` or `:name` is required in the **GET** request.  The format for a single object JSON response consists of only the object's attributes.  There is **no root node** and **no metadata** by default.  See section 5.1.4 how to change the root name.
+
+Below is an example of the format for a collection JSON response for a single domain: `GET /api/domains/23`
 
     {
-        "result" : {
```
domcleal, line 4: This isn't what was agreed in https://github.com/theforeman/theforeman.org/pull/106#discussion_r7252476

_includes/manuals/1.3/5.1.6_versioning.md
```diff
@@ -1,3 +1,9 @@
-To denote the API version, pass `Accept: version=2,application/json` in header.
+The default API version is v1 for Foreman 1.3.x
 
-There is also discussion of having routes-based versioning (i.e - `/api/v2/domains/5/subnets`) - **TBD**
+There are two ways to specify to use API v2.
+
+1) in **header**, pass `Accept: application/json,version=2`
```
domcleal: Use `#` for numbering

_includes/manuals/1.3/5.1.7_api_customize.md
```diff
@@ -0,0 +1,83 @@
+
+#### Customize Root Node for Collections
+
+The default root node name for collections is `results` but it can be changed.
+
+1) To globally change the root name for all responses to another word, change the following line in `config/rabl_init.rb`.
```
domcleal: We shouldn't recommend editing application source files.  If we want to provide this feature, it must be in a defined configuration file (i.e. Settings model or settings.yaml).

_includes/manuals/1.3/5.1.7_api_customize.md
```diff
@@ -0,0 +1,83 @@
...
+
+2) To change the object's root node name per API request, just pass `object_name=` as a URL parameter. See example below:
+
+    $ curl -k -u admin:changeme -H "Accept: version=2,application/json" https://foreman.example.com/api/domains/23?object_name=record
+
+&nbsp;
+
+#### Customize Partial Response Attributes
+
+Currently, there is no option to change or customize which attributes are returned for collections or single objects. In the future, we plan to implement customized partial responses such as `fields=field1,field2,field3` or `fields=all`. Similarly, there currently is no option to specify child nodes in an API call or to remove child nodes if they is returned by default.
```
domcleal: Can we file + link to a redmine ticket here?

_includes/manuals/1.3/5.1.7_api_customize.md
```diff
@@ -0,0 +1,83 @@
...
+
+&nbsp;
+
+#### Custom Number of Objects in Collection Per Response
+
+Foreman paginates all collections in the JSON response. The number of objects returned per request is defined in Settings > General > entries_per_page.  The default is 20.  Thus, if there are 27 objects in a collection, only 20 will be returned for the default page=1.
```
domcleal: Style nitpick: `*More > Settings > ...*`

_includes/manuals/1.3/5.1.7_api_customize.md
```diff
@@ -0,0 +1,83 @@
...
+
+    $ curl -k -u admin:changeme -H "Accept: version=2,application/json" https://foreman.example.com/api/domains?per_page=1000
+
+This will return all the objects in one request since 27 is less than the `per_page` parameter set to 1000.
+
+&nbsp;
+
+#### Custom Search of Collections Per Response
+
+Foreman uses the gem scoped_search for searching and filtering which allows all query search parameters to be specified in one string.  To filter results of a collection, justs pass `search=` as a URL parameter. See example below:
```
domcleal: s/justs/just/

domcleal:
cc @komidore64 @ehelms @daviddavis and co

_includes/manuals/1.3/5.1.3_api_json_collections.md
isratrade: Katello uses offset and limit for SQL queries since afaik they are searching on non-ActiveRecord objects. Foreman uses the will_paginate gem for AR objects.  I think it's not good to show parameters (limit, offset) in Foreman responses always as 'null' and they can't be used as input parameters. :per_page and :page are both input and output params.

domcleal: Right, ok.  Are per_page and page available on Katello?  If not, the section below in this file should list them as Foreman only.

_includes/manuals/1.3/5.1.7_api_customize.md
isratrade: already has one.  issue #3019

_includes/manuals/1.3/5.1.6_versioning.md
isratrade: `#` causes formatting errors in markdown.  It looks like 1.  2.  makes a numbered list in markdown.

_includes/manuals/1.3/5.1.4_api_json_single_objects.md
isratrade, line 4: @domcleal, yes, it's different from the 'executive decision' that @komidore64 wrote.  I should have noticed and commented earlier.

_includes/manuals/1.3/5.1.3_api_json_collections.md
isratrade: afaik, per_page and page are Foreman only.  @ehelms can you confirm that Katello doesn't use these params in its api?

_includes/manuals/1.3/5.1.4_api_json_single_objects.md
isratrade, line 4: We can wait for confirmation from @ehelms and @waldenraines to have no root node for :show actions before merging.

_includes/manuals/1.3/5.1.6_versioning.md
domcleal: Ah sorry, me muddling syntaxes again :)

_includes/manuals/1.3/5.1.7_api_customize.md
isratrade: @domcleal, I will remove this from the docs in the mean-time and put it back it when it's implemented in a config file.  Entered issue #3647 for this

domcleal: ok, thanks

isratrade:
@ehelms, @komidore64, when pages are renamed, git doesn't do a good job of showing diffs

_includes/manuals/1.3/5.1.1_request_examples.md
```diff
@@ -0,0 +1,74 @@
...
+
+#### Show a List or Collection of Objects
+
+Get of a list of domains: `GET /api/domains`
+
+Send a HTTP **GET** request.  No JSON data hash is required.
+
+    $ curl -k -u admin:changeme -H "Accept: version=2,application/json" https://foreman.example.com/api/domains
+
+This returns a collection JSON response. The format for a collection response is described in Section 5.1.2.
```
ehelms: How about a cross-link for `Section 5.1.2` ?

_includes/manuals/1.3/5.1.1_request_examples.md
```diff
@@ -0,0 +1,74 @@
...
+
+Send a HTTP **GET** request with the object's unique identifer, either `:id` or `:name`.  No JSON data hash is required.
+
+    $ curl -k -u admin:changeme -H "Accept: version=2,application/json" \
+        https://foreman.example.com/api/domains/42
+    # or
+    $ curl -k -u admin:changeme -H "Accept: version=2,application/json" \
+        https://foreman.example.com/api/domains/foo
+
+This returns the single object in JSON format. The format for a single object response is described in Section 5.1.3.
```
ehelms: `a single object`

ehelms: How about a cross-link here as well?

_includes/manuals/1.3/5.1.6_versioning.md
```diff
@@ -1,3 +1,9 @@
-To denote the API version, pass `Accept: version=2,application/json` in header.
+The default API version is v1 for Foreman 1.3.x
 
-There is also discussion of having routes-based versioning (i.e - `/api/v2/domains/5/subnets`) - **TBD**
+There are two ways to specify to use API v2.
+
+1. In **header**, pass `Accept: application/json,version=2`
+
+2. In **URL**, pass /v2/ such as GET /api/v2/hosts
```
waldenraines: :+1:

_includes/manuals/1.3/5.1.6_versioning.md
```diff
@@ -1,3 +1,9 @@
...
+
+Similarly, when API version v2 becomes the default, users will still be able to use v1 by passing `Accept: application/json,version=1` in the header or api/v1/ in the URL.
```
waldenraines: Missing blank line at the end of this file.

_includes/manuals/1.3/5.1.6_versioning.md
isratrade: @waldenraines, there is an open PR theforeman/foreman/pull/1015 to include this option in Foreman

_includes/manuals/1.3/5.1.4_api_customize.md
```diff
@@ -0,0 +1,69 @@
...
+#### Custom Number of Objects in Collection Per Response
+
+Foreman paginates all collections in the JSON response. The number of objects returned per request is defined in *Settings > General > entries_per_page*.  The default is 20.  Thus, if there are 27 objects in a collection, only 20 will be returned for the default page=1.
+
+To view the next page, just pass `page=` as a URL parameter. See example below:
+
+    $ curl -k -u admin:changeme -H "Accept: version=2,application/json" \
+        https://foreman.example.com/api/domains?page=2
+
+The example above will show the remaing 7 objects in our example of 27 objects in the collection.
```
waldenraines: Typo, remaining.

_includes/manuals/1.3/5.1.4_api_customize.md
```diff
@@ -0,0 +1,69 @@
...
+
+To increase or decrease the number of objects per response, just pass `per_page=` as a URL parameter. See example below:
+
+    $ curl -k -u admin:changeme -H "Accept: version=2,application/json" \
+        https://foreman.example.com/api/domains?per_page=1000
+
+This will return all the objects in one request since 27 is less than the `per_page` parameter set to 1000.
```
waldenraines, line 46: Wonder if we should allow `per_page=all` or something more explicit than forcing API consumers to pass `per_page=999999999` to ensure they get all results if they want them.

_includes/manuals/1.3/5.1.1_request_examples.md
```diff
@@ -0,0 +1,74 @@
...
+
+&nbsp;
+
+#### Create an Object
+
+Create a new domain: `POST /api/domains`
+
+Send a HTTP **POST** request with a JSON data hash containing the required fields to create the object. In this example, we're creating a domain.
```
ehelms: Up to now, everything has been from a third-person perspective - could aim to keep that consistency and rephrase as `In this example, a domain is being created.`

domcleal: +1 to third person everywhere if possible

_includes/manuals/1.3/5.1.2_api_json_collections.md
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+The response metadata fields are described below:
+
+* `total` - total number of objects without any search parameters
+* `subtotal` - number of objects returned with given search parameters (if there is no search, then `subtotal` equals `total`)
+* `page` (Foreman only) - page number
+* `per_page` (Foreman only) - maximum number of objects returned per page
+* `limit` - (Katello only) specified number of objects to return in collection response
+* `offset` - (Katello only) number of objects skipped before beginning to return collection.
```
waldenraines, line 50: Are we at an impasse here?  It would be highly beneficial to our API consumers to only have one set of pagination related terminology.

_includes/manuals/1.3/5.1.1_request_examples.md
```diff
@@ -0,0 +1,74 @@
...
+
+Send a HTTP **DELETE** request with the object's unique identifer, either `:id` or `:name`. No JSON data hash is required.
+
+    $ curl -k -u admin:changeme -H "Accept: version=2,application/json" -X DELETE \
+        https://foreman.example.com/api/domains/17
+    # or
+    $ curl -k -u admin:changeme -H "Accept: version=2,application/json" -X DELETE \
+        https://foreman.example.com/api/domains/foo
+
+This returns the deleted object in JSON format. The format for a single object response is described in Section 5.1.3.
```
waldenraines: We return the deleted object?  Why?

Perhaps we should just return an HTTP 204 and an empty response?

_includes/manuals/1.3/5.1.1_request_examples.md
```diff
@@ -0,0 +1,74 @@
...
+
+This returns the newly created object in JSON format. The format for a single object response is described in Section 5.1.3.
+
+&nbsp;
+
+#### Update an Object
+
+Update a domain: `PUT /api/domains/:id` or `PUT /api/domains/:name`
+
+Send a HTTP **PUT** request with the object's unique identifer, either `:id` or `:name`, plus a JSON data hash containing only the data to be updated. In this example, we just want to update the domain name.
```
ehelms: Similarly here, could opt for `In this example, only the domain name is being updated.`

_includes/manuals/1.3/5.1.2_api_json_collections.md
```diff
@@ -0,0 +1,55 @@
+Collections are a list of objects (i.e. hosts, domains, etc).  The format for a collection JSON response consists of a `results` root node and metadata fields `total, subtotal, page, per_page`. Note: for Katello objects, the metadata includes `limit, offset` instead of `page, per_page`.
```
ehelms: Formatting wise, all of the shaded blocks here are a bit much for me. If the intent is to list a set of properties and draw attention to them, I'd just use a bulleted list.

_includes/manuals/1.3/5.1.1_request_examples.md
isratrade: @waldenraines, I'm open to changing it.  Can you enter a RedMine issue and start a discussion on foreman-dev?

_includes/manuals/1.3/5.1.2_api_json_collections.md
isratrade, line 50: @waldenraines, Foreman is mostly querying ActiveRecord models and uses the will_paginate gem which accepts `:per_page` and `:page` as parameters.  From what I understand, Katello choose `:limit` and `:offset` when querying SQL that is not ActiveRecord models.  Does the metadata have to be identical for all Foreman and Katello requests?  I don't think this should be a requirement.  I can't see Katello passing `:per_page` and `:page` to SQL queries and I don't see Foreman modifying the will_pagina gem te to use `:limit` and `:offset`

_includes/manuals/1.3/5.1.2_api_json_collections.md
isratrade: @ehelms, not sure I follow.  There is a bulleted list at the bottom.  Do you suggest to remove the shaded blocks in the paragraph above or the bulleted list below?

_includes/manuals/1.3/5.1.1_request_examples.md
daviddavis: I can see some value in returning the object information. Suppose you delete the object but want to show a message like "System 'ABC' was deleted." It could save you a call to get the system name if you only have its id.

waldenraines: That seems like the client dictating the API design to me.  If you wanted to support the above use case you simply save the object's name before calling delete and display it afterwards.

daviddavis: So the use case was about only having its id before you delete it. I think returning the object information when deleting is a very common paradigm.

_includes/manuals/1.3/5.1.6_versioning.md
isratrade: @waldenraines, why is a blank line required here?

_includes/manuals/1.3/5.1.4_api_customize.md
isratrade, line 46: @waldenraines, good idea.  Can you add a Red Mine ticket?  I wonder if somebody already monkey-patched will_paginate for this.

_includes/manuals/1.3/5.1.1_request_examples.md
waldenraines: Returning an HTTP 200 with the object is consistent with the HTTP REST [spec](http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.7) (albeit a little counterintuitive from my perspective) so disregard my comments above.

_includes/manuals/1.3/5.1.1_request_examples.md
```diff
@@ -0,0 +1,73 @@
...
+
+This returns the newly created object in JSON format. The format for a single object response is described in [Section 5.1.3](manuals/{{page.version}}/index.html#5.1.3JSONResponseFormatforSingleObjects).
+
+&nbsp;
+
+#### Update an Object
+
+Update a domain: `PUT /api/domains/:id` or `PUT /api/domains/:name`
+
+Send a HTTP **PUT** request with the object's unique identifer, either `:id` or `:name`, plus a JSON data hash containing only the data to be updated. In this example, only the domain name is being updated.
```
daviddavis, line 50: PUT isn't intended for partial modifications--instead the object's entire contents should be sent and anything left out is assumed null. I'm wondering if we should either support PUT as in the REST spec or [switch to PATCH](http://tools.ietf.org/html/rfc5789).

_includes/manuals/1.3/5.1.2_api_json_collections.md
ehelms: We should update for consistency on the Katello side.

_includes/manuals/1.3/5.1.3_api_json_single_objects.md
```diff
@@ -0,0 +1,13 @@
+Single object JSON responses are used to show a single object.  The object's unique identifier `:id` or `:name` is required in the **GET** request.  The format for a single object JSON response consists of only the object's attributes.  There is **no root node** and **no metadata** by default.  See [Section 5.1.4](manuals/{{page.version}}/index.html#5.1.4CustomizeJSONResponses) for how to add a root name.
```
ehelms: We may need a caveat here - names are not necessarily unique in Katello. They are unique within an organization.

_includes/manuals/1.3/5.1.3_api_json_single_objects.md
```diff
@@ -0,0 +1,13 @@
...
+
+Below is an example of the format for a collection JSON response for a single domain: `GET /api/domains/qa.lab.example.com`
```
ehelms: Should this be `format for a single object JSON response` ?

_includes/manuals/1.3/5.1.4_api_customize.md
```diff
@@ -0,0 +1,69 @@
+
+#### Customize Root Node for Collections
+
+The default root node name for collections is `results` but it can be changed.
+
+To change the root node name per API request, just pass `root_name=` as a URL parameter. See example below:
```
ehelms: `just` is superfluous and can be removed.

_includes/manuals/1.3/5.1.4_api_customize.md
```diff
@@ -0,0 +1,69 @@
+
+#### Customize Root Node for Collections
+
+The default root node name for collections is `results` but it can be changed.
```
ehelms: Could remove `it` here.

_includes/manuals/1.3/5.1.6_versioning.md
waldenraines: It's not, we just always try to have one in Katello PRs so I was just pointing out in in case y'all had a similar soft requirement.

_includes/manuals/1.3/5.1.4_api_customize.md
```diff
@@ -0,0 +1,69 @@
...
+
+    $ curl -k -u admin:changeme -H "Accept: version=2,application/json" https://foreman.example.com/api/domains?root_name=data
+
+&nbsp;
+
+#### Customize Root Node for Single Object
+
+There is no root node as the default for single object JSON responses, but it can be added.
+
+To change the object's root node name per API request, just pass `object_name=` as a URL parameter. See example below:
```
ehelms: Can remove the `just` here.

_includes/manuals/1.3/5.1.4_api_customize.md
```diff
@@ -0,0 +1,69 @@
...
+
+To change the root node name per API request, just pass `root_name=` as a URL parameter. See example below:
+
+    $ curl -k -u admin:changeme -H "Accept: version=2,application/json" https://foreman.example.com/api/domains?root_name=data
+
+&nbsp;
+
+#### Customize Root Node for Single Object
+
+There is no root node as the default for single object JSON responses, but it can be added.
```
ehelms, line 15: I thought we had agreed per @komidore64 to have a default root node?

_includes/manuals/1.3/5.1.4_api_customize.md
```diff
@@ -0,0 +1,69 @@
...
+
+To change the object's root node name per API request, just pass `object_name=` as a URL parameter. See example below:
+
+    $ curl -k -u admin:changeme -H "Accept: version=2,application/json" https://foreman.example.com/api/domains/23?object_name=record
+
+&nbsp;
+
+#### Customize Partial Response Attributes
+
+Currently, there is no option to change or customize which attributes are returned for collections or single objects. In the future, we plan to implement customized partial responses such as `fields=field1,field2,field3` or `fields=all`. Similarly, there currently is no option to specify child nodes in an API call or to remove child nodes if they is returned by default.
```
ehelms: `Similarly, there currently is no option to specify child nodes in an API call or to remove child nodes if they is returned by default.`  converted to `Similarly, there is currently no option to specify child nodes in an API call or to remove child nodes if they are returned by default.`

ehelms: `In the future, we plan to implement customized partial responses such as`fields=field1,field2,field3`or`fields=all`.` 

converted to

In the future, customized partial responses such as `fields=field1,field2,field3` or `fields=all` may be implemented.

_includes/manuals/1.3/5.1.4_api_customize.md
```diff
@@ -0,0 +1,69 @@
...
+Foreman paginates all collections in the JSON response. The number of objects returned per request is defined in *Settings > General > entries_per_page*.  The default is 20.  Thus, if there are 27 objects in a collection, only 20 will be returned for the default page=1.
+
+To view the next page, just pass `page=` as a URL parameter. See example below:
+
+    $ curl -k -u admin:changeme -H "Accept: version=2,application/json" \
+        https://foreman.example.com/api/domains?page=2
+
+The example above will show the remaining 7 objects in our example of 27 objects in the collection.
+
+To increase or decrease the number of objects per response, just pass `per_page=` as a URL parameter. See example below:
```
ehelms: `just pass` to `pass`

_includes/manuals/1.3/5.1.4_api_customize.md
```diff
@@ -0,0 +1,69 @@
...
+    $ curl -k -u admin:changeme -H "Accept: version=2,application/json" \
+        https://foreman.example.com/api/domains?per_page=1000
+
+This will return all the objects in one request since 27 is less than the `per_page` parameter set to 1000.
+
+&nbsp;
+
+#### Custom Search of Collections Per Response
+
+Foreman uses the gem scoped_search for searching and filtering which allows all query search parameters to be specified in one string.  To filter results of a collection, just pass `search=` as a URL parameter. See example below:
```
ehelms: `just pass` to `pass`

_includes/manuals/1.3/5.1.5_nested_api_routes.md
```diff
@@ -1,5 +1,12 @@
...
-All actions are accessible in nested routes.
+For example, rather then filtering subnets by a specified domain using a search string
+
+    $ GET /api/subnets?search=name%3Dqa.lab.example.com
+
+the alternative nested route below returns the same result as the above.
+
+    $ GET /api/domains/qa.lab.example.com/subnets
+
+All actions will be accessible in nested route as in the main route.
```
ehelms: `the nested route`

_includes/manuals/1.3/5.1.2_api_json_collections.md
waldenraines, line 50: I agree that it's not necessary for all API metadata to be consistent.  However, I believe something as fundamental as paging ought to be consistent.

I'm open to the idea of changing this on the Katello side if possible (although I may be alone in this regard).

_includes/manuals/1.3/5.1.4_api_customize.md
waldenraines, line 46: @isratrade sure:  http://projects.theforeman.org/issues/3652

_includes/manuals/1.3/5.1.6_versioning.md
```diff
@@ -1,3 +1,9 @@
-To denote the API version, pass `Accept: version=2,application/json` in header.
+The default API version is v1 for Foreman 1.3.x
 
-There is also discussion of having routes-based versioning (i.e - `/api/v2/domains/5/subnets`) - **TBD**
+There are two ways to specify to use API v2.
```
ehelms: `There are two ways to specify use of the v2 API.`

_includes/manuals/1.3/5.1_api.md
```diff
@@ -1,3 +1,5 @@
-Below you will find JSON API standards for the Foreman v2 API and Katello v2 API.
+[API v1](/api.html) is the default version for Foreman 1.3.x.
 
-<div class="alert alert-warn">The Foreman/Katello API Standard is a work-in-progress. Notes and documentation found here are subject to change.</div>
+The purpose of this section is to document the JSON API conventions for the Foreman **API v2** and Katello API v2, both of which are **not released**.  To test out new features in **API v2**, see Section 5.1.6, to specify version.
```
ehelms: `This section documents the JSON API....`

_includes/manuals/1.3/5.1.6_versioning.md
waldenraines: `There are two methods of selecting API v2:`

_includes/manuals/1.3/5.1.6_versioning.md
waldenraines: Maybe change to "In the header" and "In the URL".

_includes/manuals/1.3/5.1.2_api_json_collections.md
komidore64, line 50: Katello/Foreman as a whole should either pick `:per_page`/`:page` or `:limit`/`:offset`. Supporting both would be doable, I think it'd be more confusing than anything else.

Personally, `:per_page` and `:page` seems a bit easier to understand.

_includes/manuals/1.3/5.1.4_api_customize.md
komidore64, line 15: that was the consensus: there would always be a root node but the requester can pass a param to set the root node to the desired name.

root node for single objects: `result`
root node for collections: `results`

_includes/manuals/1.3/5.1.3_api_json_single_objects.md
komidore64: perhaps we should just say to use `id`, since that is what's common between Foreman and Katello?

_includes/manuals/1.3/5.1.4_api_customize.md
isratrade, line 15: My preference was for a descriptive root name (domain, host, etc).  @ehelms, @waldenraines preferred no root name but then agreed to have a root name.  I thought it was going to be a descriptive root name (as we have it currently) and could be configured manually to be 'result'.  Apparently, I missed the notification or 'executive decision' where it was decided that 'result' is the default.  Is there an issue with changing this to  no root name as the default?  I checked more sites and this seems to be the standard.

_includes/manuals/1.3/5.1.3_api_json_single_objects.md
isratrade: Also in Foreman, not ever controller takes :id or :name as a unique identifier.  The full docs should specify this for each controller whether :name is acceptable or not. I will put a clarification here in the manual that :name isn't always unique for every controller.

_includes/manuals/1.3/5.1.2_api_json_collections.md
isratrade: @ehelms, not sure if there is any action item for me per your comment above.

_includes/manuals/1.3/5.1.1_request_examples.md
isratrade, line 50: @daviddavis, I think there's a difference between Rails RESTful and pure RESTful.  Afair, PUT and PATCH are synonyms in Rails.  Rails does not set parameters left out to null.

_includes/manuals/1.3/5.1.2_api_json_collections.md
isratrade, line 50: @komidore64, @waldenraines, I prefer to open up this discussion to the foreman-dev and katello-dev communities.  Can you please start a thread.

_includes/manuals/1.3/5.1.6_versioning.md
isratrade: I chose @waldenraines' sentence only because I prefer `API v2` over `v2 API` :-)

_includes/manuals/1.3/5.1.6_versioning.md
isratrade: @waldenraines, I'm not aware of this requirement for a blank line at the end of files, but I added one anyway.  @domcleal do you prefer blank lines?

_includes/manuals/1.3/5.1.1_request_examples.md
waldenraines, line 50: Also, you'll be hard pressed to find any RESTful API out there that 100% follows the spec especially when it comes to @daviddavis point above.

_includes/manuals/1.3/5.1.4_api_customize.md
waldenraines, line 15: @isratrade I fully support that decision :+1:

_includes/manuals/1.3/5.1.1_request_examples.md
daviddavis, line 50: @isratrade I don't think Rails thinks of PUT and PATCH as the same thing. In fact, that's [why they're switching to PATCH in Rails 4](http://weblog.rubyonrails.org/2012/2/25/edge-rails-patch-is-the-new-primary-http-method-for-updates/). 

@waldenraines I'm all for not following the spec if it makes sense or there's a good reason but seeing how PATCH was specifically designed for both full and partial updates, I'm not sure what that is here?

_includes/manuals/1.3/5.1.1_request_examples.md
daviddavis: @waldenraines yea, I don't see how it could hurt.

_includes/manuals/1.3/5.1.1_request_examples.md
waldenraines, line 50: @daviddavis the only reason I could see for using PUT instead of PATCH, regardless of the spec, is because PATCH isn't all that common and there are likely clients out there that don't support it.  That being said, I'm fine with using the more correct PATCH, just wanted to point that out.

isratrade, line 50: @daviddavis, yes I was referring to the Rails 3.x router which routes PUT to the :update action.  In Rails 4, both PUT and PATCH will route to the :update action say it appears that Rails 4 treats them as the same for routing purposes.  Have you checked if Foreman/Katello which uses Rails 3.x routes PATCH to the :update action?

daviddavis, line 50: @isratrade It doesn't AFAIK. I can double check though. I was more thinking long-term that we would change this PATCH when Rails 4 is out. I had hoped short-term though (until 4) we could still reserve PUT for full idempotent updates but given the nature of update_attributes, that won't work so waiting to use PATCH in 4 will have to suffice.

Edit: Just checked and PATCH = 404 for `resources` routes.

isratrade, line 50: @daviddavis, ok thanks for checking.  Se we agreed that we will not include PATCH in the documentation until we upgrade to Rails 4.

isratrade:
@domcleal, @ehelms, @waldenraines, @daviddavis, @komidore64 
I squashed my latest commits.  Any ACK's to merge this?

_includes/manuals/1.3/5.1.1_request_examples.md
daviddavis, line 50: :+1:

waldenraines:
ACK

ehelms:
ACK

isratrade:
@waldenraines, @ehelms, thanks for the ACK.  I don't have merge rights.  Can somebody merge this PR?  @domcleal, @ohadlevy

komidore64:
i'm not cool with this being merged yet
- we still need to figure out `:per_page / :page` vs `:limit / :offset`
- a final decision needs to be made on the root-node issue

_includes/manuals/1.3/5.1.4_api_customize.md
komidore64, line 15: so root-node will be optional for single-object but non-optional in collection responses?

what about json structure sent in requests?

_includes/manuals/1.3/5.1.1_request_examples.md
```diff
@@ -0,0 +1,73 @@
+The follow examples show the basic CRUD operations (Create, Read, Update, Delete) using the JSON API.
```
komidore64: should be

"The following examples"

_includes/manuals/1.3/5.1.1_request_examples.md
```diff
@@ -0,0 +1,73 @@
...
+
+&nbsp;
+
+#### Show a List or Collection of Objects
```
komidore64: `List or Collection` infers that a list is somehow different from a collection

we should either use `collection` or `list`

_includes/manuals/1.3/5.1.1_request_examples.md
```diff
@@ -0,0 +1,73 @@
...
+&nbsp;
+
+#### Create an Object
+
+Create a new domain: `POST /api/domains`
+
+Send a HTTP **POST** request with a JSON data hash containing the required fields to create the object. In this example, a domain is being created.
+
+    $ curl -k -u admin:changeme -H "Accept: version=2,application/json" -X POST \
+        -d { "name":"foo.bar.com","fullname":"foo.bar.com description" } https://foreman.example.com/api/domains
```
komidore64, line 40: tagging onto the root-node discussion:

should we have root-nodes be consistent for requests as well?

``` json
{
  "name": "foo.bar.com",
  "fullname": "foo.bar.com description"
}
```

vs

``` json
{
  "domain": {
    "name": "foo.bar.com",
    "fullname": "foo.bar.com description"
  }
}
```

_includes/manuals/1.3/5.1.2_api_json_collections.md
```diff
@@ -0,0 +1,55 @@
...
+* `subtotal` - number of objects returned with given search parameters (if there is no search, then `subtotal` equals `total`)
+* `page` (Foreman only) - page number
+* `per_page` (Foreman only) - maximum number of objects returned per page
+* `limit` - (Katello only) specified number of objects to return in collection response
+* `offset` - (Katello only) number of objects skipped before beginning to return collection.
+* `search` - search string (based on scoped_scoped syntax)
+* `sort`
+    * `by` - the field that the collection is sorted by
+    * `order` - sort order, either ASC for ascending or DESC for descending
+* `results` - list (or collection) of objects. See [Section 5.1.4](manuals/{{page.version}}/index.html#5.1.4CustomizeJSONResponses) for how to change the root name from 'results' to something else.
```
komidore64: need to pick `list` or `collection`

_includes/manuals/1.3/5.1.4_api_customize.md
```diff
@@ -0,0 +1,68 @@
...
+&nbsp;
+
+#### Customize Root Node for Single Object
+
+There is no root node as the default for single object JSON responses, but it can be added.
+
+To change the object's root node name per API request, pass `object_name=` as a URL parameter. See example below:
+
+    $ curl -k -u admin:changeme -H "Accept: version=2,application/json" \
+        https://foreman.example.com/api/domains/23?object_name=record
```
komidore64, line 20: the param to specify the root-node name for single-objects and collections should be the same

_includes/manuals/1.3/5.1.3_api_json_single_objects.md
```diff
@@ -0,0 +1,12 @@
...
+Below is an example of the format for a single object JSON response: `GET /api/domains/23` or `GET /api/domains/qa.lab.example.com`
+
+    {
+        "id": 23,
+        "name": "qa.lab.example.com",
+        "fullname": "QA",
+        "dns_id": 10,
+        "created_at": "2013-08-13T09:02:31Z",
+        "updated_at": "2013-08-13T09:02:31Z"
+    }
```
komidore64, line 12: if you're going to make root-nodes for single-object responses optional, you should include an example response for both cases

_includes/manuals/1.3/5.1.1_request_examples.md
daviddavis, line 40: +1 to a root node for POST/PUT since Rails populates top level params like `controller`, `action`, `format`, etc and having a root node will work better with new Rails features (strong params, serialization, etc).

_includes/manuals/1.3/5.1.1_request_examples.md
isratrade: @komidore64, I updated this to 'collection', but 'list' also appears elsewhere in the docs.  I don't think we need an iron-clad rule to only use `collection` or only use `list`

_includes/manuals/1.3/5.1.1_request_examples.md
isratrade, line 40: @daviddavis, I was also in favor of using Rails ActiveSerializers conventions which is a root-node name corresponding to the controller name (not generic 'result' for all controllers).  However, Rails doesn't require a root name for POST/PUT/DELETE, so both examples above will work.  Can you explain further what is the real benefit of using a root-node as related to strong params, serialization?

_includes/manuals/1.3/5.1.4_api_customize.md
isratrade, line 20: @komidore64, I thought of this as well, but I I want to make sure there's not a clash (causing undesired results) in the rabl helpers that I extended in rabl_init.rb.  I will check further.  In any case, this can be changed in a later PR and shouldn't be a blocker for merge.

_includes/manuals/1.3/5.1.4_api_customize.md
isratrade, line 15: @komidore64, root-node is non-optional in collection responses because there is meta-data, so a node is required.

isratrade:
I re-committed with grammer / wording changes.  I think this can be merged and we still continue the discussion about `:per_page / :page` vs `:limit / :offset` and root-node vs no root-node for single objects.

_includes/manuals/1.3/5.1.1_request_examples.md
daviddavis, line 40: For strong_parameters, the pattern is [to create a params method and require a certain root node and then permit certain values within that node](http://weblog.rubyonrails.org/2012/3/21/strong-parameters/). That said, we don't have to use strong_parameters and even if we really want to, we could use wrap_parameters to fake it. I'm less knowledgeable about the new active model serializers but it looks like they use root node by default (although this can be configured).

isratrade, line 40: @daviddavis, it seems that Rails creates the named node on PUT even if not explicit sent.
PUT -d {"architecture": {"name": "i386"}} api/architectures/2 
in the log file I see
`Parameters: {"architecture"=>{"name"=>"i386"}, "id"=>"2"}`
PUT -d {"name": "i386"} api/architectures/2
in the log file I see
`Parameters: {"name"=>"i386", "id"=>"2", "architecture"=>{"name"=>"i386"}}`

So, it seems that GET object responding without a node can be the same format as PUT/POST/DELETE without a node
Likewise, GET/PUT/POST/DELETE can all be with formatted with a "named node" as well (not 'result')

daviddavis, line 40: Still seems suboptimal to me especially when we have code like this in Katello:

https://github.com/Katello/katello/blob/master/app/controllers/api/v1/systems_controller.rb#L166

I think I'd prefer to see a root node as compared to having the attributes populated in two places (more so given that it leads to bad API code design).

daviddavis, line 40: @isratrade this doesn't seem to work in Katello:

```
PUT -d {"name": "Test"} /katello/api/systems/5124f431-9ff8-4676-a15b-143e7c6ee350
```

```
Parameters: {"name"=>"Test", "action"=>"update", "controller"=>"api/v1/systems", "id"=>"5124f431-9ff8-4676-a15b-143e7c6ee350"}
```

There's no params[:system]. Do you have wrap_parameters turned on? Regardless though, I wasn't trying to claim that not having a root node wasn't technically feasible.

komidore64:
merge it

isratrade:
@domcleal, can you merge this PR.  I don't have rights

domcleal:
Merged as 42c2aa6, thanks @isratrade and all reviewers!

PR #129, opened 2013-12-17
move the config file copy higher up in the instructions

mccun934:
If you don't have the config file copied before you run
'bundle install' you get an error stating the file can not be
found.  moving this up before the first bundle install will reduce
new developer confusion.


skottler:
Merged as 8ad19e90d06d3c8160a2c66896d98174d40c1b9a. Thanks @mccun934!

PR #130, opened 2013-12-17
Revisited our redhat development deps

lzap:
Revisited our redhat development deps


lzap:
For some reason "hub" tool wasted my subject. It should read:

```
Revisited our redhat development deps
```

PR #133, opened 2013-12-23
api V2 update of examples

isratrade:


isratrade:
The previous docs did not have the v2 foremat in the examples.
- 'results' node for index
- no node for show

domcleal:
Thanks @isratrade!

PR #135, opened 2013-12-27
Manual 1.3: Updating section 4.2.5 to include a brief example of templat...

mrcrilly:
...e variables within class parameters. Also included are links to the wiki page entries and a screenshot.


domcleal:
Thanks @mrcrilly!

PR #137, opened 2013-12-30
Updating docs to include information on the Hosts field in Puppet Classe...

mrcrilly:
Mention of the bug report in place to remove confusion.

### DIFF
--- a/_includes/manuals/1.3/4.2.2_puppet_classes.md
+++ b/_includes/manuals/1.3/4.2.2_puppet_classes.md
@@ -50,3 +50,9 @@ and then go to the Puppet Classes tab and select what classes you want in this H
 ##### Checking the results
 
 To see how Foreman is passing the classes to Puppet, go to a Host and click the YAML button. You will be shown the exact YAML data sent to the PuppetMaster - the classes will be in the "classes" hash.
+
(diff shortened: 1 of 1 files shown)


_includes/manuals/1.3/4.2.2_puppet_classes.md
```diff
@@ -50,3 +50,9 @@ and then go to the Puppet Classes tab and select what classes you want in this H
 ##### Checking the results
 
 To see how Foreman is passing the classes to Puppet, go to a Host and click the YAML button. You will be shown the exact YAML data sent to the PuppetMaster - the classes will be in the "classes" hash.
+
+##### The "Hosts" Column
+
+This column currently suffers from a [known bug](http://theforeman.org/issues/3846). This bug means this count excludes Hostgroups from final figure.
```
domcleal: This section is about the host page, rather than the puppet classes page, so it's not clear which column you're referring to here.

Also, "host groups" rather than "Hostgroups".

mrcrilly: Well put! Be right back...

domcleal:
Thanks @mrcrilly :)

PR #138, opened 2014-01-01
topbar and landing page revised.

abenari:


static/css/style.css
```diff
@@ -1,7 +1,7 @@
 html { height: 100%; overflow: auto;}
 body { height: 100%;  }
 #wrap { min-height: 100%; }
-#main {padding-bottom: 38px; padding-top:60px; }
+#main {padding-bottom: 38px; padding-top:90px; background-color: #f2f2f2;}
```
domcleal: I'm unsure about this background colour on all pages.  The manual in particular looks a little strange - especially the contrast with code blocks and alert blocks.

_includes/releasesplash.html
```diff
@@ -1,8 +1,9 @@
 <!-- releasesplash.html -->
-<a class="btn btn-success btn-small"
+<div class="quick-links">
+  <a
     title="1.3.2 release notes"
-    href="{{ site.baseurl }}manuals/1.3/index.html#Releasenotesfor1.3">Stable: 1.3.2</a>
```
domcleal, line 4: I like having the latest release version clearly displayed, many projects do something similar.  Can we keep it somewhere?

_layouts/homepage.html
```diff
@@ -1,16 +1,20 @@
...
-      <div class="hero-unit">
+    <div class="container">
+      <div class="site-description">
+        <h3>
+            {{ site.description }}
+        </h3>
+        <a class="btn btn-info btn-large" href="media.html">Learn more</a>
+        <a class="btn btn-primary btn-large" href="manuals/1.3/quickstart_guide.html">Get started!</a>
+      </div>
+      <div class="container latest-news">
```
domcleal: Something for later I expect, but a more visible news feed would be great.  Maybe from the blog, and we update that instead.

domcleal:
Small thing - the lines between the three big icons on the home page and the titles underneath (Provisioning/Config/Monitoring) aren't visible because of the page background colour.

_includes/releasesplash.html
domcleal, line 4: Note that we currently put a second button alongside to indicate the RC version as well as the stable one when we're coming up to a release.

domcleal:
Good updates, that's quite nice.  A couple of comments:
1. the screenshots are now much smaller than they were on the media page, and partly obscured (certainly the dashboard one) by the text overlay.  Maybe an option to expand them is needed?
2. there's a lot of bold text on the home page, particularly the "Through deep integration" section.  Could bold be removed from this?
3. the home page shifts height as the image rotates, since image 1 is much taller than the rest, and image 2 is a slightly different size

abenari:
@domcleal PR is ready for review, issues fixed. Screenshots are smaller than the upstream but I think they do give the general idea of what to expect when installing foreman.

domcleal:
Moving the text off of the screenshots has helped a lot, thanks.

Merged - great work @abenari!

domcleal:
Some feedback on -dev suggests we should move the three icons (provisioning/config/monitoring) up so they're more prominent again.  They're probably more important than the "through deep integration" text.

PR #139, opened 2014-01-06
Add db:seed instructions to source install

domcleal:
I'll update the 1.4 branch too for package installs, upgrades etc.  I'm only updating the 1.3 manual as it contains source install instructions for develop.


GregSutcliffe:
Thanks!

PR #150, opened 2014-01-17
Add documentation on chefproxy feature

rvrignaud:


domcleal:
Looks like a good start, thanks.  Could you rebase it (seems already to be out of sync)?  I think we added a new section the other day here, it might need to be .10.

@ares?

ares:
Seems good, covers all that is needed. We may add some setup isntructions when we finish packaging

PR #153, opened 2014-01-20
Another manual change for the 1.4 manual

abradshaw:


domcleal:
Thanks again @abradshaw!

PR #155, opened 2014-01-27
No need to install Puppet before installation any more

domcleal:


skottler:
Merged as e5e50b8adc998f4ca4f89a718965b505dfc9e0b4. Thanks!

Issue #141, opened 2014-01-09
Foreman 1.4 manual


domcleal:
Tasks:
- [x] mod_auth_kerb #3312
- [x] Chef #3699 (partially, will be expanded into a full section)
- [x] proxy SSH puppetrun #3047
- [x] proxy Salt puppetrun #3732
- [x] proxy custom command #3936
- [x] update Rails console docs to use foreman-rake console #3347
- [x] db:seed #3752
- [x] libvirt/dnsmasq providers #3943
- [x] REMOTE_USER attrs(?) #3528
- [x] mysql/mysql2
- [x] Fedora 18 gone
- [x] Hammer CLI release notes
- [x] Debian plugin installation instructions
- [x] RHSCL
- [x] compute profiles #3178
- [x] Swedish
- [x] Link all issue numbers
- [x] Contributors list
- [x] Sprint 19 release notes
- [x] update installer options
- [x] update Foreman settings section
- [x] fix refs to 1.3 in API section
- [ ] update Hammer success story examples
- [ ] SLES provisioning #3733
- [ ] FreeBSD provisioning #2572
- [x] headline features
- [x] installer module changes from sprint 18, 19
- [x] installer requirement for Puppet 2.7
- [x] API doc refresh
- [x] Menu redesign, update references to "More"
- [x] User data support
- [x] Remove user/role cycle upgrade note, bumped to 1.5
- [x] Upgrade instructions/section is 1.2 to 1.3
- [x] proxy dnscmd provider
- [ ] APIv2 spec, error responses

mmoll:
I guess this can get closed? :)

domcleal:
@mmoll I've left them open when there are still incomplete items.. maybe we could migrate the incomplete items to new issues and it may be easier for people to contribute then?

mmoll:
:+1: for single issues

domcleal:
#315, #316, #317

PR #159, opened 2014-01-30
SPNEGO authentication and REMOTE_USER_MAIL #3528

dLobatog:
This is meant to address #3528 REMOTE_USER attributes and the setup required for that using mod_auth_kerb. https://github.com/theforeman/theforeman.org/issues/141

### DIFF
--- a/_includes/manuals/1.4/5.7_spnego_authentication.md
+++ b/_includes/manuals/1.4/5.7_spnego_authentication.md
@@ -1,0 +1,78 @@
+
+Foreman 1.4 supports SPNEGO authentication. The following tutorial explains how to set up authentication with mod_auth_kerb and any IPA server. We will use an external IPA server and install mod_auth_kerb in the Foreman machines.
+
+We assume the Foreman machine is IPA-enrolled:
+{% highlight bash %}
+ipa-client-install
+{% endhighlight %}
+
+On the IPA server, we create the service:
+{% highlight bash %}
+ipa service-add HTTP/<the-foreman-hostname>
+{% endhighlight %}
+
+On the Foreman machine, we get the keytab for the service:
+{% highlight bash %}
+ipa-getkeytab -s ipa.example.com -k /etc/http.keytab -p HTTP/$( hostname )
+chown apache /etc/http.keytab
+chmod 600 /etc/http.keytab
+{% endhighlight %}
+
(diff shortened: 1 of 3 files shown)


_includes/manuals/1.4/5.7_spnego_authentication.md
```diff
@@ -0,0 +1,78 @@
+
+Foreman 1.4 supports SPNEGO authentication. The following tutorial explains how to set up authentication with mod_auth_kerb and any IPA server. We will use an external IPA server and install mod_auth_kerb in the Foreman machines.
```
domcleal: Avoid mentioning a specific Foreman version, we're in the 1.4 manual.

_includes/manuals/1.4/5.7_spnego_authentication.md
```diff
@@ -0,0 +1,78 @@
...
+:authorize_login_delegation: true
+:authorize_login_delegation_auth_source_user_autocreate: External
+{% endhighlight %}
+
+On Foreman machine, restart Apache:
+{% highlight bash %}
+service httpd restart
+{% endhighlight %}
+
+Now in your browser, if you ```kinit``` to obtain a ticket, accessing Foreman's WebUI should not ask for login/password and should display the authenticated dashboard directly. If the user was just created, page asking for the email address of this new user will be shown.
```
domcleal: "a page asking"

_includes/manuals/1.4/5.7_spnego_authentication.md
```diff
@@ -0,0 +1,78 @@
...
+
+#### Disabling auto-creation of externally authentication users
+If only already existing users should be allowed to log in, remove/comment out these lines
+{% highlight yaml %}
+:authorize_login_delegation: true
```
domcleal: I don't think this line should be removed, just the autocreate one.

It would be better to remove this little section and make it clear in the above section whether to add autocreate or not.

dLobatog: Right, it's just the autocreate setting that would need to be removed. I'll use the positive and explain how to enable it

dLobatog:
@domcleal I rewrote the parts you commented about

domcleal:
Thanks @eLobato, merged!

PR #160, opened 2014-01-30
visual changes to the landing page and a new set of slides for the learn-more page

abenari:


domcleal:
Very nice, I like that.  Couple of comments:
1. there's a large white space in the footer of the home page on Firefox, below latest news
2. there's no indication of how to browse the presentation (arrow keys only it seems).  If we can get something to click on, that'd be great, else a note to say to use arrows.

ohadlevy:
@domcleal normally when using strut it gives you a visual hint if you dont touch anything after 5 seconds, I would assume it should work here too.

domcleal:
@ohadlevy you're right!

lzap:
This is great.

domcleal:
Thanks @abenari!

PR #161, opened 2014-01-30
API V2: Adding a section for handling associations

daviddavis:


_includes/manuals/1.4/5.1.7_associations.md
```diff
@@ -0,0 +1,40 @@
...
+depending on the type of association.
+
+#### One-to-One and One-to-Many
+
+To update a one-to-one or a one-to-many association, simply set the id on the
+object. For example, to set a hostgroup for a host, simply set the hostgroup_id of
+the host.
+
+    $ curl -k -u admin:changeme -H "Accept: version=2,application/json" -X POST \
+        -d { "hostgroup_id": 42 } https://foreman.example.com/api/host/2
```
domcleal: We should use names rather than IDs in examples where possible.

_includes/manuals/1.4/5.1.7_associations.md
```diff
@@ -0,0 +1,40 @@
+Updating and creating associations are done in a few different ways in the API
+depending on the type of association.
+
+#### One-to-One and One-to-Many
+
+To update a one-to-one or a one-to-many association, simply set the id on the
+object. For example, to set a hostgroup for a host, simply set the hostgroup_id of
```
domcleal: "host group"

_includes/manuals/1.4/5.1.7_associations.md
```diff
@@ -0,0 +1,40 @@
...
+
+#### Many-to-One and Many-to-Many
+
+To update an association for an object that contains a collection of other
+objects, there are a few options. First you can set the ids:
+
+    $ curl -k -u admin:changeme -H "Accept: version=2,application/json" -X POST \
+        -d { "host_ids": [1,2,3] } https://foreman.example.com/api/hostgroup/42
+
+This will set the hosts of the hostgroup to 1, 2, and 3 and only those.
```
domcleal: ditto

_includes/manuals/1.4/5.1.7_associations.md
```diff
@@ -0,0 +1,40 @@
...
+    $ curl -k -u admin:changeme -H "Accept: version=2,application/json" -X POST \
+        -d { "hostgroup_id": 42 } https://foreman.example.com/api/host/2
+
+#### Many-to-One and Many-to-Many
+
+To update an association for an object that contains a collection of other
+objects, there are a few options. First you can set the ids:
+
+    $ curl -k -u admin:changeme -H "Accept: version=2,application/json" -X POST \
+        -d { "host_ids": [1,2,3] } https://foreman.example.com/api/hostgroup/42
```
domcleal: Missing quotes around the data?

Shouldn't this also set content-type to work?

domcleal:
This is missing from the manual index, so I can't see how it would be rendered at the moment - can you add it please?

@isratrade could you review too?

_includes/manuals/1.4/5.1.7_associations.md
daviddavis: I was following the other examples but you're right, I think I need to set the content type.

isratrade: @daviddavis, content-type is need for POST/PUT but not GET

_includes/manuals/1.4/5.1.7_associations.md
```diff
@@ -0,0 +1,48 @@
...
+        https://foreman.example.com/api/subnets/42
+
+This would set the domains for the subnet to be earendil and turgon. If domain
+with id 2 for example belonged to the subnet before the request, it would be
+removed.
+
+Some objects can contain a nested route which allow you to add or remove a set of objects:
+
+    $ curl -k -u admin:changeme -H "Accept: version=2,application/json" -X POST \
+        -d { "id": [1,2,3] } \ https://foreman.example.com/api/operatingsystems/2/architectures
```
isratrade: @daviddavis, this is not correct.  It's what bk suggested, but it correctly is not implemented so it shouldn't be in the manual for 1.4.
I suggest me we do 

```
PUT /operatingsystems/2 -d { "architectures": [{"id": 1},{"id": 2},{"id": 3}] }
```

daviddavis:
@domcleal @isratrade I think I addressed all the feedback.

domcleal:
Thanks @daviddavis, merged!

PR #165, opened 2014-02-05
Add NOPASSWD to sudoers example for running virsh

jbrouault:
Without the NOPASSWD tag, smart-proxy fails
to run virsh and the following error is logged:

ERROR -- : DHCP virsh provider error: unable
to retrive virsh info: DNS virsh provider error:
virsh call failed (pid 569 exit 1) - sudo: no tty
present and no askpass program specified


domcleal:
Good catch, thanks @jbrouault!

lzap:
Yeah thanks, I have a long session and I did not notice.

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

PR #168, opened 2014-02-08
Section 4.2.2 - Show that The Foreman only autoloads classes in modules

blbradley:
Please let me know if this comment should go anywhere else.


domcleal:
Looks good, thanks @blbradley!

PR #171, opened 2014-02-10
Post-installation import idea

ekohl:
I added myself to the contacts because I'm not sure yet if I have the time and/or knowledge to properly mentor it.


dLobatog:
Thanks @ekohl

PR #172, opened 2014-02-10
Fix typos, add additional step

larkit-ian:
- service principal name in the grant statement was inconsistent with above
- service principal name in the Smart Proxy configuration was inconsistent with above
- added step to verify that "Dynamic Update" is enabled for the zone.
- added a statement to update both forward and reverse zones, as desired


larkit-ian:
Figured I'd fix a few things that bit me hard over the last few days.  Hope it helps someone avoid going down a bad path with GSS-API and IPA... ;)

domcleal:
Thanks for the fixes @larkit-ian!

PR #180, opened 2014-02-24
Update 5.2.4_cr_gce.md

mobcdi:
Clarified the need for the user to currently transfer the private key for GCE to foreman outside of the UI


domcleal:
Thanks for the fix!

PR #193, opened 2014-03-13
Adding nameserver setting to libvirt provider

lzap:


domcleal:
Merged, thanks.

lzap:
Well I just learned from Greg that Foreman connects to DNS servers
directly, only if a setting is turned on it goes via system.

Should we mention this here?

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

domcleal:
I thought when you were looking yesterday it wasn't actually doing that?

My suspicion then looking at the code is that dnsmasq doesn't publish an SOA or NS records, so this method of going to authoritative nameservers wasn't actually working.

lzap:
You are absolutely right.

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

PR #200, opened 2014-03-25
An archiever tool is a build dependency of nokogiri

isimluk:
On Fedora 20 minimal installation there is no tar package installed
by default. That causes `gem install nokogiri -v '1.6.1'` fail with:

   Extracting libxml2-2.8.0.tar.gz into tmp/x86_64-redhat-linux/ports/libxml2/2.8.0... '''--'''ERROR
   sh: zxf: command not found
   **\* extconf.rb failed ***
   Could not create Makefile due to some reason, probably lack of necessary


domcleal:
Merged, thanks @isimluk!

PR #204, opened 2014-04-01
Update 5.2.9_cr_vmware.md

iarebatman:
Add note for those who might run into permission issues while using VCenter with Foreman.


domcleal:
Thanks!

PR #205, opened 2014-04-03
clarification on TFTP workflow

stefancocora:
Added clarification on what is needed to trigger the TFTP workflow


domcleal:
Thanks @stefancocora!

PR #210, opened 2014-04-10
subscribe_to_all_hostgroups release note

dLobatog:


domcleal:
I'd suggest moving this to the upgrade section (3.6) to either the prep or the post-upgrade tasks.

dLobatog:
Sure, this can only be done post-upgrade.

dLobatog:
@domcleal 3.6 mentions -

Please check the upgrade notes for this release, as there may be things to be aware of or additional steps:

Are you sure it should go in there? It seems like everything in 3.6 is the standard procedure to upgrade foreman, not any particularities.

domcleal:
Are you reading the right Foreman version?  The 1.4 manual says that and it was split between upgrade notes near the top and then upgrade instructions in 3.6, but in the 1.5 manual I tried to combine it.

http://theforeman.org/manuals/1.5/index.html#3.6Upgrade

dLobatog:
Nope, I was in fact reading the 3.6 section of an old manual version, thanks.
I updated the branch accordingly.

domcleal:
Thanks, merged!

PR #215, opened 2014-05-16
No need of rpmnew copy anymore

lzap:
See https://github.com/theforeman/foreman/pull/1451


domcleal:
See https://github.com/theforeman/foreman/pull/1451#issuecomment-43312151 :)

domcleal:
Merged as abf2c55 onto 1.6-wip, thanks!

PR #220, opened 2014-05-27
Added a talk and article about Foreman in german language

mattiascockburn:
Heya,

as discussed in the IRC i have added a talk i gave at a german linux conference as well as an article i have written for a german IT magazine


domcleal:
Great, thank you!

PR #224, opened 2014-06-05
On default installation, authorize_login_delegation and authorize_login_delegation_api seem false.

adelton:


ares:
ACK, see https://github.com/theforeman/foreman/blob/1.5-stable/app/models/setting/general.rb#L20

domcleal:
Thanks!

PR #225, opened 2014-06-06
Change DNS Section to show correct dns_key option

hnk:
The documentation states that "[...] This option should not be used if you plan to use Kerberos/GSS-TSIG [...]" which is a bit confusing, as you have to set this option to 'false' in order to avoid connection errors.


domcleal:
Thanks!

We're actually treating this as a bug (http://projects.theforeman.org/issues/5675), it was an unintentional change, but the doc fix is good all the same.

PR #228, opened 2014-06-12
Added doc section for jsonp to the config options

ghost:
As stated in IRC - here is the suggestion for the JSONP documentation for configuration options.


domcleal:
Thanks @rhaen!  Merged as 4346160.

PR #231, opened 2014-06-20
Explanation of LDAP filters

dLobatog:
I think this should go in for 1.6 too


_includes/manuals/1.5/4.1.1_authentication.md
```diff
@@ -14,6 +14,7 @@ Click on New Authentication Source and enter the following
 * *Account*: leave this field empty if your LDAP can be read anonymously, otherwise enter a user name that has read access to the LDAP or use $login (which will be replaced with the actual user credentials upon login)
 * *Account Password*: password for the account (if defined above and its not using the $login)
 * *baseDN*: the top level DN of your LDAP directory tree
+* *LDAP filter* (optional): a filter to restrict your LDAP queries, for instance: (memberOf=cn=foreman-users,ou=groups,dc=example,dc=net). Multiple filters can be combined using the syntax (& (filter1) (filter2)).
```
domcleal: nitpick, would be good to use backticks around the filters so it's clearer that the brackets are part of the value

dLobatog:
I've put the backticks around filters.

domcleal:
Thanks @eLobato!

PR #235, opened 2014-07-10
add dns section to libvirt example

mmoll:


domcleal:
Thanks!

lzap:
Why is that? Does it fail without it?

mmoll:
Yes, in the strange way that records are added, but removing fails. Probably because of the way DHCP and DNS is combined in dnsmasq.

PR #237, opened 2014-07-23
DO NOT MERGE fixes #6749 - regenerate API docs, fix API v2 examples in docs that show error messages

isratrade:
...


isratrade:
@inecas, note that .html is removed from most of the links, which causes a problem when I run jekyll locally. If this a problem for jekyll on theforeman.org? @domcleal

domcleal:
If it's a problem locally when you run "jekyll build" then it'll affect theforeman.org too, yep.  It just runs that and then copies the contents of _site to the website.

isratrade:
Thanks @domcleal, I will wait for @inecas. Previously, I only needed to manually change 2 links as described in the README on https://github.com/theforeman/theforeman.org

iNecas:
what command was used to generated the docs?

isratrade:
@inecas, rake apipie:cache

iNecas:
It seems apipie 0.1.3 had this issue, when testing against the latest, it worked again. This PR https://github.com/theforeman/foreman/pull/1428 updates the apipie-version, and I'm going to release new version of apipie today, so that the 1428 could be put in as well…

iNecas:
One update, I will wait with new apipie-version release when @mbacovsky confirms that all the changes are in. Anyway, if you need it just for the docs, you can update the version requirements locally in you foreman version and generate the apipie:cache, the 2.1 seemed to work fine

isratrade:
In the Foreman Gemfile we have

``` ruby
gem "apipie-rails", "~> 0.1.1"
```

I'll update to ~> '0.2'`. @domcleal, is there past history to constain the gem 0.1.3

domcleal:
@isratrade IIRC there were small breakages in the early days, so was probably pinned for safety.  I'd say keep it within the 0.1 and 0.2 series, unless @iNecas disagrees.  (Bear in mind that our stable branches last ~6 months so we need it to remain working.)

iNecas:
Yeah, although we try to keep the breaking changes at minimum, it's safer to keep the minor version locked: we try to follow semantic versioning, '~> 0.2.0' should be fine

domcleal:
Is this working OK now @isratrade?  I'd quite like a refresh of the docs.

domcleal:
#242 is updated.

PR #243, opened 2014-08-08
update README section on Updating API Auto-Generated Docs by apipie

isratrade:


isratrade:
removed section in README that is no longer need with apipie-rails 0.2.x

PR #241, opened 2014-08-04
Update 3.5.4_puppet_reports.md

boaz0:
Change the foreman-report_v2.rb.erb url to the one at 2.1_stable branch.
Puppet-foreman's master has no longer the file foreman-report_v2.rb.erb.

See more here:
https://groups.google.com/forum/#!topic/foreman-dev/XA2tXbNlPq8

PR #244, opened 2014-08-10
Update 3.5.5_facts_and_the_enc.md

boaz0:
Fix a problem that is similar to https://github.com/theforeman/theforeman.org/pull/241


domcleal:
Thanks!

PR #245, opened 2014-08-11
Use foreman-rake for backup/import

dLobatog:
Unless I'm missing some reason, users should leverage https://github.com/theforeman/foreman/blob/develop/lib/tasks/backup.rake rather than manually backing up their databases, at least for simple dumps.

### DIFF
--- a/_includes/manuals/1.8/5.5.1_backup.md
+++ b/_includes/manuals/1.8/5.5.1_backup.md
@@ -4,28 +4,12 @@ instance.
 
 ### Database
 
-#### PostgreSQL
+#### PostgreSQL, MySQL, SQLite
 
-The `pg_dump` command can be used to backup the contents of a PostgreSQL
-database to a text file. The most reliable way of backing up PostgreSQL
-database is:
+Run ```foreman-rake db:dump```. It will print a message when it finishes with the dump
+file location relative to the Foreman root.
 
-    su postgres -c "pg_dump -Fc foreman > foreman.dump"
-
-#### MySQL database
-
-The `mysqldump` command can be used to backup the contents of your MySQL
-database to a text file.
-
-    mysqldump --opt foreman > foreman.sql
-
-Please follow MySQL documentation about other options for doing backups.
-
-#### SQLite database
-
-To backup SQLite database perform this simple step:
-
-    sqlite3 /path/to/foreman/db/production.db .dump > foreman.dump
+##### SQLite disclaimer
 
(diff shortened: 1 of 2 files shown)


_includes/manuals/1.5/5.5.2_recovery.md
```diff
@@ -7,30 +7,19 @@ beginning of this chapter.
 
 Note: Foreman instance must be stopped before proceeding.
 
-#### PostgreSQL
+#### PostgreSQL and MySQL
 
-The `pg_restore` command can be used to recover contents of your PostgreSQL
-database from the database dump. The most reliable way PostgreSQL
-database is:
+Run ```rake db:import_dump file=/your/db/dump/location```. This will load your
```
domcleal: foreman-rake

dLobatog:
Updated.

domcleal:
This doesn't seem to work on an RPM installation anyway, as the foreman user (correctly) doesn't have writes to write to ~foreman.  I think that needs fixing before we can recommend it:

<pre>
[root@foreman tmp]# foreman-rake db:dump
sh: foreman.1407745041.sql: Permission denied
Backup in file foreman.1407745041.sql
</pre>

dLobatog:
https://github.com/theforeman/foreman/pull/1679 makes some changes so that it works by default on an RPM installation

domcleal:
@eLobato could you change this to update the 1.8 manual only please, as Foreman 1.8 will contain your fixes at last :)

dLobatog:
@domcleal Done, changes are for 1.8 instead of 1.5, thanks for the heads-up

_includes/manuals/1.8/5.5.1_backup.md
```diff
@@ -4,24 +4,12 @@ instance.
...
-database to a text file.
-
-    mysqldump --opt foreman > foreman.sql
-
-Please follow MySQL documentation about other options for doing backups.
-
-#### SQLite database
+#### SQLite
 
 To backup SQLite database perform this simple step:
```
domcleal: Do you think we should recommend the rake task for sqlite too, or just leave as-is?

dLobatog: I'll update it with that, I think originally the task didn't support sqlite.

domcleal: Yeah, indeed.

dLobatog:
@domcleal Updated.

domcleal:
Thanks @eLobato!

PR #246, opened 2014-08-15
Smart proxies only require port 8443.

lfisher47:
DNS smart proxy is only port 8443, but if you installed the DNS server on that host too, then you need the port open.  Changed wording to reflect this since if your DNS smart proxy points to a remote host, you don't need the DNS ports open.


domcleal:
Thanks @lfisher47!

PR #249, opened 2014-08-17
Add section on search syntax and usage

domcleal:
Copied from http://blog.theforeman.org/2012/01/advance-search-tips-in-foreman.html
with small alterations.


ohadlevy:
Thanks @domcleal

PR #252, opened 2014-08-21
Optional upgrade step to run installer

ares:
I expect some discussion on mailing list first, then we'll update/merge/close this PR.

### DIFF
--- a/_includes/manuals/1.6/3.6_upgrade.md
+++ b/_includes/manuals/1.6/3.6_upgrade.md
@@ -154,6 +154,21 @@ Debian users should reinstall plugins if desired using apt at this point.
 
 etc.
 
+##### Optional Step 3 (C) - Run foreman-installer
+
+If you used foreman-installer to set up your existing Foreman instance we 
+recommend running it again after upgrading. Note that the installer can 
+modify config files so this may overwrite your custom changes. You can run 
+the installer in noop mode so you can see what would be changed.
+
+To see what would happen
+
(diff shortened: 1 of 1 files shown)


lzap:
I like this idea, but I've been told that `--noop` in Puppet mean something that can break (while proper run works fine). :+1:

ares:
@lzap yes, this can happen if you have two resources, A depends on updating B but in `--noop` B won't  update so A fails, I think during upgrades it should not happen too ofter

_includes/manuals/1.6/3.6_upgrade.md
```diff
@@ -154,6 +154,21 @@ Debian users should reinstall plugins if desired using apt at this point.
 
 etc.
 
+##### Optional Step 3 (C) - Run foreman-installer
+
+If you used foreman-installer for existing foreman instance we recommend
```
domcleal: s/for existing foreman instance/to set up your Foreman instance/

_includes/manuals/1.6/3.6_upgrade.md
```diff
@@ -154,6 +154,21 @@ Debian users should reinstall plugins if desired using apt at this point.
...
+running it again after upgrading. Note that installer can modify config
+files so this may overwrite your custom changes. You can run installer
```
domcleal: s/installer/the installer/ (above and below) too

ares:
fixed

domcleal:
Thanks @ares!

PR #259, opened 2014-09-08
Added deviantony blog entry about a scalable architecture for Foreman & Puppet on Ubuntu 12.04

deviantony:
Everything is in the title :-)


domcleal:
Thanks @deviantony!

PR #260, opened 2014-09-09
Remove project taken by marianitadn

dLobatog:


domcleal:
Or change the status, to show it's active?

dLobatog:
I think of this as a directory for new projects for the next round, isn't it? I don't know if new interns will take started projects, so the status might not mean so much in here.

PR #261, opened 2014-09-09
LDAP security disclaimer

dLobatog:


_includes/manuals/1.6/4.1.1_authentication.md
```diff
@@ -189,6 +189,10 @@ On the *External groups* tab, click the *Add external user group* button to open
 
 When a user logs in for the first time (assuming on the fly account creation), the ldap:refresh_usergroups cronjob runs (every 30 minutes by default) or the *Refresh* button is pressed next to the external user group entry, Foreman will synchronize the group membership from LDAP.
 
+##### Security Disclaimer
+
+Please remember your external user groups will only be refreshed automatically when an user logs in, or through the ldap:refresh_usergroups cronjob. There can be a lapse of time between the user logs in and the cronjob runs, in which if the user groups in LDAP change, the user will be logged in and assigned to the wrong external user groups. This situation can be quickly fixed by manually running ```foreman-rake ldap:external_usergroups``` or by refreshing the external user groups in the UI. Otherwise, the problem will eventually get fixed when the cronjob runs or the user logs out.
```
domcleal: Only refreshed on log in subject to #7369?  At the moment it's only on the cronjob I thought.

dLobatog: Righto, to be merged after https://github.com/theforeman/foreman/pull/1744

domcleal: Can we have both? :) Have it reflect the current state, then as and when that's merged and released somewhere, we update the description.

dLobatog:
Updated to state how it works before merging https://github.com/theforeman/foreman/pull/1744

domcleal:
Thanks!

PR #265, opened 2014-09-17
Added Google Custom Search

lzap:
Experimental - this is how it looks like.

And this is how results look like:

It is possible to search in Bugzilla too, not sure if this is relevant here.

WARNING: Do not merge because the results can contain ADs and the money would
go on my account. We want this to go to Foreman account. So if you like this,
Ohad please generate new GCE (I went with all defaults) and send me the ID.

### DIFF
--- a/support.md
+++ b/support.md
@@ -23,6 +23,151 @@ If you run it without any options, it will collect data, filter out possible
 passwords or tokens and create a tarball which can be safely handed over to
 us.
 
+## Search in our issue trackers
+
+Before you even ask, try to search for your problem or error message in our
+issue tracker. Use this Google Search appliance to do that or head over
+directly to our [Issue tracker](http://projects.theforeman.org/projects/foreman/issues)
+
+<div id='cse' style='width: 100%;'>Loading</div>
+<script src='//www.google.com/jsapi' type='text/javascript'></script>
+<script type='text/javascript'>
+google.load('search', '1', {language: 'en', style: google.loader.themes.V2_DEFAULT});
+google.setOnLoadCallback(function() {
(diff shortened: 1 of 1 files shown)


lzap:
I added Red Hat Satellite bugzilla search as well which might be handy (some reports are there but missing upstream). Unfortunately we need two search boxes - I am not able to combine the two results with "keyword search: Red Hat Satellite 6" which is needed to filter out non-satellite bugs from bugzilla.

There is one way - if we modify our RedMine to include this text, then we can combine those into one search. That would be great IMHO. What you think?

lzap:
Ok the bugzilla search does not work well, it's all or nothing. Although I am able to add "Red Hat Satellite 6" keyword for the CSE so it returns more relevant results, it also shows results which are not relevant. Which might be maybe good...

domcleal:
Thanks @lzap!

lzap:
@domcleal @ohadlevy I've doublechecked and this GCE is not linked with any AdSense account.

domcleal:
@lzap phew.  sorry, I missed the do not merge in the middle of the thread :)

lzap:
Not your fault, this was social engineering attempt. I am millionaire
now! :-)

The GCE is linked to my private account and we should change this to
Ohad's one. Fortunately, AdSense must be explicitly enabled and I
haven't do that for the GCE.

Ohad can you create the same GCE and send me out your GCE ID (or HTML
snippet), I will raise new PR. It's as easy as creating new GCE and
adding this patter to the list of included sites:

```
http://projects.theforeman.org/issues/*
```

## 

Later,
 Lukas #lzap Zapletal

PR #267, opened 2014-09-23
Changed videos to media and added new articles

dLobatog:
I've changed Videos to Media since the page now includes some articles. Maybe "Learning material" would be more appropriate?


lzap:
+1

domcleal:
Thanks @eLobato!

PR #273, opened 2014-10-11
added centos7 specific details and further checks to tftp documentation

ghost:
I had some issues while deploying on CentOS 7, which were solved together with Michael Moll, from #theforeman irc. I would like to share those.


domcleal:
Thanks @Riptawr!

PR #279, opened 2014-10-16
added full path to config file ignored_environments.yml, fixed typo

ghost:


domcleal:
Thanks @Riptawr!

For future reference, you can either push more commits to the already-open PR just by doing a git push, or you can amend your existing commit and then do a "git push -f" to force-update the branch, replacing the content.  The PR'll update automatically.

PR #284, opened 2014-11-03
Console CSS and updated hammer success story

dLobatog:
- Fixes for some Hammer docs about the success story tracker and adds commands for issues that were closed
- <pre> blocks look like a white on black terminal, a little nicer and more readable than light grey on white in my opinion


domcleal:
I don't have much preference over colour scheme, but I haven't ever found the current style hard to read.  However it's not grey on white here, it's black on grey...

PR #296, opened 2014-12-11
upgraded contributors list on the release notes

karmab:
contributors list retrieved from the autogenerated page https://github.com/theforeman/foreman/blob/develop/Contributors
and added some basic formatting


domcleal:
The list of contributors here is generated by scripts/commiters.rb and only lists people who contributed between 1.6 and 1.7, so it's meant to be this way.  The contributors list in that repo is for all of history.

karmab:
i guess i m just frustrated i m not in the history list...

domcleal:
Hmm, that's strange, you did commit in that period.

domcleal:
I wonder if I ran 1.5 to 1.6 rather than 1.6 to 1.7.  Let me try updating it...

domcleal:
There we go, sorry about that!

PR #300, opened 2014-12-22
Update manual to make persistent firewall rules with ESXi.

Dan33l:
As requested on IRC, you can find here a PR about how to make persistent firewall rules with ESXi


domcleal:
Merged, thanks @Dan33l!

PR #310, opened 2015-01-16
Changed instructions highlight in Installation Scenarios

xnaveira:
This os to avoid that the commands are formatted as a code comments


domcleal:
Thanks @xnaveira!

xnaveira:
Thank you @domcleal

PR #322, opened 2015-01-28
Changed discovery 2.0 readme since https proxy url is not supported yet

orrabin:


domcleal:
Due to http://projects.theforeman.org/issues/9089 I assume.

Thanks @orrabin!

PR #323, opened 2015-01-29
Fix link for Scientific Linux.

stefanlasiewski:


domcleal:
Thanks @stefanlasiewski!

PR #325, opened 2015-02-05
XVNC port range increase + tips/tricks

AshtonDavis:
The port range in this document is no longer correct - at present, Foreman will try to connect as high as 5999.  Console was failing all over the place in my environment until I made this change.
Also, rather than keeping a copy of the firewall rule local to every system, it's more efficient to store it on shared storage and symlink.  This allows for quick rule updates and easy rule refreshing.

### DIFF
--- a/_includes/manuals/1.7/5.2.9_cr_vmware.md
+++ b/_includes/manuals/1.7/5.2.9_cr_vmware.md
@@ -12,7 +12,7 @@ Image based provisioning can be used by provisioning a new VM from a template an
 
 #### Console access
 
-Consoles are provided using VNC connections from Foreman to the ESX server, which requires a firewall change to open the respective ports (TCP 5910 to 5930)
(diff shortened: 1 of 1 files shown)


_includes/manuals/1.7/5.2.9_cr_vmware.md
```diff
@@ -12,7 +12,7 @@ Image based provisioning can be used by provisioning a new VM from a template an
 
 #### Console access
 
-Consoles are provided using VNC connections from Foreman to the ESX server, which requires a firewall change to open the respective ports (TCP 5910 to 5930)
+Consoles are provided using VNC connections from Foreman to the ESX server, which requires a firewall change to open the respective ports (TCP 5901 to 6001)
```
domcleal, line 15: I'm unsure about this high port number - https://github.com/theforeman/foreman/blob/1.7.2/app/models/compute_resources/foreman/model/vmware.rb#L452 suggests it should only be between 5901 and 5964.

domcleal:
Merged as 297d34e, thanks @ntent-ashton!

PR #326, opened 2015-02-10
Links not working & grammar change

duncaninnes:
Several links were pointing somewhere non-existant
Re-worded the configuration line slightly.


domcleal:
It seems that anchors aren't generated for fourth-level sections (e.g. 4.3.9.1) so the links still don't work.  Perhaps we should remove those and just link the top-level sections (DNS, DHCP etc)?

duncaninnes:
Can do.  Not even sure my link edits work.  I'll revise and come back.
The grammar edit for the last line should be OK.

PR #330, opened 2015-02-13
Add a note about the upgrade of the smart proxies

roidelapluie:


domcleal:
I _think_ smart proxies should restart by themselves, as the packages should contain post-install restart commands...

roidelapluie:
rpm -q --scripts foreman-proxy

```
postinstall scriptlet (using /bin/sh):
# Migrate legacy monolithic proxy config
if [ $1 == 2 -a -e /var/lib/rpm-state/foreman-proxy/settings.yml.orig ]; then
  pushd /var/lib/rpm-state/foreman-proxy >/dev/null
  if /usr/share/foreman-proxy/extra/migrate_settings.rb settings.yml.orig; then
    mv settings.yml /etc/foreman-proxy
    sed -i '/^---/ a #replace default location of "settings.d"\n:settings_directory: /etc/foreman-proxy/settings.d\n' \
      /etc/foreman-proxy/settings.yml
    rm -f settings.yml.orig
    ls *.yml >/dev/null 2>&1 && mv *.yml /etc/foreman-proxy/settings.d/
  else
    rm -f settings.yml.orig
  fi
  popd >/dev/null
fi

  /sbin/chkconfig --add foreman-proxy
  exit 0
```

domcleal:
I see a %postrun script:

<pre>
postuninstall scriptlet (using /bin/sh):
  if [ $1 -ge 1 ] ; then
    /sbin/service foreman-proxy restart >/dev/null 2>&1
  fi
</pre>

roidelapluie:
is Postuninstall triggered during an upgrade? At the right time?

domcleal:
Yeah, by the package being upgraded - for a short time, you have versions 1.7.1 and 1.7.2 both installed, and then 1.7.1 is uninstalled leading to %postun on 1.7.1's package being called.

roidelapluie:
Thanks

domcleal:
Np, thanks for raising it.  I'd like to fix the Foreman restart in a similar way, but that needs a little more thought...

PR #332, opened 2015-02-24
Update 4.3.7_smartproxy_puppet.md

LordBrain:
Added sudoers rule for puppetssh

### DIFF
--- a/_includes/manuals/1.7/4.3.7_smartproxy_puppet.md
+++ b/_includes/manuals/1.7/4.3.7_smartproxy_puppet.md
@@ -112,6 +112,11 @@ The puppetssh provider uses SSH to connect to the client using SSH keys and run 
 #:puppetssh_keyfile: /etc/foreman-proxy/id_rsa
 </pre>
 
+And then add a sudoers rule:
+
+    Defaults:foreman-proxy !requiretty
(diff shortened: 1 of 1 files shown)


_includes/manuals/1.7/4.3.7_smartproxy_puppet.md
```diff
@@ -112,6 +112,11 @@ The puppetssh provider uses SSH to connect to the client using SSH keys and run
 #:puppetssh_keyfile: /etc/foreman-proxy/id_rsa
 </pre>
 
+And then add a sudoers rule:
```
domcleal, line 115: perhaps add "if puppetssh_sudo is enabled"?  (I assume it won't sudo to root otherwise)

mmoll:
@blee1170 ping? could you add this to the 1.8 and 1.9 docs?

mmoll:
I'm leaning towards closing this as foreman-installer should be used anyway for that configuration. If wanted I'll open a PR to newer versions.

mmoll:
Closing this as there was no reaction. As already mentioned, foreman-installer should be used anyway. If anybody wants to pick this up, the howto of #451 would also be in need of an addition then.

PR #336, opened 2015-03-03
Add Docker and Foreman/Katello 101 from devconf.cz

dLobatog:
Ivan's talk is a really good introduction in my opinion (demo of discovery, provisioning, etc..)

### DIFF
--- a/_layouts/media.html
+++ b/_layouts/media.html
@@ -16,9 +16,13 @@ pagename: Media
       </ul>
       <div class="tab-content">
         <div class="tab-pane" id="talks">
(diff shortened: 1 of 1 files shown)


_layouts/media.html
```diff
@@ -16,6 +16,10 @@
       </ul>
       <div class="tab-content">
         <div class="tab-pane" id="talks">
+          <h3>Foreman & Katello: Automating the infrastructure thing - Ivan Nečas</h3>
```
domcleal: `&amp;`

mmoll:
thanks @eLobato!

PR #337, opened 2015-03-04
Consistent use of foremanproxy@example.com

kylegordon:


domcleal:
Much better, thanks @kylegordon!

PR #339, opened 2015-03-04
Make puppet_provider feature links point to config

dLobatog:
Currently section 4.3.7 of the manual goes over how to set up the configuration, but never 'where' to set it up. As a new user, I think reading this can be a head scratching moment if you don't know at all where to configure this.


domcleal:
Yeah, I was trying to put the config files in the section titles, but it didn't really work.  Thanks @eLobato!

PR #340, opened 2015-03-05
Rename OPW to Outreachy

dLobatog:
OPW is now called Outreachy, so we've been asked to rebrand it here. Also I'm removing GSoC references until the next application round.


mmoll:
thanks @eLobato!

PR #341, opened 2015-03-06
Add NYULUG talk

dLobatog:
Funnily enough Sam talks about extracting Puppet to a plugin.. in 2013, it's 2015 already! :)


mmoll:
thanks @eLobato!

PR #342, opened 2015-03-09
delete now invalid smart proxy test

mmoll:
Not very nice but OK as a quick fix not to direct people to the proxy URL which won't work without client certificate now.


domcleal:
Thanks!  We could perhaps add a recommendation to try foreman_dhcp_browser, but it could complicate matters.

PR #346, opened 2015-03-14
Update latest news and references to Salt

dLobatog:
The Google+ link is basically a way of linking the [profile page](https://plus.google.com/u/0/b/102496134326414788199/102496134326414788199/about?hl=en-GB) with the webpage, so that when people search in Google for Foreman, both the website and the Google+ community show up.


mmoll:
Looking good, the Google+ link is very small at least for me, maybe it's better to put it into the "Community" menu?

stbenjam:
Should the Salt bits be in the main manual? As it's not core, I figured it'd just belong only on the plugins page.

dLobatog:
True, but from the user point of view, I think not showing anywhere that
there is Salt support will lead them to simply think it's not there..

stbenjam:
Fair enough, we should add Chef too then.

mmoll:
#347 got merged, so this might need a quick tush-up :construction:

dLobatog:
I can put it anywhere you want, even hidden, the reason I want it there
is so that Google recognizes it and shows the community in the results :)

On 03/14, Michael Moll wrote:

> Looking good, the Google+ link is very small at least for me, maybe it's better to put it into the "Community" menu?
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/theforeman/theforeman.org/pull/346#issuecomment-80720396

## 

Daniel Lobato Garcia

@eLobatoss
blog.daniellobato.me
daniellobato.me

GPG: http://keys.gnupg.net/pks/lookup?op=get&search=0x7A92D6DD38D6DE30
Keybase: https://keybase.io/elobato

_includes/manuals/1.8/4.7_managing_salt.md
```diff
@@ -0,0 +1,2 @@
+
+Please refer to the [foreman_salt plugin documentation](https://github.com/theforeman/foreman_salt/wiki).
```
domcleal: {{site.baseurl}}/plugins etc now

domcleal:
I'm fine with the footer link, but could you also add the Google+ link to the support page?  And consider adding Chef at the same time.

dLobatog:
Right, I'll link to the manual in here https://github.com/theforeman/theforeman.org/pull/345

So this PR should be merged after that.

dLobatog:
@mmoll @domcleal Updated! I think a natural next step would be to add the Salt/Chef use cases to Installation scenarios, thoughts?

_includes/configuration.html
```diff
@@ -1,4 +1,4 @@
 
 <img class="homepage-img" src="{{ site.baseurl }}static/images/configuration.png">
 <h2>Configuration</h2>
-<p>A complete configuration management solution including an ENC for Puppet, built-in support for parameterized classes and hierarchical parameter storage.</p>
+<p>A complete configuration management solution including an ENC for Puppet and Salt, built-in support for parameterized classes and hierarchical parameter storage.</p>
```
domcleal, line 4: & Chef

support.md
```diff
@@ -223,3 +223,9 @@ We also have a few IRC-specific guidelines:
 More information: This [IRC primer](http://irchelp.org/irchelp/ircprimer.html) for new users and the [general IRC guidelines, from freenode](http://freenode.net/channel_guidelines.shtml), are also useful resources.
 
 (_Thanks to the [Puppet Labs IRC Guidelines](http://docs.puppetlabs.com/community/community_guidelines.html#irc-guidelines) for the original document_)
+
+## ![Youtube icon](static/images/youtube.png) [Youtube channel](https://www.youtube.com/channel/UCCo7AZ1oG6TbG0-dwjRqCmw)
+
+Please subscribe to our [Youtube channel](https://www.youtube.com/channel/UCCo7AZ1oG6TbG0-dwjRqCmw). We regularly upload deep dives on Foreman features, screencasts to learn how to use the different features and plugins of Foreman, discussions and much more.
```
domcleal: Nitpick, "YouTube"

domcleal:
Yeah, I don't know which way to do it with installation scenarios.  For instance, Chef has the documentation in the plugin docs, which closer matches the version of the plugin, which could be better.  We could refer to it from there too though.

_includes/configuration.html
dLobatog, line 4: We don't have an ENC for Chef...

domcleal, line 4: oops, sorry

mmoll:
Looking good, the Chef plugin docs would need to get merged first to avoid a dead link there.

mmoll:
thanks @eLobato :fireworks:

PR #349, opened 2015-03-18
Add hammer-cli-discovery deb install instructions

dLobatog:


domcleal:
Thanks @eLobato, seems I missed this between daef830 and the 3.0 docs.

PR #350, opened 2015-03-18
Fix formatting of hammer-cli yaml

dLobatog:
Current format renders well on Github preview but not on Jekyll.


domcleal:
Hm yeah.  Seems there's some more in section 4.4.  Could you fix that too please?

dLobatog:
@domcleal Thanks for the catch, done.

domcleal:
Sorry, could you also apply that fix to 2.0?  Otherwise :+1:

dLobatog:
Oh right, did that on a rush before lunch. Hope it's ok now.

domcleal:
Thanks @eLobato!

PR #351, opened 2015-03-18
Correction of two minor issues in the DHCP chapter

xenlo:
#### Two minor corrections in the DHCP chapter:
1. Chapter 4.3.5.1 dhcp.yml says that the dhcpd.conf file on redhat systems is located at /etc/dhcpd.conf. But actuall it is /etc/dhcp/dhcpd.conf (at least on rhel-6, rhel-7 and fedora-21).
2. Chapter 4.3.5.2 ISC DHCP refers to the omapi_key of the proxy's dns.yml. It should be the dhcp.yml (even the link is ok pointing to the chapter 4.3.5.1 dhcp.yml).


domcleal:
Thanks @xenlo!

PR #354, opened 2015-03-23
Wrong Link Syntax

xenlo:
This shoud be an copy from Foreman Wiki. I changed the syntax to Markdown link and provided the reference of the Wikipage : http://projects.theforeman.org/projects/foreman/wiki/Unattended_installations


domcleal:
Thanks @xenlo!

PR #358, opened 2015-03-27
add PuppetDB installer options

mmoll:


domcleal:
How about adding the puppetdb-foreman plugin to the argument list too?

mmoll:
I thought about it, but while the integration described here is for the Puppet master, this would be for Foreman and theforeman/puppet-foreman#235 would need to get done to have everything working really smooth.

domcleal:
Fair enough, thanks!

PR #360, opened 2015-03-30
Adding a note for connection refused error

SherifNagy:
While upgrading and https / apache2 service is down, you will get an error of evaluation of foreman_smartproxy, So we adding a note for the user not to panic


domcleal:
Thanks!

PR #364, opened 2015-04-06
Remove installation info duplicity

dLobatog:
Currently section 3.2.1 does not tell the user to set up EPEL, Puppet repositories, and it's a bit unfriendly compared to the quickstart installation page. In addition to that we have to maintain both sections for no benefit. I think linking to Quickstart installation should work well (or moving that to 3.2.1 and linking from Quickstart to that).


mmoll:
:+1:

domcleal:
Thanks!

PR #370, opened 2015-04-29
Improve NoVNC Documentation

roidelapluie:
In particular websockets_encrypt which takes on/off instead of
true/false.


domcleal:
Thanks @roidelapluie!

ekohl:
I'm slightly surprised by this since AFAIK the installer also writes true/false. Does that mean the installer is broken?

domcleal:
@ekohl hmm, I think it is then.  The code in the controller checks for the 'on' and 'off' strings, and assumes anything else (including true/false) would be 'auto'.

ekohl:
I'll submit a patch to fix that then.

roidelapluie:
On Thu, Apr 30, 2015 at 05:08:25AM -0700, Ewoud Kohl van Wijngaarden wrote:

> I'll submit a patch to fix that then.
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/theforeman/theforeman.org/pull/370#issuecomment-97751082

There are already 2:

https://github.com/theforeman/foreman/pull/2219
https://github.com/theforeman/foreman/pull/2340 (deprecated already)

## 

 (o-    Julien Pivotto
 //\    Open-Source Consultant
 V_/_   Inuits - https://www.inuits.eu

ekohl:
I was thinking of changing puppet-foreman, but maybe it should just be fixed in core.

ekohl:
https://github.com/theforeman/puppet-foreman/pull/319 is my go at it. It also adds a test and additional validation.

PR #378, opened 2015-05-11
fixes documentation for #10436 and corrects some other settings.d paths

smags:


mmoll:
merged, thanks @smags :tada:

@ares, I guess `4.3.4_chef_proxy.md` should be deleted as the information is in the plugin docs now, correct?

ares:
@mmoll you're right, good catch. Now the question is, should I delete it and renumber all chapters after it or should I keep their numbers and 4.3.4 number will be skipped?

domcleal:
It'd be handy if you could do the git mv, but if not then just update the index.md so the headings are correct.

ares:
opened as https://github.com/theforeman/theforeman.org/pull/380

PR #380, opened 2015-05-12
Remove Chef chapter

ares:


dLobatog:
Can we simply the section to point to the Chef manual? I ask because the Documentation buttons (haven't checked) rely on the numbers, and that might fail.

ares:
@domcleal suggested renumbering which also makes more sense to me, chef is no longer part of Foreman core in any way. But if you insist I can just add link there :-)

domcleal:
Hm, good point @eLobato.  We have links to the realm section in Foreman, so we should only change numbering in the 1.9 manual I suppose...

dLobatog:
Sounds fair, @ares mind making the changes on 1.9, and check if any documentation button would be affected? (I think only the one that links to Realms is) 
Should be good to go after that as far as I can see.

ares:
So I changed the chapter just to link (without renumbering). When we have 1.9 manual, this can be removed and links changed accordingly.

domcleal:
Thanks @ares, I'll put it on the 1.9 list.  Sorry for the back and forth.

PR #396, opened 2015-07-03
Add logging config and debugging docs

domcleal:
To replace http://projects.theforeman.org/projects/foreman/wiki/Troubleshooting#How-do-I-enable-debugging


dLobatog:
Merged as d82dfca959745744dd9dbab406044768ff31dca6, thanks @domcleal ! 
There are still a few links to the Troubleshooting wiki page you might want to check if they can be replaced or not.

PR #400, opened 2015-07-09
The option --foreman-configure-ipa-repo is no longer needed

veruu:
The option --foreman-configure-ipa-repo is no longer needed since the packages are included in RHEL/CentOS versions 6 and 7 now.

This feature was needed in older RHEL/CentOS versions, in which some packages weren't available. Now there is no need for it. We are discussing with @ares what is the best way to make the feature deprecated and remove it also from the code.

### DIFF
--- a/_includes/manuals/1.9/3.2.2_installer_options.md
+++ b/_includes/manuals/1.9/3.2.2_installer_options.md
@@ -212,10 +212,6 @@ More information about compute resources can be found in the [Compute Resources 
         <td>If disabled the EPEL repo will not be configured on RedHat family systems.</td>
       </tr>
       <tr>
-        <td style='white-space:nowrap'>--foreman-configure-ipa-repo</td>
-        <td>Enable custom yum repo with packages needed for external authentication via IPA, this may be needed on RHEL 6.5 and older.</td>
-      </tr>
-      <tr>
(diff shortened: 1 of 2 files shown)


_includes/manuals/1.9/3.2.2_installer_options.md
```diff
@@ -212,10 +212,6 @@ More information about compute resources can be found in the [Compute Resources
         <td>If disabled the EPEL repo will not be configured on RedHat family systems.</td>
       </tr>
       <tr>
-        <td style='white-space:nowrap'>--foreman-configure-ipa-repo</td>
```
domcleal, line 215: This table is auto-generated, so it needs to be removed from https://github.com/theforeman/puppet-foreman first.

mmoll:
As the parameter wasn't removed for the Puppet module version that will ship with 1.9, I'm leaning towards closing this for now.

mmoll:
As the option got removed for the 1.10 installer, could you please rebase onto current master and make the changes to the 1.10 files?

veruu:
I probably deleted my fork since I am unable to find it anywhere. I forked it once more and created a new pull request since I don't believe I am able to add it to this one.

https://github.com/theforeman/theforeman.org/pull/483

PR #406, opened 2015-08-04
Versioned apidoc

tstrachota:
Contains versioned api documentation for 1.6 .. 1.9 and updated readme file. Lot of users reference to the online api docs that contained outdated information. It was also not clear which version the documentation was generated for.


mmoll:
I clicked around a bit and this seems to be OK :+1:

domcleal:
Great, thank you @tstrachota!

PR #417, opened 2015-08-24
Added Organizations and Locations chapter for discovery

lzap:
I am providing documentation of current state for Discovery. This is fuzzy
area, we still have a room for improvement there.


domcleal:
Thanks!

PR #426, opened 2015-09-02
Update api url to new schema

dnsmichi:
The new target is http://theforeman.org/api/1.9/index.html but http://theforeman.org/manuals/1.9/index.html#5.1API links to the outdated url target.


dLobatog:
Thanks @dnsmichi, great catch :smile: !

dnsmichi:
You're welcome - I was just looking to add the url to an icinga blog post :)

PR #428, opened 2015-09-03
Display latest versions on homepage

### LINKED ISSUE #412: Display latest versions
The website used to have the latest version displayed clearly on the home page, but this was lost during #274.

The partial is still there at _includes/social.html and I'm regularly updating it, it just needs fitting into the new design somewhere.  Perhaps under the two buttons on the homepage?

domcleal:
Fixes #412 

---

Before:
screenshot from 2015-09-03 11-32-06

After:
screenshot from 2015-09-03 11-32-00


dLobatog:
Thanks @domcleal !

PR #429, opened 2015-09-03
Load Open Sans font from CDN for consistent rendering

domcleal:
Stylesheet prefers Open Sans already, so distribute it to ensure the
page is displayed consistently.


dLobatog:
Thanks @domcleal !

PR #432, opened 2015-09-07
Note AD's password change behaviour with Foreman LDAP auth

domcleal:
Document resolution of CVE-2015-5246

(from http://projects.theforeman.org/issues/11471)


mmoll:
merging, thanks @domcleal!

PR #436, opened 2015-09-21
Add projects page back to community tab

GregSutcliffe:
This is a quick re-add of a link to the projects. It merges in the Outreachy page and re-words it to be slightly more generic, so that it works for any internship program.

I've not reviewed the actual projects - I'll be contacting the people on here to check if their projects are still relevant soon, as well as soliciting new projects on the dev mailing list. New PRs can be opened for that.


GregSutcliffe:
Fixed commit message typo

projects.md
```diff
@@ -1,3 +1,27 @@
...
+layout: contribute
+title: Mentoring Projects
+---
+
+## Mentoring Projects
+
+This page details potential projects and their mentors, for participants of
+internship programs such as [Google Summer of Code](https://www.google-melange.com/),
+[Outreachy](https://www.gnome.org/outreachy/), and so forth. Each program will
+have it's own requirements and application process - this page simply lists the
```
domcleal: its

projects.md
```diff
@@ -1,3 +1,27 @@
...
+currently available projects.
+
+## I'm interested - how can I get familiar with Foreman?
+
+* Grab a cup of your favorite beverage
+* Skim through the [manual](/documentation.html)
+* Install Foreman on a virtual machine or your own computer
```
domcleal: Link to quickstart?

projects.md
```diff
@@ -1,3 +1,27 @@
...
+* In the meantime, watch a few [demo videos](/media.html)
+* If the installation hasn't finished yet, read a couple posts from the [mailing list](/support.html#Mailinglists)
+* Join us on [IRC](/support.html#IRC) at #theforeman and #theforeman-dev on Freenode. Don't be shy; tell us you're interested in one of our open mentoring projects.
```
domcleal: Perhaps prioritise -dev?

projects.md
```diff
@@ -68,3 +92,15 @@
   environment with puppet modules and parameters. This could be extracted into
   an installer to deploy a new host or maybe a Docker image. Working title for
   this is enc2answers.
+
+## Project Suggestions
+
+Send a pull request with the content of your proposal to [theforeman.org at
+Github](http://github.com/theforeman/theforeman.org). See [Dave Neary's
```
domcleal: GitHub

projects.md
```diff
@@ -68,3 +92,15 @@
...
+post](http://www.outercurve.org/Blogs/EntryId/45/Making-the-most-of-Google-Summer-of-Code-Dave-Neary-guest-blogger)
+for suggestions on how to propose ideas. Ideas should be useful to the
```
domcleal: I'd suggest putting in some paragraph breaks here, it's a bit bunched up.

projects.md
GregSutcliffe: gah, there's always one. thanks

projects.md
GregSutcliffe: do we have a generic link? the quickstart "tab" at the top goes to 1.9, so that will need updating each release...

domcleal: /manuals/latest/quickstart_guide.html would work in prod, if not locally.

projects.md
GregSutcliffe: i think i'll split it into to lines - the two channels have different purposes after all. good catch.

projects.md
GregSutcliffe: Added - why doesn't it work locally? what magic does it rely on? :)

projects.md
GregSutcliffe: Done

projects.md
GregSutcliffe: Yeah, that was a verbatim copy, pretty much. Spaced it out a bit now.

GregSutcliffe:
Updated, thanks @domcleal

projects.md
domcleal: https://github.com/theforeman/foreman-infra/blob/master/puppet/modules/web/templates/web.conf.erb#L13

It might be possible to fix that now with jekyll-redirect-from.

domcleal:
Thanks @GregSutcliffe.

PR #438, opened 2015-09-23
Rename projects to mentoring to avoid the redirect issue

GregSutcliffe:
Since we can't remove the /projects redirect...


domcleal:
Thanks!

PR #439, opened 2015-09-24
Added domain-less provisioning project to mentoring

dLobatog:
@GregSutcliffe sounds good?

### DIFF
--- a/mentoring.md
+++ b/mentoring.md
@@ -94,6 +94,15 @@ currently available projects.
   an installer to deploy a new host or maybe a Docker image. Working title for
   this is enc2answers.
 
+### [Domain-less provisioning for VMs](http://projects.theforeman.org/issues/11989)
+
+* *Status:* Not started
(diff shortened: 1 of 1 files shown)


mentoring.md
```diff
@@ -94,6 +94,15 @@ currently available projects.
   an installer to deploy a new host or maybe a Docker image. Working title for
   this is enc2answers.
 
+### Domain-less provisioning for VMs
+
+* *Status:* Not started
+* *Summary of idea*: Users can't create domains on demand on certain cloud services (AWS, DO, etc). We should fetch existing domains from compute resources, or at least allow creating virtual hosts without one. This would make Foreman more accessible as users don't necessarily want to deal with interfaces setup. We could even have an 'advanced' tick button or something that limits the Host form to the minimal requirements unless.
```
domcleal: stray "unless" at the end of the last sentence?

domcleal:
It'd be worth filing a ticket in Redmine too, since it's bordering on a regular feature request, and then linking between the two in case somebody wants to take up the mini-project.

mentoring.md
GregSutcliffe: From the preceding structure, I believe that's supposed to be something like "unless the button is checked"

GregSutcliffe:
+1 to filing an RFE, @eLobato can you update this with the ticket link?

dLobatog:
Updated!

domcleal:
Thanks!

PR #441, opened 2015-09-30
fixes #12006 - be more specific about windows

helge000:


domcleal:
@witlessbird do you want to double check this?

helge000:
@witlessbird: This only works with [PR318](https://github.com/theforeman/smart-proxy/pull/318) merged, otherwise bundle install will fail. Also, `bin/smart-proxy` will not run because puppet / ruby-augeas

domcleal:
We could recommend that you use `bundle install --without puppet puppetca development test` etc.

helge000:
@domcleal: this only helps with `bundle install`; smart proxy will not run because of missing gems; but `development test` is a good point; i'll change it right away!

domcleal:
It shouldn't need those gems unless those features are enabled, which they wouldn't normally be - e.g. you wouldn't have the realm feature enabled on a Windows installation.

helge000:
Hm: yet it does, at least while testing my changes yesterday for PR318 rebasing to develop. Running smart proxy seems to load any module and requiring any gem, even when the feature is disabled. I thought this was intended behavior?

domcleal:
I don't think that is intended, and hasn't been the case in the past at least.  Using bundler to exclude groups should mean that only a `require` from the smart proxy code should trigger it to be loaded, but none of those modules or providers ought to be loaded unless explicitly configured - which shouldn't be the case on Windows, since they're not too useful.  I'd probably have to see more details to comment further.

helge000:
@domcleal I think we should move this discussion [here](https://github.com/theforeman/smart-proxy/pull/318).

helge000:
@domcleal @witlessbird - handle bunder excludes by command line, not in bundle

dmitri-d:
Smart-proxy doesn't load dependencies for modules that are disabled. Installing smart-proxy using `bundle install --without krb5 puppet puppetca` and disabling the modules (they are disabled by default) should work (I tested smart-proxy on windows using these exact steps this week).

dmitri-d:
Thanks for updating the docs! 
- I think a note that only rubies stating with 1.9.3 are going to work on Windows would be useful. 
- I had issues with installing DevKit with ruby 2.0.0, and had to manually correct the config.yml. Might be worth mentioning?
- Some config file changes are required for smart-proxy to launch on windows: `bind_host` needs to be changed to `localhost` or interface ip address. `log_file` needs to be pointing at an existing directory.

helge000:
@witlessbird cool, your're right - using bunder excludes does sole the issue I had with puppet

helge000:
@witlessbird 

> I think a note that only rubies stating with 1.9.3 are going to work on Windows would be useful.

Hm. does't that belong to system requirements? I did not find any ruby version mentioning in anywhere so far. I did not want to start with that

> I had issues with installing DevKit with ruby 2.0.0, and had to manually correct the config.yml. Might be worth mentioning?

I think this is a bug in dev kit. Some registry keys are missing and this step is prone to fail. Should we mention another project's bugs in docs? Using the steps I did not see the issue. But I had it often enough.

> Some config file changes are required for smart-proxy to launch on windows: bind_host needs to be changed to localhost or interface ip address. log_file needs to be pointing at an existing directory.

Right. But smart-proxy does not run without config.yaml on all platforms; and this is the install and not the config section...

_includes/manuals/1.9/4.3.1_smartproxy_installation.md
```diff
@@ -19,6 +19,27 @@ You can get the latest stable code from [GitHub](https://github.com/theforeman/s
 git clone git://github.com/theforeman/smart-proxy.git
 </pre>
 
+#### Windows
+In Windows, the installation steps are different. You need to run smart-proxy from the source as well as install Ruby and Ruby DevKit. Smart-Proxy supports __ruby >= 1.9.3.__
+
+1. Run [Ruby Installer](http://rubyinstaller.org/downloads/) and add the matching and DevKit to compile native extensions. Make sure to add Ruby to `%PATH%`, you can select this option in the installer
+2. Download / clone the smart-proxy repository to a convenient location (see above, _Source Code_)
+2. Open a command prompt (`cmd.exe`) and run the following commands in order:
+  3. `ruby <devKitRoot>\dk.rb init`
```
domcleal: These need to be indented by another two spaces to render correctly.

domcleal:
Looks good to me.

_includes/manuals/1.9/4.3.1_smartproxy_installation.md
helge000: @domcleal good point; which markdown renderer should I use to generate a meaningful preview?

domcleal: We're using maruku.  If you do a `bundle install` and `rake server` then it'll run the exact same stack we use locally.

helge000:
@domcleal for some reason beyond me [this guide, already in the docs,](https://github.com/theforeman/theforeman.org/blob/gh-pages/_includes/manuals/1.9/4.3.4.3_ms_dhcp.md) slipped my attention. IMHO it clearly belongs to installation section as it is relevant for `dnscmd` also, and maybe for `realm` a long time from now in a galaxy far, far away ,)

I would merge these sections?

_includes/manuals/1.9/4.3.1_smartproxy_installation.md
helge000: Hm. this renders as intended using `maruku (0.6.1)`. Might it be a style sheet? Some files do not render[1] and I was unable to use `rake server`

[1] `Error reading file ~/theforeman.org/_layouts/api.html: undefined method 'value' for #<Psych::Nodes::Document:0x000000015333c8>`

domcleal:
Ah yes, I'd forgotten about that.  I'd suggest deleting it and referring, if even necessary, to the installation section as you suggest.

dmitri-d:
Looks good, other than the issues above.

helge000:
@witlessbird 

> We have a script that registers smart-proxy as a windows service: https://github.com/theforeman/smart-proxy/blob/develop/extra/register-service.rb

I just tried this again using develop and 19-stable. Alas, it never worked for me, that's why I moved to NSSM in the first place.

```
C:\smart-proxy\> ruby extra\register-service.rb
extra/register-service.rb:11:in `<main>': Use RbConfig instead of obsolete and deprecated Config.
extra/register-service.rb:15:in `<main>': uninitialized constant CONFIG (NameError)
```

dmitri-d:
@helge000: could you try `register-service` from https://github.com/theforeman/smart-proxy/pull/285?

helge000:
@witlessbird I put both options in for now and try this again tomorrow.

helge000:
@witlessbird pls see my comments on [PR285](https://github.com/theforeman/smart-proxy/pull/285)

domcleal:
Thanks @helge000!  I rebased it to the 1.10 section and merged with a few minor formatting fixes as c8d70d8.

PR #445, opened 2015-10-08
Headline features for 1.10

domcleal:
Do these seem like reasonable things to highlight from http://theforeman.org/manuals/1.10/index.html#Releasenotesfor1.10?  I'm trying to keep to fewer than six.

The only other one might be the partial reorganisation of the Puppet class edit form, and a mention of layout changes on the host edit parameters tab, e.g. to show full parameter names.

Once agreed we can add some screenshots and screencasts.

### DIFF
--- a/_includes/manuals/1.10/1.2_release_notes.md
+++ b/_includes/manuals/1.10/1.2_release_notes.md
@@ -2,6 +2,48 @@
 
 ### Headline features
 
+#### Orchestration rebuilder
+
+A new "Rebuild Config" option is available from the dropdown menu on *All hosts* to rebuild DHCP, DNS and TFTP records and files for one or more selected hosts.
+
+When creating or editing a host, Foreman instructs the smart proxy during orchestration to add DHCP or DNS records to the service (e.g. ISC DHCP server) that it manages.  If the service state is lost, or the server is reinstalled, the reservations and records would be lost.  The rebuild action will force a rewrite of all orchestrated records, by deleting and recreating them.
+
+A corresponding API is available at `PUT /api/hosts/:id/rebuild_config`.
+
+#### Global statuses
+
+The status of a host as shown on the *All hosts* list is now derived from a set of detailed statuses, which can be viewed in the tooltip or from the host page in the *Properties* table.
+
+New detailed statuses have been added for configuration management (Puppet) and build state, and plugins are now able to add their own to this list, allowing them to easily show a host in a warning or error state if desired.
+
+#### Cloning
+
+Host and host group cloning has improved immeasurably in this release, with new features and many bug fixes.  For hosts that are on a compute resource (virtual machines), Foreman now copies the details of the storage and networking for most providers, plus other attributes such as memory, CPUs etc. when creating the new host.  The new host is still installed from scratch (using the selected provisioning method), but should be created with an equivalent configuration.
+
+Bug fixes allow proper cloning of parameters, Puppet class and config group associations, parameter overrides and more, for both hosts and host groups.
+
+A related improvement in image provisioning of new VMware hosts applies networking and storage configuration from the host Virtual Machine tab to the templated VM.
+
+#### Unsetting attributes
+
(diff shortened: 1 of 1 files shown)


_includes/manuals/1.10/1.2_release_notes.md
```diff
@@ -2,6 +2,40 @@
 
 ### Headline features
 
+#### Orchestration rebuilder
```
ohadlevy, line 5: orchestration is only mentioned internally in the code, I wonder if there is a better user facing name for it?

domcleal, line 5: We use it in the manual introduction and the workflow section too.  It seems reasonable to use to describe the process to me.

_includes/manuals/1.10/1.2_release_notes.md
```diff
@@ -2,6 +2,40 @@
...
+
+Some attributes have traditionally been hard to "unset" using the web interface.  When creating a host with a host group, you can now select which attributes (e.g. Puppet master, environment) are used from the host group, or they can be explicitly unset by depressing a new *inherit* button and changing the value.  This allows a host to be created with some attributes from a host group, but excluding others.  Note that hosts still copy the attributes, and will not continue to inherit changes from the host group.
+
+Host root passwords, compute resource, LDAP server password fields etc. now show whether a password is set and using the adjacent edit button, the value can be edited or unset.
+
+#### DNS plugin support
+
+As well as the built in nsupdate and MS AD providers, the Smart Proxy now supports more DNS servers through a new pluggable interface.  Plugins for [PowerDNS](https://github.com/theforeman/smart_proxy_dns_powerdns) and [Route53](https://github.com/theforeman/smart_proxy_dns_route53) are already available, and new plugins can [be written easily](http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Smart-Proxy_Plugin#Adding-a-DNS-provider).
```
ekohl, line 35: Should we aim to have this fully supported in the installer and then mention that it is?

domcleal, line 35: Yes, that'd be ideal.  I don't think I've done any packaging of the route53 plugin yet though!

ekohl, line 35: Given there are no unpackaged dependencies, I don't think that's hard to do.

domcleal, line 35: It needs the route53 gem packaged unfortunately.

_includes/manuals/1.10/1.2_release_notes.md
GregSutcliffe, line 5: +1, I think Orchestration is fine, particularly when coupled with the first paragraph below which clearly explains the intention.

_includes/manuals/1.10/1.2_release_notes.md
```diff
@@ -2,6 +2,40 @@
...
+
+New detailed statuses have been added for configuration management (Puppet) and build state, and plugins are now able to add their own to this list, allowing them to easily show a host in a warning or error state if desired.
+
+#### Cloning
+
+Host and host group cloning has improved immeasurably in this release, with new features and many bug fixes.  For hosts that are on a compute resource (virtual machines), Foreman now copies the details of the storage and networking for most providers, plus other attributes such as memory, CPUs etc. when creating the new host.  The new host is still installed from scratch (using the selected provisioning method), but should be created with an equivalent configuration.
+
+Bug fixes allow proper cloning of parameters, Puppet class and config group associations, parameter overrides and more, for both hosts and host groups.
+
+t related improvement in image provisioning of new VMware hosts applies networking and storage configuration from the host Virtual Machine tab to the templated VM.
```
GregSutcliffe: Typo in column 1?

domcleal: Thanks, fixed.

GregSutcliffe:
Looks good. I think I agree with your selection, the only other things that jump out at me are the LDAP connection test button (v useful to LDAP users, but also pretty obvious) and the Puppet parameters "merge default" stuff. Probably less suitable than the above, but just pointed out for your consideration.

domcleal:
There's no LDAP test feature that's been merged (http://projects.theforeman.org/issues/7537, RfT), and I'll see about adding a bit around parameters and mention the merge feature.

domcleal:
I've added a new parameters section to the headline features, mentioning the merge default option.

domcleal:
Merged as 97373e3.

PR #446, opened 2015-10-08
Re-add blog link to header bar

GregSutcliffe:
Two commits here:

The first simply re-adds the missing link to config.yml so that people can click on it. That should definitely be merged.

The second is debateable - I really dislike the frame wrapper around the Blogger external page. It just looks very odd with both vertical and horizontal scrollbars within the page. I think it's cleaner to redirect to the blog directly. This commit adds an external link redirect layout (that we could also use for other pages) and sets the blog to use it. Seems to work for me.

Worth noting that I have a longer term plan to move the blog to this repo and publish it using Jekyll (subject to discussing this plan with interested parties :P), so whatever we do here is probably temporary. As such, if the consensus is just to merge the first commit, that's fine.


_config.yml
```diff
@@ -23,5 +23,7 @@ navigation:
   url: documentation.html
 - text: Community
   url: contribute.html
+- text: Blog
+  url: blog.html
```
domcleal: Shouldn't this just link straight to the blog URL instead of having the redirect?  (The redirect may be useful for the sake of bookmarks/search engines, but not sure it's needed otherwise.)

GregSutcliffe: I tried putting an external url there, but our layout causes it to be rendered as "http://theforeman.org/http://blog.theforeman.org". I couldn't see an easy way of fixing the loop over site.navigation in the navigation layout, so a separate handler seemed easier.

domcleal:
Could the blog link be on the documentation page?  I don't think it's too bad at the top level, but if support is linked from docs, I don't think it's any more deserving.

_config.yml
domcleal: Something like https://github.com/domcleal/theforeman.org/commit/4c31f31653eccfad73a6132b25739a0dabdc53d6 would do it.

GregSutcliffe:
I think it's as equally deserving as the Videos, rather than the support links - I'd like to see our content promoted well, rather than people having to hunt for it. Both are content forms, and should be promoted equally.

If the issue is to do with the top-bar getting a bit crowded (which I could get behind, especially as I'm considering splitting out Development and Community into two links, a la ovirt.org) then perhaps we could rename to a Content link and move the blog to one of the menu items there?

domcleal:
Yeah, it was the proliferation of links I was concerned about, as I think the addition of the documentation landing page helped us consolidate it once already.

I don't feel it's as deserving as the videos link, because the quality and quantity of content on the blog is quite inferior.  I would prefer for people to be able to find the documentation or videos more easily than the blog, as they're more likely to be current and informative.

That said, it obviously needs linking from somewhere, I'm just not sure where.  I would lean towards the documentation page.  If you added a "Content" page, what would be on it, compared to Documentation?

GregSutcliffe:
That's fair, although more and better content for the blog is very much on my agenda - as techies, I tend to feel we use things we can cut-n-paste from regularly in our work. We've been lax in that area (no fault here, just a lack of time). However, I can't fault your reasoning based on it's _current_ state.

My thought with the previous comment was to rename the Videos link to Content (or Media as it once was, or any other term that represents a variety of useful material about Foreman that people may wish to read/watch/listen). However that doesn't feel quite right either, now that I reflect on it.

Looking around the web, it seems that it's either the blog _is_ the homepage (which I don't think we want), or linked from an About page. On that basis, I think it's fine to put it under Documentation for now, and we can revist once I've improved the blog quality ;)

GregSutcliffe:
Updated as discussed, link is on the Doc page. I've removed the old blog files as they aren't needed for an external link

domcleal:
Thanks!

PR #452, opened 2015-10-13
Update website gems to new GitHub Pages stack

GregSutcliffe:
Seems to work for me. Things checked:
- Manual version selector
- Manual quickstart instructions dropdown
- Video embedding
- Slide deck on intro page
- Plugin manuals
- Support page
- External link to blog

Sadly it doesn't fix the youtube modals on the 1.7/1.8 release notes, but you can't win them all :)

Do we need to do something to replace the :development group in the Gemfile?


domcleal:
I don't think the Bundler group matters for anything, can't think why we'd need it, removing it's fine.

Gemfile.lock
```diff
@@ -2,71 +2,131 @@ GEM
   remote: http://rubygems.org/
```
mmoll: I have no idea about the GH stack but is it really required to commit Gemfile.lock into this repo?

domcleal: It isn't required.  I guess it's just an argument about whether we want to pin the website tightly to a set of versions (either via Gemfile.lock or a version pin in Gemfile) so we don't get any surprises when the website is built.

That said, the PR seems to demonstrate that the stability and compatibility of the stack between 2013 and 2015 is excellent.  It's likely that GitHub would advertise if there was an upcoming incompatible change.

GregSutcliffe: From what I saw in testing, deprecation of features seems well handled, and we test locally with Jekyll before push anyway. I have no strong opinion either way, so I'm happy to drop the Gemfile.lock in this PR if need be.

mmoll: I see... it's just always the first step for me to delete Gemfile.lock when working with packaged rubygems/jekyll :)

domcleal: Yeah, let's drop it then.

GregSutcliffe: Done

GregSutcliffe:
Seond question: should we pin GHP?

domcleal:
I don't think so, I think the conclusion from the Gemfile.lock discussion is that this stack is stable enough to not require pinning or locks.

GregSutcliffe:
:+1:

mmoll:
:+1:

domcleal:
Seems we have a similar issue to core:

<pre>
12:04 $ rake server
jekyll serve --watch
/home/dcleal/.rvm/gems/ruby-2.0.0-p247@theforeman.org/gems/execjs-2.6.0/lib/execjs/runtimes.rb:48:in `autodetect': Could not find a JavaScript runtime. See https://github.com/rails/execjs for a list of available runtimes. (ExecJS::RuntimeUnavailable)
        from /home/dcleal/.rvm/gems/ruby-2.0.0-p247@theforeman.org/gems/execjs-2.6.0/lib/execjs.rb:5:in `<module:ExecJS>'
        from /home/dcleal/.rvm/gems/ruby-2.0.0-p247@theforeman.org/gems/execjs-2.6.0/lib/execjs.rb:4:in `<top (required)>'
</pre>

Can I suggest we follow the same pattern and have a therubyracer group, which people can disable if they like?  https://github.com/theforeman/foreman/blob/develop/bundler.d/therubyracer.rb

domcleal:
Gemfile.lock also needs adding to .gitignore.

mmoll:
:+1: for a therubyracer group

GregSutcliffe:
Sure, do we need to version it as per core, or just a bare gemname in this group?

domcleal:
I don't think the version's important, just the gem name and 'require'.

GregSutcliffe:
Updated

domcleal:
Thanks @GregSutcliffe, seems to work.

PR #456, opened 2015-10-21
Missing </div> in documentation

dLobatog:
**Problem**
A missing </div> in documentation is lost and misaligns the icons.

**Solution**
Put the closing div in there.


domcleal:
Thanks!

PR #459, opened 2015-10-23
Document Smart Proxy 1.10 settings

domcleal:
- DNS providers and split config files
- DHCP server setting
- SYSLOG log file setting

### DIFF
--- a/_includes/manuals/1.10/4.3.5.3_gsstsig.md
+++ b/_includes/manuals/1.10/4.3.5.3_gsstsig.md
@@ -53,11 +53,13 @@ Note the `\047` is written verbatim, and don't forget the semicolon.  ACLs shoul
 
 #### Proxy configuration
 
-Next, update the proxy configuration file (`/etc/foreman-proxy/settings.d/dns.yml`) with the following settings:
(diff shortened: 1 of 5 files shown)


_includes/manuals/1.10/4.3.5.3_gsstsig.md
```diff
@@ -53,11 +53,11 @@ Note the `\047` is written verbatim, and don't forget the semicolon.  ACLs shoul
...
 
     :dns_provider: nsupdate_gss
+
+And the DNS GSS configuration file (`/etc/foreman-proxy/settings.d/dns_nsupdate_gss.yml`) with:
+
     :dns_tsig_keytab: /etc/foreman-proxy/dns.keytab
     :dns_tsig_principal: foremanproxy/proxy.example.com@EXAMPLE.COM
-    #:dns_key: false
-
-To work correctly, `:dns_key:` must be commented out.
```
mmoll, line 63: not sure if that's still the case, but a note that dns_key must not be set might still be in order

domcleal, line 63: Ah yes, I think it could still affect it if specified.  Thanks, I've added a new comment.

mmoll:
thanks!

PR #469, opened 2015-11-03
Use old-style permalink for old G+ comments

GregSutcliffe:
Turns out you _can_ test the comment system locally - it just uses the href in the comments widget. So, this adds a conditional - anything imported from Blogger has a blogger-id in the frontmatter, so we know to use the old-style URL for comments.

### DIFF
--- a/_layouts/post.html
+++ b/_layouts/post.html
@@ -45,7 +45,11 @@
 
       <hr/>
       <div class="g-comments"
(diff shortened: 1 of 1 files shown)


GregSutcliffe:
@domcleal this should drop the need for a CNAME, I think - we just need redirects for the old URLs.

domcleal:
Thanks @GregSutcliffe!

PR #470, opened 2015-11-04
blog: announce the new blog site

GregSutcliffe:
Once reviewed/merged I'll port it to the old site as well, which should be the last blog there. That will just leave the redirects and niggles for later today/this week.

### DIFF
--- a/_posts/2015-11-05-presenting_the_new_foreman_blog.md
+++ b/_posts/2015-11-05-presenting_the_new_foreman_blog.md
@@ -1,0 +1,72 @@
+---
+layout: post
+title: Presenting the new Foreman Blog!
+date: 2015-11-05 09:25:28
+author: Greg Sutcliffe
+tags:
+- foreman
+---
+
+Some while ago, I realized that the entire Foreman website was written in
+Markdown and stored in GitHub, where any contributor can help to improve it -
+but not the blog! An area almost entirely composed of text; an area where guest
+posts would be hugely valuable; an area where fixes to old popular posts would
+be fantasic; all written on Blogger, where we can't easily get contribution
+from the community. Sigh.
+
+Well, no longer! Today we're unveiling the new Foreman Blog!
+
+<!--more-->
+
+## What's new?
+
+In terms of content - very little. The old posts have been imported from
+Blogger via their XML export tool, and then converted to Markdown via the
+awesome [Pandoc](pandoc.org) tool. Images were downloaded from blogger too, and
+Gists loaded via the *jekyll-gist* gem which was already part of the website
+stack. However, the vast majority of the text is unchanged.
+
+The URL has changed though - it's now
+[http://theforeman.org/blog]({{site.baseurl}}blog)
+
+## What's changed?
+
+We've tried pretty hard to be consistent. Obviously the style of the pages now
+comes under the general styling of the overall Foreman website, so that's a
+plus. Later this week we'll redirect *blog.theforeman.org* to *theforeman.org*
+so that the permalinks stored in search engines will continue to work.
+
+There is an RSS feed, of course. You can get it at
+[/feed.xml]({{site.baseurl}}feed.xml) and there's an RSS icon at the top
+of the blog index. Please do update your RSS readers if you're using the old Blogger
+feed.
+
+Post comments appear to be present and correct, so there should be no
+changes or continuity loss from Blogger.
+
+## Great! How can I help?
+
+As part of my role as the new Community Lead, I'll definitely be posting more
+about all sorts of Foreman related things, as well as my own activities to keep
+the community informed. In addition, the main goal for changing to Markdown was
+to enable community contribution, so ...
+
+If you see any bugs or problems with any old blog posts, please do let us know
+in the [issue tracker](https://github.com/theforeman/theforeman.org/issues) or
+even better, open a Pull Request to fix the problem (see
+[Contributing]({{site.baseurl}}contribute.html#Documentationandwebsite)
+for a guide on that). You can find the raw text of the posts
+[here](https://github.com/theforeman/theforeman.org/tree/gh-pages/_posts).
+
+If you want to rework an old blog post for a newer version of Foreman, please
+feel free to do so - many of the older blog posts (such as the one on EC2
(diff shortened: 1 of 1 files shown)


_posts/2015-11-04-presenting_the_new_foreman_blog.md
```diff
@@ -0,0 +1,72 @@
...
+
+As part of my role as the new Community Lead, I'll definitely be posting more
+about all sorts of Foreman related things, as well as my own activities to keep
+the community informed. In addition, the main goal for changing to Markdown was
+to enable community contribution, so ...
+
+If you see any bugs or problems with any old blog posts, please do let us know
+in the [issue tracker](https://github.com/theforeman/theforeman.org/issues) or
+even better, open a Pull Request to fix the problem (see
+[Contributing](http://theforeman.org/contribute.html#Documentationandwebsite)
```
domcleal: `{{site.baseurl}}`

_posts/2015-11-04-presenting_the_new_foreman_blog.md
```diff
@@ -0,0 +1,72 @@
...
+## What's new?
+
+In terms of content - very little. The old posts have been imported from
+Blogger via their XML export tool, and then converted to Markdown via the
+awesome [Pandoc](pandoc.org) tool. Images were downloaded from blogger too, and
+Gists loaded via the *jekyll-gist* gem which was already part of the website
+stack. However, the vast majority of the text is unchanged.
+
+The URL has changed though - it's now
+[http://theforeman.org/blog](http://theforeman.org/blog)
```
domcleal: ditto, and feed.xml

_posts/2015-11-04-presenting_the_new_foreman_blog.md
GregSutcliffe: Fixed.

_posts/2015-11-04-presenting_the_new_foreman_blog.md
GregSutcliffe: Fixed, thanks.

_posts/2015-11-04-presenting_the_new_foreman_blog.md
```diff
@@ -0,0 +1,72 @@
...
+## What's new?
+
+In terms of content - very little. The old posts have been imported from
+Blogger via their XML export tool, and then converted to Markdown via the
+awesome [Pandoc](pandoc.org) tool. Images were downloaded from blogger too, and
+Gists loaded via the *jekyll-gist* gem which was already part of the website
+stack. However, the vast majority of the text is unchanged.
+
+The URL has changed though - it's now
+[http://theforeman.org/blog]({{site.baseurl}}/blog)
```
domcleal: Remove the extra `/`, it'll break the link.

_posts/2015-11-04-presenting_the_new_foreman_blog.md
```diff
@@ -0,0 +1,72 @@
...
+
+As part of my role as the new Community Lead, I'll definitely be posting more
+about all sorts of Foreman related things, as well as my own activities to keep
+the community informed. In addition, the main goal for changing to Markdown was
+to enable community contribution, so ...
+
+If you see any bugs or problems with any old blog posts, please do let us know
+in the [issue tracker](https://github.com/theforeman/theforeman.org/issues) or
+even better, open a Pull Request to fix the problem (see
+[Contributing]({{site.baseurl}}/contribute.html#Documentationandwebsite)
```
domcleal: ditto, extra `/`

lzap:
Side note: Links in our feed have extra slash in them. It's worth fixing _before_ we announce the blog (the fix might duplicate all the records in user's readers).

Example link: http://theforeman.org//2015/10/foreman-community-newsletter-october.html

domcleal:
I think duplication is inevitable as the IDs will have changed domain, but sure, now is a good time to fix it.

GregSutcliffe:
Thanks for the typos.

Re the double slashes - jekyll-feed is doing that (here: https://github.com/jekyll/jekyll-feed/blob/98cd38e7220b7e692f7f0f1186228a5673267adf/lib/feed.xml#L3). I don't see a way to configure it, other than to drop `baseurl`, but we use that extensively in the manual... perhaps we could rename it to a custom variable for the manual?

domcleal:
I think it has a particular meaning in Jekyll, not just a custom-named variable.  It's set or used when you run the server on a non-root URL too, e.g. on GitHub.  I'd be very wary about messing with it.

GregSutcliffe:
Fair enough. In that case I don't see that we have much option but to leave the double slash in there then :(

GregSutcliffe:
Interestingly the [default Jekyll config](http://jekyllrb.com/docs/configuration/#default-configuration) seems to use `baseurl: ""`. I tried it locally, and as @domcleal says, it broke a lot of things, however, [this commit](https://github.com/GregSutcliffe/theforeman.org/commit/0e801bec4b8cd6b3aba0114db7b89662c4b075c0) seems to resolve all the issues I could find. Thoughts?

GregSutcliffe:
Scratch that, I was over-complicating as usual. [This page](http://jekyllrb.com/docs/github-pages/#project-page-url-structure) points out using "{{site.baseurl}}/" for includes, which seem to be the only broken thing. See [this commit](https://github.com/GregSutcliffe/theforeman.org/commit/b93092f53627a0528c4dc5d532369b24e46a45b3) for my attempt.

EDIT: the point being, that `baseurl` shouldn't ever have a trailing slash - it's "" or "/suburl"

domcleal:
Yes, I think you're right about the trailing slash.  It seems quite safe to remove because we have the `<base>` element.  Both were added in #17.

Suggest you make your commit into a PR?

GregSutcliffe:
@domcleal cool, PR #471 submitted. I assume we'll depend on that here, to get the feed right.

GregSutcliffe:
Date updated since we didn't get this out yesterday. I'd _really_ like to get this out today if we can, so I can check if the G+ comments work on new posts.

domcleal:
Thanks, both merged.

PR #472, opened 2015-11-05
Use page.title for blog posts in header

GregSutcliffe:
Blog posts don't seem to have `pagename` set, so the page title is just "Foreman". Duplicating `title` to `pagename` if the frontmatter of each post seems like a bad idea, so this just adds a conditional to all use of `title` for blog posts.


domcleal:
Thanks!

PR #475, opened 2015-11-06
Make Guest Posts reposted from community authors stand out easily

GregSutcliffe:
If we're going to solicit guest posts, they should be obvious to readers. This adds a consistently-styled Guest Post badge to the index and the post page, and automatically adds the "authorized repost" found in @dLobatog's posts, if a url is provided.

The CSS might be overkill, but I didn't see anything suitable to inherit from - the other styling inherits from `btn` and this isn't a button.


dLobatog:
Neat, that's good :smile:

dLobatog:
Merged as 27c08f5d113741bc46ea6ea161397d02c033f157, thanks @GregSutcliffe

PR #478, opened 2015-11-12
Add 1.10 highlight preview links

GregSutcliffe:
Sound levels seem better to me this time. The parameters one is 48s which is a tad long, but I couldn't see anything else worth cutting. Suggestions welcome if others think it needs to be cut down further.


GregSutcliffe:
YouTube links are currently unlisted. I'll make them public once this is merged.

domcleal:
Looks good, thanks!

PR #483, opened 2015-11-19
 The option --foreman-configure-ipa-repo is no longer needed

veruu:
The option --foreman-configure-ipa-repo is no longer needed since the
packages are included in RHEL/CentOS versions 6 and 7 now.

This feature was needed in older RHEL/CentOS versions, in which some
packages weren't available. Now there is no need for it.

The option is already removed from 1.10 installer.


mmoll:
merged, díky @veruu!

Issue #395, opened 2015-07-01
Foreman 1.10 manual


domcleal:
Usual stuff:
- [x] Headline features (#445)
- [x] Headline feature screencasts
- [x] Itemised release notes (up to 8th Oct)
- [ ] Foreman setting changes (#509)
- [x] Smart proxy setting changes (#459)
- [x] Installer options refresh (#508)
- [x] Contributors list
- [x] Translator contributors
- [x] Installer module release notes
- [x] API documentation per README (http://projects.theforeman.org/issues/12778)

1.10 specifics:
- [x] #10959 - Let users know gravatar is now off by default and to turn it on if they want it back
- [x] #11012 - update menu references from "Puppet classes" to "Classes"
- [x] #10356 - API POST/create responses now 201, same attrs as GET
- [x] #4419 - Puppet class editing UI refresh (#448)
- [ ] #9591 - Added support for "inherit" state. (#512)
- [x] #11413 - SQL logger default now false
- [x] #4841 - ruby193 references change to tfm
- [x] #11323 - smart proxy syslog configuration
- [x] #10731 - smart matchers can now merge defaults
- [x] #11449 - nodejs/therubyracer compiler options when installing from source
- [x] #11676 - smart proxy :dhcp_server option
- [ ] #7167 - settings now have a more user-friendly name in the UI  (#509)
- [ ] #8979 - upgrades of Hammer (only) will require SCL packages to be available (#510)

ares:
would it make sense to add
- [x] Global status and substatus descriptions (#515, originally #511)

maybe new section 4.9 Monitoring in 4. General Foreman

domcleal:
Sounds good!

domcleal:
Closing as 1.10 has been released.  The remaining unfinished items are filed as new tickets, linked above.

PR #486, opened 2015-11-20
mention therubyracer bundler group in dev setup

mmoll:
together with 364b162 this should suffice for the `#11449` point in #395


domcleal:
Thanks @mmoll!

PR #491, opened 2015-12-03
Add case studies to media page

dLobatog:
Case studies seem to me like the most relevant videos for people
wondering whether to use Foreman or not, more than talks, demos or
deep dives, hence why I put them on top.

I use a youtube embed playlist for this, which I suggest we use for the
rest of the videos, or at least for community demos and deep dives. It's
a rather painless way to get this page auto updated without any updates
to this repository.

The only reason why I haven't done deep dives and community demos in a
playlist too is because our playlists in the Foreman channel only accept
videos within that channel. We would have to disable (Set as official
series for this playlist) on the channel to allow videos from outside.

### DIFF
--- a/_layouts/media.html
+++ b/_layouts/media.html
@@ -9,6 +9,13 @@ pagename: Media
 			</div>
 		</div>
 		<div id="doc" class="col-md-9">
+			<div id='case_studies'>
+				<h1>Case studies</h1>
+				<iframe width="640" height="360"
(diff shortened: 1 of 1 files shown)


GregSutcliffe:
I think we can unset that option for the demos/deepdives - I don't really see what benefit it brings to us anyway. I've merged this anyway as it's useful as-is, we can do the other two once the playlists are updated.

Thanks @dLobatog, merged as d8e5728!

PR #501, opened 2015-12-28
Change 'current versions' by last versions

dLobatog:
Current version would be fine if there was just one version listed.
However, with two versions listed on the homepage, it's unclear
which one is actually the most current. I think 'last' or
'supported' versions would be more accurate

### DIFF
--- a/_includes/version.html
+++ b/_includes/version.html
@@ -1,6 +1,6 @@
 <div class="version-marker">
   <p>
-    Current versions:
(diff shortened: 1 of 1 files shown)


domcleal:
"Latest" would read slightly better than "Last" I think, is that OK?

dLobatog:
@domcleal Updated, thanks!

domcleal:
Thanks!

PR #508, opened 2016-01-14
Refresh installer options docs

domcleal:
Regenerated the automatic docs from the parameters, and updated the list
of modules and links to projects.


GregSutcliffe:
Merged as 282268e, thanks!

PR #524, opened 2016-01-25
Document our current bug and feature requrest raising process

GregSutcliffe:
This is what I _believe_ our current process is, but we've never actually documented it, so I thought I'd add it. I'll fire the PR over to the dev-list for further comments.

### DIFF
--- a/contribute.md
+++ b/contribute.md
@@ -81,7 +81,7 @@ Patches to fix bugs or add new features are always appreciated. If you are going
 
 We try to keep a one commit per bug/feature policy, please try to create an issue which is specific for your patch details.
 
-Please make sure there is a [Redmine issue](http://projects.theforeman.org/projects/foreman/issues) open for the change you are going to submit, as you will want to reference it in your commit message; this is very helpful when generating release notes.
+Please make sure there is a [Redmine issue](/contribute.html#Bugreporting) open for the change you are going to submit, as you will want to reference it in your commit message; this is very helpful when generating release notes.
 
 * Create a feature/topic branch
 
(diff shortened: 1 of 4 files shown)


contribute.md
```diff
@@ -121,8 +121,22 @@ Helping out other users in the "Forums" is always useful. Frequent problems or q
 
 Testing is also *very* welcome, for any issue encountered, please open a bug / feature request.
 
+### Bug reporting
+Even the simplest of bugs reported helps us make the project better. The issue tracker is located at [http://projects.theforeman.org/projects/foreman/issues](http://projects.theforeman.org/projects/foreman/issues), and you should follow these guidelines:
```
domcleal, line 133: This links to the Foreman project specifically, so I'd add a bullet point below to ensure that the correct project is chosen first.

domcleal:
The sections above that specify a Redmine ticket is required should link internally to this new section I think.

contribute.md
GregSutcliffe, line 133: Added, thanks

GregSutcliffe:
Yeah, that makes sense, i'll grep it out and update shortly.

GregSutcliffe:
@domcleal (and anyone else), i've also updated the plugins page and, but with two different styles (one is replacing the issues/new/ link with the new section, the other adds the new link as further text). Which do you prefer? I think I prefer replacing over adding, but I'm open to arguments. I can then update all of them to be consistent.

contribute.md
```diff
@@ -134,7 +149,7 @@ We're trying to maintain high quality, authoritative documentation in the Forema
 Both the manual and the web site are contained in our [theforeman.org repository](https://github.com/theforeman/theforeman.org).  See the README.md to get started.  Contributions to this project are licensed under Creative Commons Attribution-ShareAlike 3.0.
 
 ### Design and User interface
-Foreman has a very basic design and user interface. Any improvements to it including new themes, skinning, or interface adjustments could help every user.
+Foreman has a very basic design and user interface. Any improvements to it including new themes, skinning, or interface adjustments could help every user. Design discussion should happen on [the dev mailing list](/support.html#Mailinglists).
```
dLobatog: Nitpick, the first sentence has been on these docs since the early days when Foreman was all 'blue' (dunno how to call that nor I do remember the version, 0.9?), I don't think it holds true anymore? 
I'd also mention we are trying to follow Patternfly guidelines (http://patternfly.org/) so any help with that would be appreciated

dLobatog:
I'd be fine with moving all of it to the handbook - my initial thought was that our handbook would be more or less like this https://playbook.thoughtbot.com/ , a single point of reference for all project related processes

domcleal:
> @domcleal (and anyone else), i've also updated the plugins page and, but with two different styles (one is replacing the issues/new/ link with the new section, the other adds the new link as further text). Which do you prefer? I think I prefer replacing over adding, but I'm open to arguments. I can then update all of them to be consistent.

Looks fine as it is, since the plugin link is linking to a specific subproject.

contribute.md
GregSutcliffe: Fixed, check you like my wording :)

GregSutcliffe:
@domcleal fair enough, I wasn't sure if we should go for consistency,

@dLobatog :+1: but if it's ok with you, let's do that in a new PR? I'll raise an issue.

contribute.md
```diff
@@ -134,7 +149,7 @@ We're trying to maintain high quality, authoritative documentation in the Forema
 Both the manual and the web site are contained in our [theforeman.org repository](https://github.com/theforeman/theforeman.org).  See the README.md to get started.  Contributions to this project are licensed under Creative Commons Attribution-ShareAlike 3.0.
 
 ### Design and User interface
-Foreman has a very basic design and user interface. Any improvements to it including new themes, skinning, or interface adjustments could help every user.
+Foreman 1.11 and above uses [Patternfly](http://patternfly.org/) as it's base design. Any improvements or suggestions on how implement this better, or on re-implementing particular pages are very welcome, and could help every user. Design discussion should happen on [the dev mailing list](/support.html#Mailinglists).
```
ehelms: s/how implement/how to implement

GregSutcliffe: Bleah, ta!

aaschofield:
@GregSutcliffe As a recent starter in terms of submitting PR's. What is the policy regarding actually fixing things? You mention there should be a redmine issue for something you want to work on. But what if I want to work on an (unassigned) issue where there is a redmine for? Do we just assign them to ourselves?

GregSutcliffe:
@aaschofield In general, yes. The system will auto-assign a ticket to a user when a pull request comes in, if it can figure out the mapping, but if you want to mark a ticket in advance, feel free to assign it.

This is vaguely mentioned under http://theforeman.org/contribute.html#SubmitPatches but maybe it could be clearer?

GregSutcliffe:
@dLobatog logged #527 for the merge later - any other blockers on this?

contribute.md
```diff
@@ -134,7 +149,7 @@ We're trying to maintain high quality, authoritative documentation in the Forema
 Both the manual and the web site are contained in our [theforeman.org repository](https://github.com/theforeman/theforeman.org).  See the README.md to get started.  Contributions to this project are licensed under Creative Commons Attribution-ShareAlike 3.0.
 
 ### Design and User interface
-Foreman has a very basic design and user interface. Any improvements to it including new themes, skinning, or interface adjustments could help every user.
+Foreman 1.11 and above uses [Patternfly](http://patternfly.org/) as it's base design. Any improvements or suggestions on how to implement this better, or on re-implementing particular pages are very welcome, and could help every user. Design discussion should happen on [the dev mailing list](/support.html#Mailinglists).
```
domcleal: "its base design"

dLobatog:
:+1: from me for the changes here, I didn't look in other pages but this ought to be fine.

GregSutcliffe:
@domcleal fixed

domcleal:
Thanks @GregSutcliffe.

PR #529, opened 2016-01-30
Details on ssh finish step

agx:


mmoll:
Looks reasonable, but as I'm only using PXE, I can't say if the content is correct.. anybody?

ares:
sounds good, I'd only make a note that there is a second way using user-data template (if the image supports cloud init) and the log file seems to be stored in /tmp/ (which was probably meant by "same directory")

agx:
Hi,
On Mon, Feb 08, 2016 at 12:41:00AM -0800, Marek Hulán wrote:

> sounds good, I'd only make a note that there is a second way using user-data template (if the image supports cloud init) and the log file seems to be stored in /tmp/ (which was probably meant by "same directory")

I've added a small paragraph about user_data. Can't really expand on
this since I'm not using it atm.

Note that the logs are not in /tmp/ anymore but in ~ of the user
connecting.

Cheers,
 -- Guido

_includes/manuals/1.10/4.4.3_prov_templates.md
```diff
@@ -40,3 +40,10 @@ The final entry, Operating System default, can be set by editing the Operating S
...
+
+##### Templates Details
+###### Finish Templates
+Finish templates are one of two methods to customize and finish the installation for image based installs, the other alternative being user_data templates. In contrast to user_data they are available for all hypervisors that support image based installs where the foreman server can reach the newly installed machine via ssh and scp. The script generated from the finish template is copied by the Foreman to the newly installed system via scp using username and password specified with the image. It is then executed by connecting again via ssh, making the script executable and either executing it directly or via sudo if the specified user is not root. Standard output and standard error are logged to a file in the same directory named *bootstrap-UUID.log*.
+
+###### User Data Templates
+User Data templates are one of two methods to customize and finish the installation for image based installs, the other alternative finish templats. In this case the installed machine must not be reachable via ssh by the Foreman server but the machine must be able to reach the host via http(s).
```
mmoll: maybe it makes more sense to move the "the other alternative" parts to a very short general introduction in "Templates Details"

dLobatog: typo - 'templats' instead of 'templates'. 

"the installed machine does not need to be reachable" reads a bit better I'd say. "must not be reachable" sounds like the if the host is reachable by Foreman user-data won't work. 

I don't think http/https has to do anything at all with user-data. 'these templates are run by cloud-init to handle early initialization of a cloud instance' is enough if you're not familiar with it. Foreman sends these templates to the compute resource in question (via fog) when the machine is created, the machine might not even have a public IP but it'll run these templates.

dLobatog:
@agx I made a small comment about user-data inline, please take a look, update the PR and I'll be glad to merge. Thanks for the docs :smile: !

_includes/manuals/1.10/4.4.3_prov_templates.md
agx: @dLobatog  isn't http/https needed to ping back to foreman that the install finished like in:

```
<%# Contact Foreman to confirm instance is built -%>
phone_home:
 url: <%= foreman_url('built') %>
 post: []
 tries: 10
```

to notify Foreman that the instance is built? Example is from https://github.com/theforeman/community-templates/blob/develop/cloudinit/userdata_cloudinit.erb

_includes/manuals/1.10/4.4.3_prov_templates.md
```diff
@@ -40,3 +40,11 @@ The final entry, Operating System default, can be set by editing the Operating S
...
+
+##### Templates Details
+For image based installs there are two methods to customize and finish the installation. *Finish templates* and User *Data Templates*:
+###### Finish Templates
+Finish templates are available for all hypervisors that support image based installs where the foreman server can reach the newly installed machine via ssh and scp. The script generated from the finish template is copied by the Foreman to the newly installed system via scp using username and password specified with the image. It is then executed by connecting again via ssh, making the script executable and either executing it directly or via sudo if the specified user is not root. Standard output and standard error are logged to a file in the same directory named *bootstrap-UUID.log*.
+
+###### User Data Templates
+User Data Templates are available for hypervisors that support customization via tools like *cloud-init*. In this case the installed machine does not need be reachable via ssh by the Foreman server but the machine must be able to reach the host via http(s) to notify the Foreman that it finished the setup.
```
dLobatog, line 50: 'the machine must be able to reach the host' is very ambiguous, which machine is Foreman? I take it it's the host, I'll change this on merge to 'the installed machine must be able to reach Foreman or a proxy to notify the setup is finished'

dLobatog:
Merged as 0d8f0d1702c83b78714fb645ca88d26bd9f52989 , thanks @agx

PR #530, opened 2016-02-02
limit gem-pages version to one that used jekyll 2

mmoll:
github-pages 44 is bringing jekyll 3 and this page is not yet ready for it.


domcleal:
Thanks @mmoll.

PR #532, opened 2016-02-03
Specify that the deployed host will get its template from 80/443

ghost:


domcleal:
Thanks, would you mind making the same change on port 3000 below too?  It (WEBrick) is not the default installation, but works equally to when using Passenger in that it's used for provisioning templates.

ghost:
Thanks for reviewing

domcleal:
Thanks @SkySymbol!

PR #541, opened 2016-02-10
Remove explicit HTTP protocols from embedded content

domcleal:
Will prevent mixed content issues when the site's served over HTTPS.


ekohl:
:+1:

GregSutcliffe:
Merged as 13e8a9c, thanks @domcleal

PR #546, opened 2016-02-24
Offer different plugin install instructions per Foreman version

dLobatog:
I guess we should make this a snippet and be able to put it in all plugins. I found problems with these 3, but I'm sure there are more. Sorry for not doing that just yet - i'm submitting this to avoid people looking at these instructions and getting mad when they don't work

### DIFF
--- a/plugins/foreman_cockpit/1.0/index.md
+++ b/plugins/foreman_cockpit/1.0/index.md
@@ -42,6 +42,10 @@ To install Foreman Cockpit, run the following:
 
       # yum install tfm-rubygem-foreman_cockpit
 
+or this if your Foreman version is 1.9 or older
+
+      # yum install ruby193-rubygem-foreman_cockpit
+
 * Fedora
 
       # yum install rubygem-foreman_cockpit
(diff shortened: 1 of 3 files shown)


plugins/foreman_cockpit/1.0/index.md
```diff
@@ -42,6 +42,10 @@ To install Foreman Cockpit, run the following:
 
       # yum install tfm-rubygem-foreman_cockpit
 
+or this if your Foreman version is 1.9 or older
```
domcleal, line 45: missing bullet point at the start of these lines I think, the formatting looks inconsistent with the others

plugins/foreman_openscap/0.5/index.md
```diff
@@ -82,8 +82,25 @@ Foreman OpenSCAP (foreman_openscap), Smart Proxy OpenSCAP (smart_proxy_openscap)
...
+
+To install Foreman OpenSCAP, run the following:
+
+* Red Hat based (RHEL, CentOS...)
+
+      # yum install tfm-rubygem-foreman_openscap
+
+or this if your Foreman version is 1.9 or older
+
+      # yum install rubygem193-rubygem-foreman_openscap
```
domcleal, line 97: ruby193

mmoll:
this needs a rebase and the comments above addressed :)

ehelms:
@dLobatog planning to revisit this? Otherwise I think we should just close it out.

mmoll:
closing after inactivity.

PR #550, opened 2016-02-25
JSON with releases

xprazak2:
JSON file with info about released versions gets generated by Jekyll. Its content is based on already existing docs but we can add whatever we like to the misc.yaml. For now, we have core versions, latest releases of core and versions of plugins that have docs on wiki.

### DIFF
--- a/_data/releases.yml
+++ b/_data/releases.yml
@@ -1,0 +1,833 @@
+---
+core:
+  1:
+    12:
+    - 1.12.0-RC1
+    11:
+    - 1.11.2
+    - 1.11.1
+    - 1.11.0-RC3
+    - 1.11.0-RC2
+    - 1.11.0-RC1
+    10:
+    - 1.10.4
+    - 1.10.3
+    - 1.10.2
+    - 1.10.1
+    - 1.10.0
+    - 1.10.0-RC3
+    - 1.10.0-RC2
+    - 1.10.0-RC1
+    9:
+    - 1.9.3
+    - 1.9.2
+    - 1.9.1
+    - 1.9.0
+    - 1.9.0-RC3
+    - 1.9.0-RC2
+    - 1.9.0-RC1
+    8:
+    - 1.8.4
+    - 1.8.3
+    - 1.8.2
+    - 1.8.1
+    - 1.8.0
+    - 1.8.0-RC3
+    - 1.8.0-RC2
+    - 1.8.0-RC1
+    7:
+    - 1.7.5
+    - 1.7.4
+    - 1.7.3
+    - 1.7.2
+    - 1.7.1
+    - 1.7.0
+    - 1.7.0-RC2
+    - 1.7.0-RC1
+    6:
+    - 1.6.3
+    - 1.6.1
+    - 1.6.0
+    - 1.6.0-RC2
+    - 1.6.0-RC1
+    5:
+    - 1.5.3
+    - 1.5.2
+    - 1.5.1
+    - 1.5.0
+    - 1.5.0-RC2
+    - 1.5.0-RC1
+    4:
+    - 1.4.5
+    - 1.4.4
+    - 1.4.2
+    - 1.4.1
+    - 1.4.0
+    - 1.4.0-RC2
+    - 1.4.0-RC1
+    3:
+    - 1.3.2
+    - 1.3.1
+    - 1.3.0
+    - 1.3.0-RC4
+    - 1.3.0-RC3
+    - 1.3.0-RC2
+    - 1.3.0-RC1
+    2:
+    - 1.2.0-RC2
+    - 1.2.0-RC1
+    - 1.2.3
+    - 1.2.2
+    - 1.2.1
+    - 1.2.0
+    - 1.2.0-RC3
+    0:
+    - 1.0.1
+  0:
+    4:
+    - 0.4.2
+    - 0.4.1
+    1:
+    - 0.1.1
+    - 0.1.1RC5
+    - 0.1.1RC4
+    - 0.1.1RC3
+    - 0.1.1RC2
+    - 0.1.1RC1
+    - 0.1.1-1
+    - 0.1.0
+    - 0.1.0RC5
+    - 0.1.0RC4
+    - 0.1.0RC3
+    - 0.1.0RC2
+    - 0.1.0RC1
+    0:
+    - 0.0.4
+    - 0.0.4rc5
+    - 0.0.4rc4
+    - 0.0.4rc3
+    - 0.0.4rc2
+    - 0.0.3
+    - 0.0.2
+    - 0.0.2rc1
+    - 0.0.1
+    - 0.0.1-6
+    - 0.0.1-5
+    - 0.0.1-4
+    - 0.0.1-3
+    - 0.0.1-2
+    - 0.0.1-1
+plugins:
+  foreman-tasks:
+    0.7.15:
+      requires_foreman: '1.9'
+    0.7.14:
+      requires_foreman: '1.9'
+    0.7.13:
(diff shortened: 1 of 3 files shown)


_plugins/versions_generator.rb
```diff
@@ -0,0 +1,41 @@
+require 'json'
+require 'yaml'
+
+module Jekyll
+  class VersionsGenerator < Generator
```
domcleal: Custom plugins can't be run on GitHub Pages, I don't think it's a good idea to add one.

xprazak2:
I removed the generator, now we have only static files.

_data/misc.yaml
```diff
@@ -0,0 +1,45 @@
+releases:
```
domcleal: This file should probably be called releases, since that's what it contains.

_data/misc.yaml
```diff
@@ -0,0 +1,45 @@
...
+  core:
+    latest:
+      1-11: 1.11.0-RC1
+      1-10: 1.10.2
+      1-9: 1.9.3
+    all:
```
domcleal: What's this list for?  Should it contain pre-releases?

_data/misc.yaml
```diff
@@ -0,0 +1,45 @@
...
+      - "1.3"
+      - "1.4"
+      - "1.5"
+      - "1.6"
+      - "1.7"
+      - "1.8"
+      - "1.9"
+      - "1.10"
+      - "1.11"
+  plugins:
```
domcleal: This list's incomplete, is it only going to contain plugins with documentation in this repo?

_data/misc.yaml
```diff
@@ -0,0 +1,45 @@
...
+    foreman_ansible:
+      - "0.x"
```
domcleal: What's the format of these fields?  It seems limited to only major and/or minor versions, but the Foreman releases above contains patch revisions.

I'm not sure how it'd be maintained if patch revisions _were_ added though.

lzap: +1 Discovery does a lot of minor revisions, for example the first official release for 1.11 RC will be 5.0.1. We should be able to list them all there. Hopefully Foreman shows the incoming plugin updates to users too.

lzap: By the way you can get the lists from https://rubygems.org/gems/foreman_discovery/versions

_includes/version.html
```diff
@@ -1,8 +1,8 @@
 <div class="version-marker">
   <p>
     Latest versions:
-      <a href="{{ site.baseurl }}documentation.html#1.11">1.11.0-RC1</a>
+      <a href="{{ site.baseurl }}documentation.html#1.11">{{ site.data.misc.releases.latest.1-11 }}</a>
```
lzap: Aren't you missing `core` in the variable name?

_data/misc.yaml
domcleal: The other complication with plugins is that not all updates are applicable to all versions.  The "latest" plugin release might only be available in a nightly or later distribution, but the user would have to update Foreman before it makes sense to install it.

lzap:
In manual and plugin layouts we use some tricks to show warning boxes for old or unreleased documentation. Since you introduce now this data, this could be used to actually show these boxes as well if you provide "latest" versions for all the plugins.

https://github.com/theforeman/theforeman.org/blob/gh-pages/_layouts/plugin.html#L11-Lundefined

_data/misc.yaml
lzap: Yeah, maybe we could implement plugin matrix for each individual item and then render those for each plugin. Discovery is a good example of maintaining this in a table if you want some input data. Then we could use the data also to provide correct notifications to users about incoming plugin updates.

xprazak2:
I added the missing versions of core into the list. I think the current format will not be too difficult to maintain, even with patch versions. I do not see a reason why it should not contain pre-releases, since we currently advertise 1.11.0-RC1 as latest version on the front page.

I also plugin into the list together with required foreman version. It should allow us to determine the latest plugin version for the foreman version that user has.

_data/releases.yml
```diff
@@ -0,0 +1,674 @@
...
+    - 0.0.2rc1
+    - 0.0.1
+    - 0.0.1-6
+    - 0.0.1-5
+    - 0.0.1-4
+    - 0.0.1-3
+    - 0.0.1-2
+    - 0.0.1-1
+plugins:
+  foreman-digitalocean:
```
domcleal: Why hyphenated?  The engine and gem name are both lower case.

_data/releases.yml
```diff
@@ -0,0 +1,674 @@
...
+  foreman-docker:
+    1.4.0:
+      requires_foreman: '1.4'
+    0.0.3:
+      requires_foreman: '1.4'
+    0.0.2:
+      requires_foreman: '1.4'
+    0.0.1:
+      requires_foreman: '1.4'
+    2.0.0:
```
domcleal: Is this list meant to be sorted?  It'll be easier to maintain.

_data/releases.yml
```diff
@@ -0,0 +1,674 @@
...
+      requires_foreman: '1.4'
+    1.0.0:
+      requires_foreman: '1.4'
+    0.2.0:
+      requires_foreman: '1.4'
+  foreman-one:
+    0.0.4:
+      requires_foreman: '1.8'
+  foreman-tasks:
+    rubygem-foreman-tasks-0.6.1-1:
```
domcleal: Looks incorrect...

_data/releases.yml
```diff
@@ -0,0 +1,674 @@
...
+    1.0.1:
+      requires_foreman: '1.4'
+    1.0.0:
+      requires_foreman: '1.4'
+    0.2.0:
+      requires_foreman: '1.4'
+  foreman-one:
+    0.0.4:
+      requires_foreman: '1.8'
+  foreman-tasks:
```
domcleal, line 121: also, gem or engine name?

xprazak2:
Ah, inconsistencies in repo names, engine names and gems and in some cases even releases. Great. Will be back with changes

domcleal:
Yeah, it's not great, but it happens despite the advice.  I would tend towards using the name that is registered in Foreman's plugin database at runtime.

xprazak2:
foreman-taks is the only plugin that has hyphen in name, which can be seen at runtime, so I kept it here as well. Everything else should have underscore now and be sorted in descending order.

_data/releases.yml
```diff
@@ -0,0 +1,822 @@
...
+  - 0.0.2rc1
+  - 0.0.1
+  - 0.0.1-6
+  - 0.0.1-5
+  - 0.0.1-4
+  - 0.0.1-3
+  - 0.0.1-2
+  - 0.0.1-1
+:plugins:
+  foreman_digitalocean:
```
domcleal: File not alphabetised?

_data/releases.yml
```diff
@@ -0,0 +1,822 @@
+---
+1:
+  11:
+  - 1.11.0-RC1
```
domcleal: missing RC2

domcleal:
The branch doesn't merge cleanly, please rebase it.

> foreman-taks is the only plugin that has hyphen in name, which can be seen at runtime

So going back to my earlier questions, which name is being used for this data?  The gem name?  (Since foreman-tasks' engine name is `foreman_tasks`, with an underscore.)

xprazak2:
The gem name is used. It is also the one that is registered in Foreman runtime database afaik.

xprazak2:
This feature will be replaced by [rss feed dashboard widget](https://github.com/theforeman/foreman/pull/3400)

PR #559, opened 2016-03-06
remove extra html for G+ and Twitter

mmoll:
Pages are reachable with and without .html, but adding this unconditionally can lead to ".html.html" links


ohadlevy:
@mmoll thanks!

PR #561, opened 2016-03-07
Update references to Rails versions and source installations

domcleal:


mmoll:
merged, thanks @domcleal!

PR #566, opened 2016-03-08
Update smart proxy DHCP module configuration for 1.11

domcleal:
DHCP module now uses separate providers rather than the "vendor"
setting, also rewritten to match the DNS section in style.


mmoll:
merged, thanks @domcleal!

PR #569, opened 2016-03-09
Add headline features for 1.11

domcleal:


mmoll:
:+1: maybe the new Patternfly UI is also worth mentioning here?

domcleal:
Perhaps, I was thinking about it but I don't think there's much to say about it.

mmoll:
That's also correct... merged, thanks @domcleal!

GregSutcliffe:
I'll get clips done for those next week

PR #570, opened 2016-03-10
Add generated 1.11 API docs

domcleal:
Previous releases have only had English API docs, should I rebuild with that just or keep every supported language?  It does add an awful lot of files, but since we have the content, it seems a shame not to publish it.


mmoll:
@tstrachota :question: :point_up:

tstrachota:
Since it's available. +1 for publishing it.

mmoll:
merged, thanks @domcleal!

PR #580, opened 2016-03-30
Add preview clips for release highlights

GregSutcliffe:
Videos are currently unlisted but links should work - will make them public once we're done. I think the audio may still need tweaking on a couple, but I have the raw files so it can be re-rendered easily enough. Comments welcome.

### DIFF
--- a/_includes/manuals/1.11/1.2_release_notes.md
+++ b/_includes/manuals/1.11/1.2_release_notes.md
@@ -4,6 +4,8 @@
 
 #### Improvements to host and host group parameter editing
 
+* [YouTube video: Inline parameter editing](https://youtu.be/PXaVBB88b2Y)
+
 The parameters tabs of the host and host group edit pages have been significantly improved with a few changes:
 
 * Overridden smart class parameters and variables are now edited inline, replacing the overridden value. This removes the need to scroll when overriding parameters to the overridden list.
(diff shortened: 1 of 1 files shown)


domcleal:
The audio on the middle two videos (smart proxies, bulk actions) is somewhat lower than the other two (test buttons is higher than inline params, but both are OK), could you increase those?

Otherwise fine, thanks.

ohadlevy:
I agree on the audio as well, i had to bump volume across the videos multiple time.
I also found it odd that your select/dropdowns have  a black background, making it odd looking.

and in general, I would recommend to have more content in your foreman instance, as one host demo is not very appealing.

thanks!

GregSutcliffe:
Thanks guys, I have no idea why my mic is so variable (these were all recorded on the same day...). On the plus side, I've now learned how to use ffmpeg to normalise to a set dB level, so once I re-upload they should be consistent from now on.

Thanks for the catch on the weird black menus, I hadn't really noticed. GTK is doing some weird interaction with Firefox lately - if I get chance I'll re-record that segment.

GregSutcliffe:
Ok, all videos re-uploaded and URLs changed here. I've normalised to 0dB so it should be _much_ louder now. I did also manage to fix the GTK issue, so the smart-proxy vid is a new recording without the black menus.

domcleal:
Merged, thanks @GregSutcliffe.

PR #583, opened 2016-04-04
Adds documentation on the ARF reports migration process from 0.4.x to…

shlomizadok:
… 0.5.x

### DIFF
--- a/plugins/foreman_openscap/0.5/index.md
+++ b/plugins/foreman_openscap/0.5/index.md
@@ -110,6 +110,36 @@ Edit ```/etc/foreman-proxy/settings.d/openscap.yml``` with the appropriate setti
 
 This puppet module will automatically install foreman_scap_client (if not installed) and will configure the client's ```/etc/foreman_scap_client/config.yaml``` with parameters which are needed for the operation of foreman_scap_client.
 
+# 3. Upgrading from 0.4.x
+
+As with any Foreman plugin, the recommended upgrade path is done via `yum upgrade`. This will ensure that all of the packages are with the right version, dependencies
+and data migrations.
+
+__Note: ARF reports are not automatically migrated and need a manual step__
+
+## 3.1 Migrating ARF reports from 0.4.x
+
+ARF reports in 0.5.x and Foreman > 1.11 are now part of Foreman's reports (Reports STI), and the physical ARF report XML is now saved at the Smart-Proxy.
+
+This requires a special migration, which needs both Foreman and Smart-Proxy up and running, with latest respective OpenSCAP plugins installed.
+ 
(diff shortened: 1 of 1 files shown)


plugins/foreman_openscap/0.5/index.md
```diff
@@ -110,6 +110,33 @@ Edit ```/etc/foreman-proxy/settings.d/openscap.yml``` with the appropriate setti
...
+This requires a special migration, which needs both Foreman and Smart-Proxy up and running, with latest respective OpenSCAP plugins installed.
+ 
+Since we cannot assure this during upgrade, migration of ARF reports has moved to a rake task which should be performed after upgrading.
+
+To upgrade ARF reports from 0.4.x to 0.5.x:
+
+* Upgrade foreman_openscap to version 0.5.x
+* Upgrade smart_proxy_openscap to version 0.5.x
+* Restart Foreman and Smart-Proxy
+* Run `foreman-rake foreman_openscap:migrate[<proxy_id>]` (Where `<proxy_id>` is the id of the proxy with OpenSCAP feature, please note it should be inside the square brackets)
```
ares: Could you please add a description howto determine the id? I guess reading from `hammer proxy list` or URL in proxy detail page is the easiest.

plugins/foreman_openscap/0.5/index.md
```diff
@@ -110,6 +110,33 @@ Edit ```/etc/foreman-proxy/settings.d/openscap.yml``` with the appropriate setti
...
+ARF reports in 0.5.x and Foreman > 1.11 are now part of Foreman's reports (Reports STI), and the physical ARF report XML is now saved at the Smart-Proxy.
+
+This requires a special migration, which needs both Foreman and Smart-Proxy up and running, with latest respective OpenSCAP plugins installed.
+ 
+Since we cannot assure this during upgrade, migration of ARF reports has moved to a rake task which should be performed after upgrading.
+
+To upgrade ARF reports from 0.4.x to 0.5.x:
+
+* Upgrade foreman_openscap to version 0.5.x
+* Upgrade smart_proxy_openscap to version 0.5.x
```
ares, line 131: I think `foreman-rake db:migrate` should be run here after update right? Normally installer does it I think but if user would do `yum update` only they would have old schema.

plugins/foreman_openscap/0.5/index.md
shlomizadok: Added. Thanks

plugins/foreman_openscap/0.5/index.md
shlomizadok, line 131: I have added (though I thought yum upgrade does migration too).
Anyhow it is a safe step to add.

ares:
Thanks @shlomizadok, merging!

PR #585, opened 2016-04-05
link to Netways webinar on different ways to provision systems added

dgoetz:
Hi,

I added a link to our webinar on  different ways to provision systems with Foreman (in german) which is the start of a new series we will run this year.

Regards,
 Dirk


dLobatog:
Merged as 1823d441e15ad2c3126130fa688369278149cfc5, thanks @dgoetz!

PR #586, opened 2016-04-05
Ubuntu Trusty users should set _PassengerRuby_ not _PassengerRoot_

sodabrew:
https://www.phusionpassenger.com/library/config/apache/reference/#passengerruby


domcleal:
Thanks!

PR #588, opened 2016-04-08
Add events system to website

GregSutcliffe:
Here's a first pass at an events system for the site. I've stayed off the front page for now, so this is 99% new additions, with just a couple of links on the Docs and Community pages.

The JS/CSS is mostly taken from https://github.com/OSAS/community-website/ (community.redhat.com/events) and tweaked for our usecases.

The new assets (FullCalendar & Moment) do have gems, but not on rubygems.org (only on rails-assets.org) which would require multiple source entries in the Gemfile. Bundler didn't seem to like that too much, so I opted to grab the assets from the gem and drop them into /static instead.

Testing out is easy enough, just bundle update, render the site as normal and hit /events.


_plugins/events.rb
```diff
@@ -0,0 +1,124 @@
...
+        cal.add_event(cal_event)
+      end
+      cal.to_ical
+    end
+
+  end
+
+end
+
+Liquid::Template.register_filter(Jekyll::Events)
```
domcleal, line 124: Plugins aren't compatible with GitHub pages, which is something we've tried to retain in this repo (e.g. in #550).

GregSutcliffe, line 124: I guess that raises two questions:
(1) We don't deploy on GitHub pages, so why do we care about that? The rendering clearly works
(2) If we do care, how else can this extra code be loaded? It's pretty substantial, I wouln't want to try and do it in Liquid tags.

domcleal, line 124: > (1) We don't deploy on GitHub pages, so why do we care about that?

The site used to be hosted on GitHub Pages, and it still can at the moment.  It may provide a way to host the website if needed or allow us to change back one day if we wanted, to save self-hosting.

This repo still has the Pages service enabled, and it can be enabled on forks too.  So far, compatibility with Pages has been retained by updating to the latest github-pages gem and security rules etc.

> (2) If we do care, how else can this extra code be loaded?

It's not possible to load extra Ruby code on GitHub Pages AFAIK.

GregSutcliffe, line 124: In that case, I propose we do use a plugin. Reasoning:

(1) Given our other hosting requirements and costs (packages, tarballs, jenkins servers, etc) the website is the least of our worries. If funding dried up, we'll see it coming.
(2) From what I read of the GitHub Pages docs, you can still render the site locally and push the static result to GitHub Pages. Thus should we ever need to move back to GitHub Pages, there is an interim method for doing so with this code, which could then be removed or refactored as required
(3) I don't see any other way to format this data using the basic Liquid filters

_data/events.yml
```diff
@@ -0,0 +1,55 @@
...
+  speaker: 'Julien Pivotto'
+  start: 2016-04-28 09:30 CEST
+  end: 2016-04-28 10:30 CEST
+  url: 'https://www.netways.de/events_schulungen/osdc/program/julien_pivotto_automating_a_rd_lab_with_foreman_what_can_be_hard/'
+  description: |
+    This will tell the story of how we automated our R&D environment with the
+    Foreman and what were the biggest problems we are facing. It will contain a
+    very short introduction about the Foreman but the main subject will really
+    be about experience.
+- title: 'Case Study: Discovery and Custom Plugins with Imzi Zvik'
```
dLobatog: Typo, Imri ?

_includes/calendar.js
```diff
@@ -0,0 +1,93 @@
...
+
+    // Position, inject tooltip text, and fade tooltip in
+    $tip.offset({
+      top: $item.offset()["top"] - 20,
+      left: $item.offset()["left"] - $tip.width() - 40
+    }).html(eventText).stop().fadeTo(200, 1);
+  };
+
+  // Hide the tooltip
+  tooltipHide = function(event, jsEvent, view) {
```
dLobatog: Why not making all these functions outside of the `.ready` scope? I don't see any reason to define them inline and save them in variables

_plugins/events.rb
```diff
@@ -0,0 +1,124 @@
...
+
+    def format_event_date start_date, end_date
+      # Compares start_date with end_date to decide how to format the date correctly
+      return "" unless start_date
+
+      if start_date.class == Date
+        # no times, just use dates
+        if end_date.nil? || start_date == end_date
+          # one-day event
+          return start_date.strftime('%-d %B, %Y')
```
dLobatog: Useless return? There are retuns all over the place in here which I don't think are needed - also you may split this in multiple functions or a date/time parser class

events/index.html
```diff
@@ -0,0 +1,58 @@
...
+<div class='col-md-12'>
+  <div class='col-md-6 pull-left'>
+    <p>Here you'll find all the Foreman-related events where members of the
+    community are talking about their use of, or contribution to, Foreman.
+    Attending one of these events is a great way to learn more about Foreman!<p>
+    <p>You can <a href='/events/all.ics'>subscribe to these events</a> via our ICS feed!</p>
+    <p>Know of an event not listed here? .. or want to fix some details? Please
+    <a href='https://github.com/theforeman/theforeman.org/blob/gh-pages/_data/events.yaml'>contribute event info</a>!</p>
+  </div>
+  <div class='col-md-6 col-offset-6 pull-right'>
```
dLobatog: No need for the offset when there's already a col-md-6 div before. In fact if you use the right class `col-md-offset-6` (the current one doesn't exist) it will misalign things

events/index.html
```diff
@@ -0,0 +1,58 @@
...
+    <div id='calendar'></div>
+  </div>
+</div>
+
+{% for event in events %}
+<div class="blog-item post-content" id="{{ event.title | slugify }}">
```
dLobatog, line 7: This needs some padding on top, it appears right under the calendar (no space)

static/css/fullcalendar.css
```diff
@@ -0,0 +1,1124 @@
+/*!
```
dLobatog, line 1: Whoa can we load this from somewhere else..?

dLobatog:
screenshot from 2016-04-11 15-00-42

The page looks like it's missing some CSS fixes. In particular notice:
- Lack of space between title and first 2 blocks of content
- Weird indentation on the first div ("here you'll find.."), notice how other pages don't have that indent.
- Horizontal line in between title and divs - other pages don't have that
- Lack of space between calendar and event list - this is easily fixed by adding some padding.

Other than that :+1: . I don't really like bundling up heavy assets though, so if you could look at reliable external cdn-like sources or a way to put them here that'd be great :smile:

_plugins/events.rb
dLobatog, line 124: Given we can push the static to GH Pages as well if we ever needed to do so (seems unlikely to me but who knows), I don't see how breaking the compatibility would be a very big issue.

GregSutcliffe:
@dLobatog thanks for the review, I'll work on that next, while the plugin debate continues :)

@domcleal I still don't really understand _why_ we care about GH-Pages support given the apparently low chances of us using it again. As a test, I pushed my fork to GH-Pages (http://theforeman.emeraldreverie.org/events/ if you wish to look) and it renders fine, just without the calendar/ics file. That tells me we could still deploy to GH-Pages without issue if we really had to. I'm not sure if your concern is having a 100% broken deployment or a 100% working deployment, so perhaps this knowledge is sufficient?

If not, it did occur to me that I could probably take the work the Ruby plugin is doing and re-implement it in pure JS, which should load fine on GH-pages. The sticky bit might be the ICS file since that will still have to be static, but ICS is a well-defined format so it _might_ be possible to do it with the standard GH-Pages toolset. Or a middle road of the /events page being pure JS and the only Ruby usage is for the ICS file - that way the ICS file is the only thing that breaks if we ever go back to GH-Pages.

I'm pretty sucky at JS though, so it might be some effort for me - I don't really want to start on a pure JS version unless there's a decent chance we can all agree on it. Thoughts?

_data/events.yml
GregSutcliffe: Thanks!

_includes/calendar.js
GregSutcliffe: Moved to outside the `.ready`.

_plugins/events.rb
GregSutcliffe: I've cleaned up the returns, it's a bad habit. Happy to clean it up further if we're keeping the Ruby implementation.

events/index.html
GregSutcliffe: Fixed, thanks.

events/index.html
GregSutcliffe, line 7: Added 20px to the base of the calendar - adding to the top of the text list just didn't seem to be working

static/css/fullcalendar.css
GregSutcliffe, line 1: So I dug into this. We can indeed use a second gem source in the Gemfile, with the correct syntax. However, I can't find a way to load from gem assets rather than `/static`. There is a [jekyll-assets](https://github.com/jekyll/jekyll-assets) plugin but that brings us back to the discussion about plugins - even then, I'm not 100% sure it can load from gems.

I guess there's a reason jquery is already in /static...

GregSutcliffe:
@dLobatog comments (mostly) addressed. I've also switched to an h2 for the title to match the month on the calendar, lines up nicely now.

dLobatog:
@GregSutcliffe :+1: from me as it stands now. 
@domcleal any reason to think theforeman.org might have to go back to GH pages? 

In any case, we can just take the Ruby events plugin, compile it to JS using [Opal](https://github.com/opal/opal) and use that.

GregSutcliffe:
@dLobatog thanks. For the record, Opal would work for the calendar on `/events` but not for the ical stuff on `/events/all.ics` - that's because it's going to be calendar applications rather than browsers that are using that URL, and I expect they won't run Javascript.

_plugins/events.rb
domcleal, line 124: That's OK then, I was concerned about it raising alerts or similar with plugin code in this repo, as GH Pages is enabled (due to the branch name).  It doesn't appear to raise any warnings, just ignores it.

GregSutcliffe:
@domcleal thanks for the clarification. @dLobatog added a few more things to events.yml, and squashed everything up.

dLobatog:
Merged as ba6719fbf22e04b342ee384d2bda94f3e61f4d70, thanks @GregSutcliffe!

PR #593, opened 2016-04-14
Fix ICS file dates and tootlips

GregSutcliffe:
Two commits here. The first adds the off-by-one date fix to the ICS file (eg if you import the current ICS to you calendar, you'll see that OSDC is only showing as 2 days, not 3).

The second moves the javascript functions for the tooltips back inside document.ready. They currently aren't working from outside the ready, and I can't work out why. In addition http://stackoverflow.com/questions/2645344/functions-inside-or-outside-jquery-document-ready seems to imply that making things global when you don't need to isn't a good idea.


dLobatog:
Merged, thanks @GregSutcliffe !

PR #594, opened 2016-04-14
NoVNC tip about FQDN

lzap:
When you force-ignore a certificate in Firefox, console does not work in SSL
mode, until you use FQDN. Looks like browsers remember this per port.


domcleal:
Thanks @lzap!

PR #600, opened 2016-04-27
Clarify hash usage with foreman-installer

aairey:
The docs tell you to use a hash in the form of `{show_diff => true, stringify_facts => false}`, which is puppet syntax.

However with foreman-installer you need to use the kafo syntax for passing hash arguments, thus passing them individually.

```
foreman-installer \
--puppet-server-additional-settings=show_diff:true \
--puppet-server-additional-settings=stringify_facts:false
```

The docs for the foreman options are auto-generated and contain the puppet syntax only.

Added a note to the docs so this should be less confusing for users.
@domcleal asked me to update the docs on this via IRC.


domcleal:
Thanks @aairey!

PR #604, opened 2016-05-01
fixes #13728 - fix TOC of foreman_docker page

mmoll:
Also, don't use "Introduction", because this renders to

``` html
<h1 id="introduction">2. Introduction</h1>
```

and the introduction style from style.css matches there.


domcleal:
Thanks @mmoll!

PR #606, opened 2016-05-04
fixes typo - budle exec -> bundle exec

fwdIT:
came across a typo when setting up a new dev environment

-1. Populate database: `budle exec rake db:seed` and take note of the password it generates
+1. Populate database: `bundle exec rake db:seed` and take note of the password it generates


GregSutcliffe:
Merged as 1854cf9, thanks!

PR #609, opened 2016-05-25
document proxy ssl_disabled_ciphers option

tbrisker:


domcleal:
Thanks, looks good. Could you also add a shorter description to the general settings bit in section 4.3.2?  It doesn't need to be detailed, it can refer/link to the later section for detail (e.g. the list of current and known ciphers should be left out).

tbrisker:
@domcleal done

_includes/manuals/1.12/4.3.10_smartproxy_ssl.md
```diff
@@ -29,6 +29,28 @@ Configure the locations to the SSL files in `/etc/foreman-proxy/settings.yml`, p
 #- foreman.dev.domain
 </pre>
 
+##### SSL cipher suites
+
+By default, the smart proxy permits the following SSL cipher suites:
+* ECDHE-RSA-AES128-GCM-SHA256
```
domcleal, line 36: I think you need a line break above here, it's not rendering as a list.

tbrisker, line 36: fixed

domcleal:
Thanks @tbrisker!

Issue #240, opened 2014-08-03
Signing key for Debian packages not available as a secure download


danielkza:
The manual (as of version 1.5.2) instructs you to start the installation on a Debian system by runnning the following commands:

``` sh
echo "deb http://deb.theforeman.org/ wheezy 1.5" > /etc/apt/sources.list.d/foreman.list
echo "deb http://deb.theforeman.org/ plugins 1.5" >> /etc/apt/sources.list.d/foreman.list
wget -q http://deb.theforeman.org/pubkey.gpg -O- | apt-key add -
apt-get update && apt-get install foreman-installer
```

Notice how the signing key is downloaded through plain HTTP. Simply replacing http with https in the URL doesn't work: strangely I don't even get a server error, just no response at all. The key has to be available from a secure source to even be useful at all, otherwise you can be MITM-ed and never notice any errors or failures.

Issue #539, opened 2016-02-10
Move websites to HTTPS


domcleal:
This is a task list for things that need moving to HTTPS across the website and related bits.  In a week or so's time, we [should have HTTPS available](https://github.com/theforeman/foreman-infra/pull/211) for all sites hosted on the main web server (this website repo, apt and yum repos) but we need to update links and instructions to point to it.  Eventually we can redirect HTTP too.
- [x] Enable HTTPS on web server
- [x] Ensure no mixed content on theforeman.org, e.g. iframes, CSS/JS from external sources (introduction.md uses an iframe to http://slideshare.net for example): #541
- [x] Change apt GPG key download to HTTPS, aka #240: #616
- [ ] ~~Change apt repo URLs to HTTPS~~
- [x] Change yum foreman-release.rpm download to HTTPS: #616
- [ ] Change foreman-release .repo files to HTTPS
- [ ] Change yum repo references in the manual to HTTPS
- [x] Change puppet-foreman repo URLs to HTTPS
- [x] Redirect http://theforeman.org to HTTPS
- [ ] ~~Redirect http://deb.theforeman.org to HTTPS~~
- [ ] Redirect http://yum.theforeman.org to HTTPS

ekohl:
Debian doesn't install apt-transport-https by default so that may be something to be aware of. https://qa.debian.org/popcon.php?package=apt says 6.84% has it installed.

domcleal:
Good to know.  I don't think there's a huge benefit to retrieving the packages themselves over HTTPS, so we don't have to do that.  The main thing is to get the initial GPG key over secure transport, which is downloaded directly from deb.theforeman.org for apt, and from a direct foreman-release.rpm download for yum.

ekohl:
I agree so I'd make them both available allowing users to choose, but wouldn't implement the Debian redirect.

lzap:
For the record, I've moved my blog (hosted via github pages) to
CloudFlare CDN and I am very happy. They provide SSL for free and there
are various setups possible including SSL termination and HTTP or HTTPS
with invalid certificate as a followup.

https://lukas.zapletalovi.com/2015/12/served-by-cloudflare.html

It provides also great statistics and other optimizations.

## 

Later,
 Lukas #lzap Zapletal

domcleal:
> Enable HTTPS on web server

Now enabled on all vhosts, we can work through the other tasks.

domcleal:
[...]

PR #616, opened 2016-06-01
Change all foreman-release and pubkey.gpg downloads to HTTPS

domcleal:
Fixes GH-240

### DIFF
--- a/_includes/manuals/1.11/2.1_quickstart_installation.md
+++ b/_includes/manuals/1.11/2.1_quickstart_installation.md
@@ -71,7 +71,7 @@ yum-config-manager --enable rhel-7-server-optional-rpms rhel-server-rhscl-7-rpms
   </p>
 
 {% highlight bash %}
-rpm -ivh http://yum.puppetlabs.com/puppetlabs-release-el-6.noarch.rpm
+rpm -ivh https://yum.puppetlabs.com/puppetlabs-release-el-6.noarch.rpm
 {% endhighlight %}
 </div>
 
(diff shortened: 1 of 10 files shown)


_includes/manuals/1.11/2.1_quickstart_installation.md
```diff
@@ -147,7 +147,8 @@ yum -y install http://yum.theforeman.org/releases/{{page.version}}/f21/x86_64/fo
 {% highlight bash %}
 echo "deb http://deb.theforeman.org/ jessie {{page.version}}" > /etc/apt/sources.list.d/foreman.list
 echo "deb http://deb.theforeman.org/ plugins {{page.version}}" >> /etc/apt/sources.list.d/foreman.list
-wget -q http://deb.theforeman.org/pubkey.gpg -O- | apt-key add -
+apt-get -y install ca-certificates
+wget -q https://deb.theforeman.org/pubkey.gpg -O- | apt-key add -
```
GregSutcliffe, line 151: This won't work out of the box for Debian - https support isn;t enabled by default in apt (sigh, i know):

```
greg@topaz:~$ cat /etc/apt/sources.list.d/foreman.list 
deb https://deb.theforeman.org/ jessie 1.11
deb https://deb.theforeman.org/ plugins 1.11
greg@topaz:~$ sudo apt-get update
E: The method driver /usr/lib/apt/methods/https could not be found.
N: Is the package apt-transport-https installed?
E: The method driver /usr/lib/apt/methods/https could not be found.
N: Is the package apt-transport-https installed?
```

It's probably enough to add apt-transport-https to the preceeding line (and ditto all the ones below)

domcleal, line 151: The instructions on lines 148/149 remain HTTP, and it's only the wget (on this very line) that is HTTPS. It doesn't say to put an HTTPS URL into the sources file.

GregSutcliffe:
Other than the Debian comment, any blockers to this now that https is enabled? It seems to work for me once I enabled apt-transport-https:

```
greg@topaz:~$ apt-cache policy foreman
 *** 1.11.2-1 0
        500 https://deb.theforeman.org/ jessie/1.11 amd64 Packages
```

_includes/manuals/1.11/2.1_quickstart_installation.md
GregSutcliffe, line 151: Ah good point, I missed that. Let me rephrase then - do we want to recommend an https repo? I'm in two minds....

domcleal, line 151: I don't think so, I'm OK with leaving it HTTP and just ensuring that all public key downloads are HTTPS (also see #539). This also gives users the benefit of caching from any inline proxy servers.

GregSutcliffe, line 151: True enough, let's leave it be.

domcleal:
> Other than the Debian comment, any blockers to this now that https is enabled?

Not that I'm aware of, it works OK in my tests.

GregSutcliffe:
Merged as 85a4f36, ty @domcleal!

PR #617, opened 2016-06-03
Document Puppet 4 compatibility and installation

domcleal:
Recommends the use of either PC1 or 3.x repos depending on the OS
availability, documents compatibility with Puppet 4 and Puppet Server
versions, and adds notes about differing SSL certificate paths
throughout the manual.


_includes/manuals/1.12/2.1_quickstart_installation.md
```diff
@@ -133,9 +153,18 @@ yum -y install https://yum.theforeman.org/releases/{{page.version}}/f24/x86_64/f
...
+    Using Puppet 4.x is recommended, which is available from the Puppet Labs repository.
+    You may skip this and use Puppet 3.x from Debian without a problem, however it has reduced community support.
+
+    To use Puppet 4.x with Puppet Agent and Puppet Server:
   </p>
 
+{% highlight bash %}
+apt-get -y install ca-certificates
+wget https://apt.puppetlabs.com/puppetlabs-release-pc1-trusty.deb
+dpkg -i puppetlabs-release-pc1-trusty.deb
```
mmoll: th two lines above should be jessie as distribution

mmoll:
I have to admit, I didn't check, but I think `puppet kick` has been removed ith Puppet 4, so that should be mentioned in the puppetrun docs.

_includes/manuals/1.12/2.1_quickstart_installation.md
domcleal: Thanks, fixed.

domcleal:
> I have to admit, I didn't check, but I think puppet kick has been removed ith Puppet 4, so that should be mentioned in the puppetrun docs.

Yes, it looks like it has been. I've added this information to the puppetrun docs.

mmoll:
merged, thanks @domcleal!

dLobatog:
Literally going through this now, thanks @domcleal & @moll, will report if anything goes wrong on my setup :)

PR #620, opened 2016-06-03
fix TOC of the manuals

mmoll:


domcleal:
Is this just fixing the visual markdown, or is there a bug it fixes that I'm not seeing?

mmoll:
In the toc at the left, point 5 is not visible, leading to a layout where all the points 4.\* and 5.\* are on the same level.

domcleal:
Thanks @mmoll, I see it now after updating gems. Merged!

PR #623, opened 2016-06-06
Add supported OSes preparation step, remove old OS instructions

domcleal:


mmoll:
merged, thanks @domcleal!

PR #625, opened 2016-06-07
Document 1.12 upgrade step to install openstack/rackspace

domcleal:


mmoll:
merged, thanks @domcleal!

PR #636, opened 2016-06-22
pet html validator

mmoll:


_includes/header.html
```diff
@@ -51,9 +51,9 @@
   {% endif %}
 		<div class="container">
       {% if page.url != "/" %}
-		    <nav class="navbar navbar-default navbar-static-top" role="navigation">
```
ekohl, line 54: I think this role is used by bootstrap.

mmoll, line 54: It's recommended in the bootstrap docs for accessibility reasons, but it seems not to have any effects CSS wise. As this role addition is marked as warning by the W3 Validator, I guess newer browsers handle <nav> tags as navigation elements anyway.

domcleal:
Thanks @mmoll.

PR #638, opened 2016-06-23
Fixes #14999 - correct link to hammer plugin in salt docs

orrabin:
The docs for later versions of the plugin moved to relative urls.


mmoll:
thanks @orrabin!

PR #646, opened 2016-06-28
Remove ovirt_provision upgrade note, fixed in new version

domcleal:


mmoll:
merged, thanks @domcleal!

PR #647, opened 2016-06-28
Add hardware requirements to installation guide

domcleal:
Recommends 4GB memory as standard, to allow 2GB for Puppet Server and
the same for Foreman's default Passenger configuration.


dLobatog:
Merged as 9faddb11820bbb295d5f2e72125beacddda1174f, thanks @domcleal!

PR #650, opened 2016-06-29
events: separate dates from external URLs

GregSutcliffe:
It's easy to miss the fact that there's a URL link in the date section of an event entry - this moves the date to be more visible and puts the URL below.

@mmoll @dLobatog thoughts?


dLobatog:
Makes sense to me, I would've definitely missed the URLs

dLobatog:
Merged as b0039ca505baf81a094e77572d1afd30e070231a, thanks @GregSutcliffe!

PR #651, opened 2016-06-29
Add a News Teaser banner to the top of the site

GregSutcliffe:
This is to (on occasion) draw attention to something important happening (events, new releases, etc). It adds a hopefully-not-too-obnoxious banner to the top of the site, if `_config.yml` is configured to do so.

As a starter, I've pointed it to the birthday page, since I want to promote that pretty heavily.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -14,6 +14,8 @@ excerpt_separator: <!--more-->
 permalink: /:year/:month/:title.html
 blogname: Foreman Blog
 blogdesc: News, notes and noise from around the Foreman community
+teaseurl: /2016/06/foremans-7th-birthday-events.html
+teasetext: Foreman's 7th Birthday is coming! See if there's a party near you!
 
 safe:        false
 
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -44,7 +44,10 @@
 </head>
 
 <body>
- {% if page.url != "/" %}
+  {% if site.teaseurl != nil %}
+    <a href='{{ site.teaseurl }}' class="news-tease">{{ site.teasetext }}</a>
+  {% endif %}
+  {% if page.url != "/" %}
     <header id="page-header" style='padding: 0'>
 	{% else %}
     <header id="page-header">
--- a/static/css/style.css
+++ b/static/css/style.css
@@ -1186,3 +1186,21 @@ a.fc-event.old {
   border-color: #9d8d9e;
 	background-color: rgba(2, 93, 140, 0.2);
 }
+
+/* Banner alert */
+
+a.news-tease {
+  background-color: rgb(0, 62, 95);
+  box-sizing: border-box;
+  color: rgb(255, 255, 255);
+  display: block;
+  line-height: 20px;
+  padding-bottom: 5px;
+  padding-left: 20px;
+  padding-right: 20px;
+  padding-top: 5px;
+  text-align: center;
+  text-decoration-line: none;
+  text-decoration-style: solid;
+  text-decoration-color: rgb(255, 255, 255);
+}


domcleal:
There's a news section on the homepage and the blog, please use that if necessary.  I don't think a third news section is required, especially on every single page.

GregSutcliffe:
I'm aware of the news section - I'd have used it if I thought it was appropriate. In my opinion, it is almost completely useless at the bottom of a page 3 screens long - I don't recall the last time I saw or read it. It certainly won't be seen by people who (a) are going direct to links in their history, or (b) are using the navbar to get to where they want to go.

There will be a blog post on the birthday events, of course - I'm waiting for the last two events to confirm dates and venues. That doesn't help raise awareness with the above two usecases either.

To re-iterate, this is intended as an infrequent-use tool to raise awareness of something deemed important to the community. I'm not proposing to leave it enabled permanently. I'm hearing from around the community that we have a problem with people being unable to discover upcoming events and news - this seems like a good way to improve that.

GregSutcliffe:
Addendum, since you raise the issue of a third news section: this isn't a new section, it's just a link to something we've already created. In hindsight, the `/events/birthday.html` page probably _should_ have been a blog entry, but I'd still want this PR to link to it.

ehelms:
:+1: Some news is good for the news section, but big announcements or events that we want to intentionally draw users attentions to make sense to have this sort of banner. Analogous to news sites that put up breaking news banners this will draw users attentions to it no matter what part of the site they frequent.

domcleal:
This isn't a big announcement that it's imperative that people who aren't trying to view the news or blogs section must see, it shouldn't be shown on every single page.  Consider improving the news section if it's poorly placed.

dLobatog:
:+1: for the same reasons @ehelms cited. I'm not sure the birthday is that important, but it'd be interesting to have it for new releases, CVE fixes... in any case my point is that the current news section is not very discoverable.

dLobatog:
:+1: for the same reasons @ehelms cited. I'm not sure the birthday is that important, but it'd be interesting to have it for new releases, CVE fixes... in any case my point is that the current news section is not very discoverable.

GregSutcliffe:
@domcleal This feels like mostly a difference of opinion here - I'm going to try and break down the various sections of opinion so we can try to see where we agree / disagree.

> This isn't a big announcement that it's imperative that people who aren't trying to view the news or blogs section must see

Perhaps we have a different view on the website's purpose - I'm getting the sense that you see it solely for distributing technical information, while I see it as the central hub for our community - technical _and_ social. That's going to lead to friction, I guess, and I'd be happy to discuss how we can resolve that.

For me, this is largest collection of simultaneous events we've ever tried to organise, aimed at bringing together our community to celebrate the huge amount of work that's gone into the last year of releases - it is indeed a big thing. I think that we should give every possible chance for any member of the community to be aware of this.

> it shouldn't be shown on every single page.

According to our analytics, the average website session in the last 6 months is only 2.32 pages. That doesn't give us much room to manoeuvrer if a user doesn't hit one specific page. The banner is designed to be as unobtrusive as possible, since it's the same colour scheme (plus a black alpha channel) and hides on scroll.

It's worth noting that I do consider this to be an experiment - currently I hear complaints from the community about how hard it is to know what's happening, so I'm trying to solve them. If I hear just as many complaints after we try this, I'll be happy to remove it. I'll also be happy to remove it after any revamp of the news section - see below.

> Consider improving the news section if it's poorly placed.

Here we agree, and I definitely want to, but it's got to be done right. Firstly, I don't to end up in a PR argument again, if I can avoid it - it would be helpful to have a discussion about what changes you'd be happy with before I spend significant time on a PR that will potentially be abandoned. Should we take that redesign to the dev list? Secondly, I'm on a tight timeline to maximise awareness of these birthday events. It seems unlikely that such a redesign would be ready in time, although I'm happy to agree to give it high priority and revert this as soon as possible - assuming the redesign serves the same purpose better. 

@dLobatog I hadn't thought of CVEs, that's a good point. Maximising awareness of those is important too.

GregSutcliffe:
Updated to account for #654

domcleal:
> Here we agree, and I definitely want to, but it's got to be done right. Firstly, I don't to end up in a PR argument again, if I can avoid it - it would be helpful to have a discussion about what changes you'd be happy with before I spend significant time on a PR that will potentially be abandoned. Should we take that redesign to the dev list?

If you wish, or just open a pull request with the proposed changes.

GregSutcliffe:
@domcleal sent to the list for discussion - since it's most commonly you & I that differ on this stuff, your input would be good. For clarity, I'll state that I'd still like to move forward with this, as a short-term solution.

Regarding this PR - For those not able to render the site locally, here's a screenshot of the proposal:

2016-06-30-152535_1106x564_scrot

ehelms:
So far the only negative I hear is that it will be on every page which is also a positive because for those that go to certain pages always, or their browser auto-completes the pages they visit most often this would give them useful information. And yes, I consider community events, especially large ones that span multiple-geographies useful. This helps everyone and hurts no one and the longer it sits the less informed our users are and the less time they have to plan around the information.

domcleal:
I don't disagree that it's interesting information, but the current implementation is unnecessarily obnoxious in its placement above all other content and navigation on the page.

I might be persuaded to the idea of having a news item (i.e. the latest from the news block, or blog post) mentioned on other pages than the blog/homepage, but only if it was better placed in the page design/hierarchy than this is.  

The information's now available one click away from every page on the site through #654, through RSS and can also be added to the main news block.

domcleal:
> I might be persuaded to the idea of having a news item (i.e. the latest from the news block, or blog post) mentioned on other pages than the blog/homepage

I should also say that the news items are _already_ on every page, but at the bottom of the layout.  Perhaps using the bottom of the side menu bar, or placing the item carefully between the nav and content would be possible.

mmoll:
closing after inactivity.

PR #654, opened 2016-06-30
blog: Move birthday events to a blogpost

GregSutcliffe:
This came up in https://github.com/theforeman/theforeman.org/pull/651#issuecomment-229415855 - the `events/birthday` page isn't particularly discoverable if you're mostly using the blog/RSS feed.

I don't see any reason why it shouldn't be a blog post (it can still be updated as events come in), so I've moved it over and added a redirect.


dLobatog:
Merged as ddd1defa52108bc8edb3cf97c0744e45cb666f5c, thanks @GregSutcliffe!

PR #656, opened 2016-07-04
Add birthday to frontpage and news footer

GregSutcliffe:
This is a lower-impact version of #651 which adds a new include for the footer and top of the frontpage. It's structured as a separate include so we can blank it or change if there's other alerts we want to promote in the future.


dLobatog:
:+1: from me

ohadlevy:
thanks @GregSutcliffe - I would still like to see it in a more visible space, but since time is short, this is better then nothing. thanks!

PR #657, opened 2016-07-05
Update GCE instructions

dLobatog:


_includes/manuals/1.11/5.2.4_cr_gce.md
```diff
@@ -10,23 +10,24 @@
...
 #### Setting up the cloud project
 
-All Google Compute Engine access is contained within a "project" set up via the Google Developers Console.  Access [the Google Developers Console](https://console.developers.google.com/), and click *Create Project*.
+All Google Compute Engine access is contained within a "project" set up via the Google Developers Console.  Access [the Google Developers Console](https://console.developers.google.com/iam-admin/projects), and click *Create Project*.
 
 ![Create Project](/static/images/screenshots/gce/create_project.png)
 
-In the project settings, enable the Compute Engine feature by clicking one of the sections below *Compute Engine* in the left hand pane.  This will trigger an initialization job.
+By default, your project will have the Compute Engine and App Engine services enabled.
+Now go to [the API manager](https://console.developers.google.com/apis/api/compute_component), and select the Google Compute Engine API
```
domcleal: missing a full stop at the end of this line

_includes/manuals/1.11/5.2.4_cr_gce.md
```diff
@@ -10,23 +10,24 @@
...
 
-![Initializing Compute Engine](/static/images/screenshots/gce/enable_compute.png)
+![API Manager](/static/images/screenshots/gce/api_manager.png)
 
 Next, under *APIs &amp; auth*, then *APIs*, click the *Enable API* button.
 
 ![Enable API](/static/images/screenshots/gce/enable_api.png)
 
-Under *Credentials*, click *Create new Client ID* in the OAuth section and select *Service account* in the popup box.  A new account will be created for Foreman to use.
+Under *Credentials*, click *Create Credentials -> Create service account key* and choose your service account for Compute Engine.
```
domcleal: Nitpick, use `>` rather than `->` for consistency within the manual.

domcleal:
The enable_compute image is still referenced from the 1.10 and older documentation, so it might be worth either: a) keeping it, or b) add the new screenshots in a different directory so the old instructions keep the old images.

Edit: inline comments apply equally to 1.11 and 1.12 manuals.

_includes/manuals/1.11/5.2.4_cr_gce.md
```diff
@@ -10,23 +10,24 @@
...
 ![Create Project](/static/images/screenshots/gce/create_project.png)
 
-In the project settings, enable the Compute Engine feature by clicking one of the sections below *Compute Engine* in the left hand pane.  This will trigger an initialization job.
+By default, your project will have the Compute Engine and App Engine services enabled.
+Now go to [the API manager](https://console.developers.google.com/apis/api/compute_component), and select the Google Compute Engine API
 
-![Initializing Compute Engine](/static/images/screenshots/gce/enable_compute.png)
+![API Manager](/static/images/screenshots/gce/api_manager.png)
 
 Next, under *APIs &amp; auth*, then *APIs*, click the *Enable API* button.
```
domcleal: `*Enable*` according to the new screenshot

dLobatog:
@domcleal How about I copy this section to all the previous versions? (1.8, 1.9, and 1.10) The old docs are misleading at best, as Google's interface has changed now

domcleal:
Yep, can do.

dLobatog:
Updated for all versions (we should think of a way to extract these common docs and just import them in each version...)

domcleal:
Merged, thanks @dLobatog.

Jekyll supports include statements (it's how the manual's built from manuals/_/_.md) so it's trivial to combine the duplicate content if that makes sense. Generally we copy so it can be adjusted in newer versions only with less effort.

PR #666, opened 2016-07-18
Updated Red Hat logo

lzap:
Kris Bowen from Brand Communication team in Red Hat noticed we are using
incorrect logo for RHEL. There's the official black`n`white type, we are using
"decolorized" standard one. I am replacing it with the correct one.

Also I am changing the name to the official "Red Hat Enterprise Linux" name.


lzap:
This is how it looks like now:

image

dLobatog:
Merged as d833ed25d52f56d706f7f0d7f85237df4ad38b4e, thanks @lzap!

PR #669, opened 2016-07-21
Update Salt plugin URL in section 4.7

karras:
This commit changes the Salt plugin URL referenced in section 4.7 to be version independent and to be aligned with the Chef link in 4.8.


domcleal:
Thanks @karras!

PR #670, opened 2016-07-21
Add section 4.7 Managing Ansible and reorder chapters alphabetically

karras:
This commit adds a separate chapter for the Ansible Plugin and reorders the "4.X Managing XYZ" chapters alphabetically.


domcleal:
Thanks @karras!

PR #673, opened 2016-07-22
add foreman intro slides

timogoebel:
Just some slides I've been using as an intro to Foreman.


GregSutcliffe:
I don't see any objections - merging! Thanks @timogoebel!

PR #683, opened 2016-08-18
Remove birthday frontpage banner

GregSutcliffe:
Sad though I am to see the end of it, it's not our birthday now. I've left the include in place and commented it out so people don't need to go look up the CSS class next time we want to use it.


mmoll:
merged, thanks @GregSutcliffe!

PR #690, opened 2016-09-01
Clustering Foreman steps

oogs:
I realize not everyone will want to cluster foreman, but those of us that want to, can get a significant head start by spelling out a few small steps. I've added those under "Advanced Foreman". This covers some of what Chris Pisano's blog touches on as well as my recent experiences in getting multiple foreman servers to work together. I'm not sure how much hand-holding is necessary for these steps, I can flesh it out more if needed.

### DIFF
--- a/_includes/manuals/1.12/5.8.1_database.md
+++ b/_includes/manuals/1.12/5.8.1_database.md
@@ -1,2 +1,2 @@
 
-All Foreman instances in a cluster/group must point to the same database. This can be done during the initial installation (through flags or altering foreman\_installer\_answers.yaml) or by directly altering {% highlight bash %}$app_root/config/database.yaml{% endhighlight %} (usually a symlink to the real file) and pointing the correct lifecycle to your Foreman DB, then restarting Foreman.
+All Foreman instances in a cluster/group must point to the same database. This can be done during the initial installation (through flags or altering __foreman\_installer\_answers.yaml__) or by directly altering __$app_root/config/database.yaml__ (usually a symlink to the real file) and pointing the correct lifecycle to your Foreman DB, then restarting Foreman.
--- a/_includes/manuals/1.12/5.8.1_database.md
+++ b/_includes/manuals/1.12/5.8.1_database.md
@@ -1,2 +1,2 @@
 
-All Foreman instances in a cluster/group must point to the same database. This can be done during the initial installation (through flags or altering __foreman\_installer\_answers.yaml__) or by directly altering __$app_root/config/database.yaml__ (usually a symlink to the real file) and pointing the correct lifecycle to your Foreman DB, then restarting Foreman.
+All Foreman instances in a cluster/group must point to the same database. This can be done during the initial installation (through flags or altering `foreman_installer_answers.yaml`) or by directly altering `/etc/foreman/database.yaml` and pointing the correct environment (usually production) to your Foreman DB, then restarting Foreman.
--- a/_includes/manuals/1.12/5.8.1_database.md
+++ b/_includes/manuals/1.12/5.8.1_database.md
@@ -1,2 +1,2 @@
 
-All Foreman instances in a cluster/group must point to the same database. This can be done during the initial installation (through flags or altering foreman_installer_answers.yaml) or by directly altering $app_root/config/database.yaml (usually a symlink to the real file) and pointing the correct lifecycle to your Foreman DB, then restarting Foreman.
+All Foreman instances in a cluster/group must point to the same database. This can be done during the initial installation (through flags or altering foreman\_installer\_answers.yaml) or by directly altering {% highlight bash %}$app_root/config/database.yaml{% endhighlight %} (usually a symlink to the real file) and pointing the correct lifecycle to your Foreman DB, then restarting Foreman.
--- a/_includes/manuals/1.12/5.8.1_database.md
+++ b/_includes/manuals/1.12/5.8.1_database.md
@@ -1,0 +1,2 @@
+
+All Foreman instances in a cluster/group must point to the same database. This can be done during the initial installation (through flags or altering foreman_installer_answers.yaml) or by directly altering $app_root/config/database.yaml (usually a symlink to the real file) and pointing the correct lifecycle to your Foreman DB, then restarting Foreman.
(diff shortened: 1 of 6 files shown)


_includes/manuals/1.12/5.8_multiple_foremen.md
```diff
@@ -0,0 +1,4 @@
+
+The following steps are suggested when configuring multiple Foreman instances to work together. They will ensure that data, passwords, and cookies are shared between multiple instances.
+
+__$app_root__ is wherever you installed Foreman, usually __/usr/share/foreman__.
```
domcleal: Prefer using backticks (`) for highlighting paths etc, it uses code style formatting.

_includes/manuals/1.12/5.8.1_database.md
```diff
@@ -0,0 +1,2 @@
+
+All Foreman instances in a cluster/group must point to the same database. This can be done during the initial installation (through flags or altering __foreman\_installer\_answers.yaml__) or by directly altering __$app_root/config/database.yaml__ (usually a symlink to the real file) and pointing the correct lifecycle to your Foreman DB, then restarting Foreman.
```
domcleal: "lifecycle" should probably be "environment", and should probably recommend changing the production env as that's the usual one used.

_includes/manuals/1.12/5.8.2_encryption_rb.md
```diff
@@ -0,0 +1,2 @@
+
+As described in [5.2.10](/manuals/{{page.version}}/index.html#5.2.10PasswordEncryption), passwords stored locally in Foreman's DB are encrypted. In order for multiple Foreman instances to encrypt and decrypt passwords correctly, they all need to have the same encryption key.
```
domcleal: I'd suggest adding the file path here, e.g. "[..] have the same encryption key defined in `/etc/foreman/encryption_key.rb`."

_includes/manuals/1.12/5.8.3_secret_token.md
```diff
@@ -0,0 +1,8 @@
+
+The last file required to make a Foreman cluster work is __$app_root/tmp/secret_token.rb__. This file should contain a secret token that is modeled after __$app_root/config/initializers/local\_secret\_token.rb__. __$app_root/tmp/secret_token.rb__ is used by __$app_root/config/initializers/secret_token.rb__ to sign cookies. Once you have created the new secret_token.rb, restart Foreman and clear Foreman's cache:
```
domcleal: I'm unsure that this file is necessary, it would only have an effect if the local_secret_token.rb initializer was missing. It might be best to only ensure that the local secret token initialiser is correctly generated and in place?

_includes/manuals/1.12/5.8.3_secret_token.md
```diff
@@ -0,0 +1,8 @@
...
+{% highlight bash %}
+foreman-rake tmp:cache:clear
+foreman-rake tmp:sessions:clear
+{% endhighlight %}
+
+Note: The file __$app_root/config/initializers/secret_token.rb__ is uncommented, the logic behind this step rests on suggestions from IRC and testing. Without this change, the user may need to log in multiple times, may run in to "Invalid Authenticity Token"/CSRF issues, or it might just work for no clear reason.
```
domcleal: I don't think these comments about whether it may or may not work, or how you heard about it should be in the manual. The info about the symptom may be useful though.

_includes/manuals/1.12/5.8.3_secret_token.md
oogs: I wasn't aware of this file being necessary at first, however, creating `$app_root/tmp/secret_token.rb` and setting it to the same value across all my Foreman servers did solve my issue: https://groups.google.com/forum/?fromgroups#!topic/foreman-users/OeOvX_IBkcw . I've had mixed results, I do have multiple dev instances without this file that work (work = can log in w/o Invalid Authenticity Token warning), yet my bigger prod instance doesn't work without this file. If you'd like, we can chat about this on IRC a bit, but I can tell you that `$app_root/config/initializers/local_secret_token.rb` was present and set the same across all servers before I tried secret_token.rb.

domcleal: Hm, that's peculiar. I'd suggest a couple of things:
1. If you'd like to investigate the issue, adding debugging to the top of secret_token.rb might help: `File.open(Rails.root.join('tmp', 'token.log'), 'a') { |f| f.puts "secret_token=#{Foreman::Application.config.secret_token}" }`, then check tmp/token.log matches the token set in local_secret_token.rb. It should be set if the file exists. I'd like to avoid recommending using tmp/secret_token if possible.
2. Failing that, this should recommend that local_secret_token.rb is set identically across all servers, and that `tmp/secret_token` is also set identically if it exists.
3. Also, `tmp/secret_token.rb` should have the .rb extension removed, it's just `tmp/secret_token`

domcleal:
Thanks for the update, comment added in the "outdated" discussion.

_includes/manuals/1.12/5.8.3_secret_token.md
oogs: I think that point # 3 disproves 5.8.3 - if it's not reading `tmp/secret_token.rb` because of the extension, then it's reading the local_secret_token that I had set the same across the board. Let me re-do my testing since I can replicate this behavior in my dev environment now, it could be that I made two changes at once and attributed success to the second change.

oogs: You're right. I tried point 1 and it's showing the token in my local_secret_token, not the one in secret_token. I'll amend the entry shortly.

_includes/manuals/1.12/5.8.3_secret_token.md
```diff
@@ -0,0 +1,9 @@
+
+The last file required to make a Foreman cluster work is `$app_root/config/initializers/local_secret_token.rb`. This is used by `$app_root/config/initializers/secret_token.rb` to sign cookies. This should be set the same across all Foreman servers in your cluster. Once you have set `local_secret_token.rb`, restart Foreman and clear Foreman's cache:
```
domcleal: The "used by ".../secret_token.rb" is obsolete here, you could just say ", which is used to sign cookies."?

Thanks for checking it and updating, that's much better!

domcleal:
Thanks @chbaldwi!

PR #696, opened 2016-09-12
Document Puppet Server 2.6 compatibility and workaround in 1.12

domcleal:


dLobatog:
Merged as 28df46f7eba991430a6146b7b3f02eb8a7b631da, thanks @domcleal!

PR #697, opened 2016-09-13
Document 1.13 Foreman/smart proxy settings

domcleal:
Also corrects name of the native MS DHCP provider, per
https://github.com/theforeman/smart-proxy/commit/36956f8.


mmoll:
thanks @domcleal!

Issue #699, opened 2016-09-14
npm install step is missing from the manual


gnurag:
Under **Installing from source** section of the manual, npm install step is missing. This step is correctly mentioned in contributing.html page however.

PR #700, opened 2016-09-14
(#699) npm install step is missing from the manual

gnurag:
- Added npm install command under Installing from source section of the manual (#699)
- Updated GNU GPL license name to use its correct form


domcleal:
Thanks, a few points:
- Please add npm to the dependencies towards the top of the section
- `rake webpack:compile` is also required prior to starting the Rails server
- ditto in the upgrade section (3.6), suggest running `npm up` or similar, plus webpack:compile

gnurag:
Hi @domcleal, the upgrade section does not have a separate section for source-only installations so haven't tried that yet.

domcleal (changes requested):
The upgrade section has "Step 2 (D) - git checkouts only" which is for source installations, the step(s) can go in there.

There also seem to be unrelated commits that are already on `gh-pages` in your branch, please remove them.

gnurag:
Closing this PR and opening a fresh clean PR.

PR #704, opened 2016-09-20
Update installer options against 1.13.0-RC1

domcleal:
Parameters themselves are auto-generated docs.


mmoll:
merged, thanks @domcleal!

PR #675, opened 2016-07-26
Clarify EL6 removal for Foreman 1.13 in release notes

domcleal:

PR #705, opened 2016-09-20
Remove EL6 installation instructions

domcleal:


ohadlevy:
@domcleal EL6 PR (https://github.com/theforeman/foreman-packaging/pull/1344) according to @ehelms is nearly done. hence, IMHO we are keeping EL6 support for one more release.

domcleal:
We're not, 1.13.0-RC1 doesn't contain EL6 packages, it missed it by a _long_ way. This is correct.

ohadlevy:
I respectively disagree.

domcleal:
Like with #675, I'd prefer that the docs are accurate, whether or not you personally agree with the change. If the situation changes again then the docs can be updated trivially.

Can I reopen this PR and ask please for a re-review?

PR #709, opened 2016-09-23
fixes #16669 - correct filenames for Puppet SSH config editing

domcleal:


mmoll:
merged, thanks @domcleal!

PR #712, opened 2016-09-27
Adding blogpost about IBM POWER

lzap:


dLobatog:
Thanks @lzap !

lzap:
Damn the filename is named "new-post". Let's keep it (because of RSS), I used rake task to create the file. Strange.

PR #713, opened 2016-10-05
Expand IRC guidelines to cover other communication, and add note on reporting violations

GregSutcliffe:
Mostly a restructuring and rewording to change IRC specific language, but I added a couple of things to the mailing list section - I stopped short of recommending no-HTML, no-top-post, but we could add that if people want....

### DIFF
--- a/support.md
+++ b/support.md
@@ -6,6 +6,28 @@ title: Support
 # Support
 ------------------------------
 
+## Community Guidelines
+
+We have a variety of communication tools that the Foreman Community uses to get
+work done. The guidelines below apply in general when interacting with other
+Foreman users & developers:
+
+* Be nice: Treat people with respect and consideration.
+* Be helpful: Be patient with new people and be willing to jump in to answer questions.
+* Be patient: People often have many other things to do, so allow time for your question to be answered
+* Stay calm: When in doubt, assume the best intentions and do not let emotions get out of control
+* Try to keep conversations on-topic where possible - going off on a tangent doesn't solve anyone's problem.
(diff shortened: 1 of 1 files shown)


support.md
```diff
@@ -6,6 +6,28 @@ title: Support
...
+## Community Guidelines
+
+We have a variety of communication tools that the Foreman Community uses to get
+work done. The guidelines below apply in general when interacting with other
+Foreman users & developers:
+
+* Be nice: Treat people with respect and consideration.
+* Be helpful: Be patient with new people and be willing to jump in to answer questions.
+* Be patient: People often have many other things to do, so allow time for your question to be answered
+* Stay calm: Give people the benefit of the doubt and try not to let emotions get out of control.
```
mmoll: I'm not a native English speaker, but I'd rather like to see something along the lines of "When in doubt, assume the best intentions and do not let emotions get out of control."

support.md
```diff
@@ -34,6 +56,19 @@ There is also a low-traffic list dedicated to release and security announcements
 
 * [foreman-announce](https://groups.google.com/forum/?fromgroups#!forum/foreman-announce) or email `foreman-announce+subscribe@googlegroups.com`
 
+### Mailing List Guidelines
+
+In addition to the [general guidelines](/support.html#CommunityGuidelines) above, we also have a few IRC-specific guidelines:
```
mmoll: copypasto: "IRC-specific"

support.md
GregSutcliffe: I like that better too, updating.

support.md
GregSutcliffe: Gah, note to self - spell-checking isn't everything... thanks!

dLobatog:
Thanks @GregSutcliffe @mmoll !

PR #714, opened 2016-10-05
email.yml is actually email.yaml

traylenator:
Not sure if you fix back versions of documentation.


domcleal:
Bug fixes like this might as well be fixed in current/recent versions as well, if you like? (i.e. 1.12 too)

traylenator:
Fixed in >= 1.11 now.

domcleal:
Thanks @traylenator.

PR #716, opened 2016-10-11
add instructions for external database servers

mmoll:


GregSutcliffe (approved):

_includes/manuals/1.12/3.2.3_installation_scenarios.md
```diff
@@ -1,6 +1,20 @@
 
 The Foreman installer can accommodate more complex, multi-host setups when supplied with appropriate parameters.
 
+#### Using an external database server
+
+Per default foreman-installer will install a PostgreSQL database server onto the Foreman host and create its database. An external MySQL or PostgreSQL database server with an already created, empty database can be used with the following arguments:
```
domcleal: An empty DB isn't entirely correct, since the user will have to run db:migrate/seed against it too. You might want to include this as a post-installation step?

domcleal (changes requested):

_includes/manuals/1.12/3.2.3_installation_scenarios.md
mmoll: Indeed, added. https://github.com/theforeman/puppet-foreman/issues/473 is related.

_includes/manuals/1.12/3.2.3_installation_scenarios.md
```diff
@@ -1,6 +1,26 @@
...
+  --foreman-db-host=dbserver.example.com \
+  --foreman-db-database=dbname \
+  --foreman-db-username=dbuser \
+  --foreman-db-password=dbpassword
+{% endhighlight %}
+
+As a post-installation step, to populate the database correctly, run:
+
+{% highlight bash %}
+foreman-rake db:migrate db:seed apipie:cache:index
```
domcleal: Please split the db:migrate/seed into separate commands, we generally don't test them being run in the same process (db:seed + apipie:cache:index should be fine) as db:migrate makes a lot of in-process changes.

domcleal (changes requested):

_includes/manuals/1.12/3.2.3_installation_scenarios.md
mmoll: done.

domcleal:
Thanks @mmoll.

PR #717, opened 2016-10-12
Update foreman-remote-execution/nightly/index.md

duncaninnes:
Removed reference specifying Foreman 1.10 repository to install from.  Also removed reference to EPEL as this is configured with a successful Foreman install.


domcleal:
Merged, thanks @duncaninnes.

PR #731, opened 2016-11-07
Move styleguide

zjhuntin:
This PR creates a style guide in the documentation page that is based on our guides from Katello. If there is more that I can integrate from the Foreman side please let me know.


domcleal:
There already _is_ a detailed style guide in the handbook, why is this needed? Katello's docs should be on their own website.

PR #734, opened 2016-11-09
indicate use of rake db:test:prepare for testing

thomasmckay:


contribute.md
```diff
@@ -62,7 +62,7 @@ If you experience problems related to therubyracer or libv8, you can install wit
 
 ### Setup test environment
 1. Copy `config/database.yml.example` to `config/database.yml`
-1. Create your database: `bundle exec bin/rake db:migrate`
+1. Create your database: `bundle exec bin/rake db:test:prepare`
```
domcleal, line 65: db:test:prepare is not sufficient, as it would require schema.rb generated by db:migrate. db:test:prepare is also unnecessary with `rake test`.

domcleal (changes requested):

PR #745, opened 2016-11-16
Update foreman-remote-execution/nightly/index.md

duncaninnes:
Edit the port used to curl the pubkey from the API to reflect the new default of 9090


adamruzicka:
I'm sorry, I might have mislead you a bit in my email about this. According to the Foreman[1] and Katello[2] docs, the default port used by smart proxy for https is different in Foreman-only and Foreman+Katello deployments.

[1] - https://theforeman.org/manuals/1.13/index.html#3.InstallingForeman section 3.1.4
[2] - http://www.katello.org/docs/3.2/installation/capsule.html

duncaninnes:
No problem. I'll see if I can come up with a suitable wording to cater for both options.

PR #768, opened 2016-12-08
Fix confusing SSH instructions for Azure

dLobatog:
The instructions literally say to use the image name to SSH into the instance, it's the username that should be used.


mmoll:
merged, gracias @dLobatog!

PR #772, opened 2016-12-12
Update partition tables documentation

Klaas-:
* explain how dynamic partition tables work
* change Dynamic Partition tables section to Per-Host Partition tables

### DIFF
--- a/_includes/manuals/1.14/4.4.4_prov_ptables.md
+++ b/_includes/manuals/1.14/4.4.4_prov_ptables.md
@@ -3,6 +3,15 @@ Partition templates are a subset of normal Provisioning Templates. They are hand
 
 Foreman comes with pre-created templates for common Operating Systems, but it is good to edit the template, check it's content and it's Family setting. If the Family is wrong, be sure to go back to Operating Systems and associate it with your Operating System.
 
-#### Dynamic Partition tables
+#### Per-Host Partition tables
 
-When creating a new Host, you will be given the option to create a Dynamic Partition table. This is essentially a 'one-off' partition table that is stored with the host and used only for that host. It replaces the choice of Partition Table from the normal list of those associated with the selected OS.
+When creating a new Host, you will be given the option to create an individual partition table. This is essentially a 'one-off' partition table that is stored with the host and used only for that host. It replaces the choice of Partition Table from the normal list of those associated with the selected OS.
+
+#### Dynamic Partition tables
+Some operating systems allow you to create partition tables via scripts. At the moment Kickstart and AutoYaST based systems can use this feature. Partition templates starting with #Dynamic are interpreted as scripts rather than static partition tables. The Provisioning Template needs to support this feature (search for @dynamic). This enables you to make choices on the fly during provisioning (or re-provisioning).
+
+##### Kickstart Dynamic Partition tables
+Kickstart will run dynamic partition tables as a pre-install bash script. This script needs to create a complete Kickstart partition table in '/tmp/diskpart.cfg'. This partition table will then be read by anaconda for the installation.
+
+##### AutoYaST Dynamic Partition tables
+AutoYaST will run dynamic partition tables as a pre-install bash script. This script needs to create a AutoYaST XML file in '/tmp/profile/modified.xml'. The modified.xml file will be read by YaST after your script has finished. An example for getting the same functionality as with Kickstart would be to create your XML partition table in /tmp/diskpart.cfg and sed it together with the original AutoYaST XML like this: ```sed '/<\/ntp-client>/ r /tmp/diskpart.cfg' /tmp/profile/autoinst.xml > /tmp/profile/modified.xml```
--- a/_includes/manuals/1.14/4.4.4_prov_ptables.md
+++ b/_includes/manuals/1.14/4.4.4_prov_ptables.md
@@ -40,7 +40,7 @@ EOF
 ```
 
 ##### AutoYaST Dynamic Partition tables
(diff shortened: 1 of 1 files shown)


_includes/manuals/1.14/4.4.4_prov_ptables.md
```diff
@@ -3,6 +3,15 @@ Partition templates are a subset of normal Provisioning Templates. They are hand
 
 Foreman comes with pre-created templates for common Operating Systems, but it is good to edit the template, check it's content and it's Family setting. If the Family is wrong, be sure to go back to Operating Systems and associate it with your Operating System.
 
+#### Per-Host Partition tables
+
+When creating a new Host, you will be given the option to create an individual partition table. This is essentially a 'one-off' partition table that is stored with the host and used only for that host. It replaces the choice of Partition Table from the normal list of those associated with the selected OS.
+
 #### Dynamic Partition tables
+Some operating systems allow you to create partition tables via scripts. At the moment Kickstart and AutoYaST based systems can use this feature. Partition templates starting with #Dynamic are interpreted as scripts rather than static partition tables. The Provisioning Template needs to support this feature (search for @dynamic). This enables you to make choices on the fly during provisioning (or re-provisioning).
```
domcleal: Use backticks around `#Dynamic` and `@dynamic` so it's clear that it's literal.

_includes/manuals/1.14/4.4.4_prov_ptables.md
```diff
@@ -3,6 +3,15 @@ Partition templates are a subset of normal Provisioning Templates. They are hand
...
+
+##### Kickstart Dynamic Partition tables
+Kickstart will run dynamic partition tables as a pre-install bash script. This script needs to create a complete Kickstart partition table in '/tmp/diskpart.cfg'. This partition table will then be read by anaconda for the installation.
```
domcleal: 1. "as a pre-install bash script using a %pre scriplet."
1. ".. read by anaconda for the installation by using `%include /tmp/diskpart.cfg`." (to show it's not automatic)

domcleal (changes requested):
Generally good, perhaps consider including a very small/trivial example of a dynamic script to show the placement of the comment and the sort of thing that can be done. (i.e. content from http://projects.theforeman.org/projects/foreman/wiki/Dynamic_disk_partitioning).

Klaas-:
Okay, I added examples and enhanced the wording, squash on commit please if there are no other suggestions

_includes/manuals/1.14/4.4.4_prov_ptables.md
```diff
@@ -3,6 +3,142 @@ Partition templates are a subset of normal Provisioning Templates. They are hand
...
+zerombr yes
+clearpart --all --initlabel
+part swap --size "$vir_mem" 
+part /boot --fstype ext3 --size 100 --asprimary
+part / --fstype ext3 --size 1024 --grow
+EOF
+```
+
+##### AutoYaST Dynamic Partition tables
+AutoYaST will run dynamic partition tables as a pre-install bash script. This script needs to create a AutoYaST XML file in ```/tmp/profile/modified.xml```. The modified.xml file will be read by YaST after your script has finished. An example for getting the same functionality as with Kickstart would be to create your XML partition table in ```/tmp/diskpart.cfg``` and sed it together with the original AutoYaST XML like this: ```sed '/<\/ntp-client>/ r /tmp/diskpart.cfg' /tmp/profile/autoinst.xml > /tmp/profile/modified.xml```. Inserting after ntp-client section is just a suggestion that uses the same style the community-tamplates do. The following example is based on ```disklayout_lvm.erb```.
```
domcleal: 1. "tamplates" -> "templates"
1. Not sure the line about this being a suggestion is necessary.
1. "disklayout_lvm.erb" is an internal filename within the codebase, I don't think it's helpful to reference it. Either use the name or remove that comment.

_includes/manuals/1.14/4.4.4_prov_ptables.md
```diff
@@ -3,6 +3,142 @@ Partition templates are a subset of normal Provisioning Templates. They are hand
...
+          <partition_nr config:type="integer">2</partition_nr>
+          <size>max</size>
+        </partition>
+      </partitions>
+      <pesize></pesize>
+      <type config:type="symbol">CT_DISK</type>
+      <use>all</use>
+    </drive>
+    <drive>
+      <device>/dev/vg00</device>
```
domcleal, line 94: Can you reduce the size of this example? My point was that it should be a small or very trivial example that's a few lines long, not a complete piece of code if possible - this is very large.

The points it should get across are:

1. `#Dynamic` header
1. Generate the modified file.

You could probably delete this second device and many of the separate partitions.

domcleal (changes requested):

_includes/manuals/1.14/4.4.4_prov_ptables.md
Klaas-: this is the default community-templates example for an lvm host, but yeah I can make it shorter

_includes/manuals/1.14/4.4.4_prov_ptables.md
Klaas-: Will fix typo and change to name of that template

Klaas-: and that line about the suggestion - I wanted to make clear that its an xml document, so it doesn't matter where you include the section :)

_includes/manuals/1.14/4.4.4_prov_ptables.md
domcleal: Yeah, no need to have the entire default here. If a user needs the default, it's in their installation.

Klaas-:
@domcleal is this acceptable?

domcleal (approved):

domcleal:
Merged, thank you @Klaas-.

PR #773, opened 2016-12-14
Changing use_puppet_default to omit

orrabin:


domcleal:
Would you mind also fixing section 4.2.5 which contains references to "Use Puppet default"? It may need an updated screenshot too, sorry.

domcleal (approved):

domcleal (changes requested):
Please also update the name given in the "Facts, reporting and parameters" release notes category above.

orrabin:
@domcleal I also changed the screenshots, most of them contained use_puppet_default and the rest I changed because of the difference in css.
I added a new folder for 1.14 because I didn't want to change just part of the 1.11 screenshots. I could change all of the 1.11 parameter schreenshots and then use those that don't contain use_puppet_default here, should I do that or open a separate pr?

domcleal (approved):

domcleal:
What you've done is great, just replace other screenshots in the 1.14 directory as and when we can. No need to change previous release documentation at this stage.

domcleal:
Merged, thanks @orrabin.

PR #778, opened 2016-12-21
Fix confusing SSH instructions for Azure

dLobatog:


domcleal:
This appears identical to 69828fd, I think it's already in gh-pages.

mmoll:
indeed.

PR #766, opened 2016-12-22
PXE loader option for unattended installation

lzap:
#12634 added a PXE loader option to host creation which changes the boot filename specified in the DHCP reservation, to be used to support different architectures and boot modes. This should be documented in the unattended installation section.


lzap:
Added the notes into 1.14 for now.

mmoll:
merged, as 2cf9de62 with 1.13 notes added, díky @lzap!

PR #781, opened 2016-12-22
Smart proxy requirements/config changes for 1.14

domcleal:


mmoll:
merged, thanks @domcleal!

PR #782, opened 2016-12-22
Document 'audit' logger for fact import statistics

domcleal:


mmoll:
merged, thanks @domcleal!

PR #786, opened 2016-12-30
Puppet 4.4+ environment_classes in auth.conf

dLobatog:
Puppet versions 4.4 and newer need a section to allow
environment_classes in auth.conf.

Adding these rules here will not be a problem for people using Puppet older than 4.4, as those paths will not resolve to anything. For people who use newer Puppet versions, it will allow them to import classes.

### DIFF
--- a/_includes/manuals/1.14/4.3.6_smartproxy_puppet.md
+++ b/_includes/manuals/1.14/4.3.6_smartproxy_puppet.md
@@ -50,6 +50,10 @@ path /puppet/v3/environments
 method find
 allow *
 
+path /puppet/v3/environment_classes
+method find
+allow *
+
 path /puppet/v3/resource_type
 method search
 allow *
@@ -67,6 +71,16 @@ The [HOCON-formatted auth.conf style](https://docs.puppet.com/puppetserver/lates
     allow: "*"
     sort-order: 500
     name: "puppetlabs environments"
+},
+{
+    match-request: {
+    path: "/puppet/v3/environment_classes"
+       type: path
+       method: get
+    }
+    allow: "*"
+    sort-order: 500
+    name: "puppetlabs environment classes"
 },
 {
     match-request: {


_includes/manuals/1.13/4.3.6_smartproxy_puppet.md
```diff
@@ -50,6 +50,10 @@ path /puppet/v3/environments
 method find
 allow *
 
+path /puppet/v3/environment_classes
```
domcleal: Unsure if this is applicable to 1.13, isn't this only used on 1.14?

domcleal (changes requested):

_includes/manuals/1.13/4.3.6_smartproxy_puppet.md
dLobatog: Correct - I'll change it

dLobatog:
@domcleal Updated to remove the changes in 1.13 as the proxy doesn't call v3/environment_classes in that version, thanks for taking a look

_includes/manuals/1.14/4.3.6_smartproxy_puppet.md
```diff
@@ -70,6 +74,16 @@ The [HOCON-formatted auth.conf style](https://docs.puppet.com/puppetserver/lates
 },
 {
     match-request: {
+    path: "/puppet/v3/environment_classes"
+       type: path
+       method: get
+      }
```
domcleal: Could you fix the indentation to match the other blocks here, and lines 82/83? (remove 3 leading spaces)

domcleal (changes requested):

dLobatog:
@domcleal Updated to remove the leading spaces, sorry!

domcleal (approved):

domcleal:
Merged, thanks @dLobatog.

PR #788, opened 2017-01-04
Fix Katello Puppet integration image links

dLobatog:
The current links didn't display because the URL is wrong


ehelms (approved):

ehelms:
Thanks @dLobatog !

PR #790, opened 2017-01-04
List networking capability for each CR

domcleal:
Includes listing of new IPv6 support for Rackspace.


mmoll:
merged, thanks @domcleal!

PR #792, opened 2017-01-04
Remove EL6 from Quickstart intro

domcleal:
Missed from f5336a7 when removing it from the dropdowns.


mmoll:
merged, thanks @domcleal!

PR #796, opened 2017-01-10
Add 3.3 Release Notes

ehelms:
Stub for 3.3 release notes, taken from (https://goo.gl/AqRGv3). @jlsherrill mind doing a review of possible release note worth items as well? Will work on/seek out descriptions after we settle.

PR #798, opened 2017-01-10
Add 3.3 Release Notes

bbuckingham:
This PR will replace: https://github.com/theforeman/theforeman.org/pull/796

Stub for 3.3 release notes, taken from (https://goo.gl/AqRGv3). @jlsherrill mind doing a review of possible release note worth items as well? Will work on/seek out descriptions after we settle.

### DIFF
--- a/plugins/katello/3.3/release_notes/release_notes.md
+++ b/plugins/katello/3.3/release_notes/release_notes.md
@@ -4,17 +4,83 @@ title: Release Notes
 version: 3.3
 ---
 
-# Katello Nightly () Release Notes
+# Katello 3.3 (Baltic Porter) Release Notes
 
-{% if page.version and page.version != '3.3' %}
 For the full release notes, see the [Changelog](https://github.com/Katello/katello/blob/KATELLO-{{ page.version }}/CHANGELOG.md).
-{% else %}
-For the full release notes, see the [Changelog](https://github.com/Katello/katello/blob/master/CHANGELOG.md).
-{% endif %}
 
 ## Features
 
+### Updated UI
+
+Simplify UI by moving towards a single item view in accordance with patternfly best practices and in preparation for upcoming vertical navigation.
+
+### Improved Manifest Import and Delete
+
+Manifest import and delete actions have been refactored in anticipation of Candlepin 2.0. Now users are able to see a step-by-step breakdown of tasks for these actions in the Dynflow console.
+
+### Username and Password for Repositories
+
+Set the username and password on repositories that require authentication for access during synchronization.
+
+### Package Applicability
+
+See what packages need updating on your host. Previously only available via repositories with Errata, now you can see exactly what updates are available for each host and apply them directly.
+
+### Content View Enhancements
+
+ * Package Filtering by Architecture: Create package filters on yum content to either include or exclude packages based upon their architecture (e.g. noarch, x86_64, i386...etc).
+
+ * Promote Version to Multiple Environments: Promote a content view version to multiple environments at once. Prior to 3.3, users could only promote a content view version to one environment at a time. This feature is currently only available via the API and CLI.
+
+ * Publish "Latest" Version in Composite Content Views: When configuring a Composite Content View a specific version of each component is selected. Now you can simply select 'Latest' and at publish time the latest version will be calculated and used within the composite. 
+
+ * Define and publish Composite Content Views Containing Overlapping Repositories: When configuring Composite Content Views, you can now select component views that were published from the same source repository.  Prior to Katello 3.3, if two component views were published using the same source repository (e.g. Centos7), they could not both be added to the same composite content view.
+
+### Smart Proxy Content
+
+Smart Proxies with content can now be configured with their own download policy. This allows the main server to synchronize the full contents of its configured repositories while the Smart Proxy can be configured to use an "On Demand" policy and only synchronize what is actually used by clients.
+
+### Tracer Support
+
+After updating packages on a host, a service will some times need to be restarted in order for the new version to take affect. Tracer is a utility to help detect these services and Katello 3.3 supports hosts reporting this information. To use this feature, simply update to the newest katello-agent and the reports will be sent whenever a package is updated. You can see what services need to be restarted and restart them via Remote Execution from the new Content Host Details Traces page.
+
 ## Deprecation Warnings
+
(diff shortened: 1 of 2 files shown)


plugins/katello/3.3/release_notes/release_notes.md
```diff
@@ -4,16 +4,26 @@ title: Release Notes
...
+### Username and Password for Repositories
+
+### Content View Version Promotion to Multiple Environments
+
+### Publish "Latest" Version in Composite Content Views
+
+### Capsule Download Policy
+
+### Tracer Support
+
```
jlsherrill, line 44: After updating some packages on a host sometimes a service will need to be restarted in order for the new version to take affect.  Tracer is a utility to help detect these services and Katello 3.3 supports hosts reporting this information.  Simply update to the newest katello-agent and these reports will be sent whenever a package is updated.  You can see what services need to be restarted and restart them via Remote Execution from the new Content Host Details Traces page.

plugins/katello/3.3/release_notes/release_notes.md
```diff
@@ -4,16 +4,26 @@ title: Release Notes
...
+### Improved Manifest Import and Delete
+
+### Username and Password for Repositories
+
+### Content View Version Promotion to Multiple Environments
+
+### Publish "Latest" Version in Composite Content Views
+
+### Capsule Download Policy
```
jlsherrill: Actually this should be called "Smart Proxy Content".  In fact we should probably include this as a feature so its more clear. 

Smart Proxies with content can now be configured with their own download policy.  This allows the main server to sync the full contents of its configured repositories while the Smart Proxy can be configured to use an "On Demand" policy and only sync what is actually used by clients.

plugins/katello/3.3/release_notes/release_notes.md
```diff
@@ -4,16 +4,26 @@ title: Release Notes
...
 
 ## Features
 
+### Updated UI
+
+### Improved Manifest Import and Delete
+
+### Username and Password for Repositories
+
+### Content View Version Promotion to Multiple Environments
```
jlsherrill: You might group this one and the next with "Content View Improvements" and add:

* Arch available for content view package filters
* Allow Composite content views to be publish containing components with different copies of the same repositories, combining the contents as part of the composite.

plugins/katello/3.3/release_notes/release_notes.md
```diff
@@ -4,16 +4,26 @@ title: Release Notes
...
-# Katello Nightly () Release Notes
+# Katello 3.3 (Baltic Porter) Release Notes
 
-{% if page.version and page.version != '3.3' %}
 For the full release notes, see the [Changelog](https://github.com/Katello/katello/blob/KATELLO-{{ page.version }}/CHANGELOG.md).
-{% else %}
-For the full release notes, see the [Changelog](https://github.com/Katello/katello/blob/master/CHANGELOG.md).
-{% endif %}
 
 ## Features
```
jlsherrill, line 12: New Feature:  Package Applicability

See what packages need updating on your host.  Previously only available via repositories with Errata, now you can see exactly what updates are available for each host and apply them directly.

plugins/katello/3.3/release_notes/release_notes.md
daviddavis: Promotion to multiple environments allows users to promote a content view to multiple environments at once. Before 3.3, users could only promote a content view version to one environment at a time. Right now this feature is only available via the API and CLI.

plugins/katello/3.3/release_notes/release_notes.md
```diff
@@ -4,16 +4,26 @@ title: Release Notes
...
 
+### Updated UI
+
+### Improved Manifest Import and Delete
```
daviddavis, line 17: Manifest import and delete actions were refactored for 3.3 in anticipation of Candlepin 2.0. Now users should be able to see a step-by-step breakdown of tasks for these actions in the Dynflow console.

plugins/katello/3.3/release_notes/release_notes.md
```diff
@@ -4,16 +4,26 @@ title: Release Notes
...
+
+### Username and Password for Repositories
+
+### Content View Version Promotion to Multiple Environments
+
+### Publish "Latest" Version in Composite Content Views
+
```
jlsherrill, line 34: When configuring a Composite Content View a specific version of each component is selected.  Now you can simply select 'Latest' and at publish time the latest version will be calculated and used within the composite.

plugins/katello/3.3/release_notes/release_notes.md
bbuckingham: Added

plugins/katello/3.3/release_notes/release_notes.md
bbuckingham: Added

plugins/katello/3.3/release_notes/release_notes.md
```diff
@@ -4,16 +4,42 @@ title: Release Notes
...
 ## Features
 
+### Updated UI
+
+### Improved Manifest Import and Delete
+
+Manifest import and delete actions were refactored for Katello 3.3 in anticipation of Candlepin 2.0. Now users are able to see a step-by-step breakdown of tasks for these actions in the Dynflow console.
+
+### Username and Password for Repositories
+
```
parthaa, line 22: Can set the username/password to access upstream yum/puppet repositories or docker registries that require http basic authentication.

parthaa: Or may be

 Allows for the setting of username and password when syncing authenticated http and https repositories

plugins/katello/3.3/release_notes/release_notes.md
```diff
@@ -4,16 +4,26 @@ title: Release Notes
...
 
-{% if page.version and page.version != '3.3' %}
 For the full release notes, see the [Changelog](https://github.com/Katello/katello/blob/KATELLO-{{ page.version }}/CHANGELOG.md).
-{% else %}
-For the full release notes, see the [Changelog](https://github.com/Katello/katello/blob/master/CHANGELOG.md).
-{% endif %}
 
 ## Features
 
+### Updated UI
```
waldenraines, line 13: Simplify UI by moving towards a single item view in accordance with patternfly best practices and in preparation for vertical for upcoming vertical navigation.

bbuckingham: Added

jlsherrill:
ACK

PR #803, opened 2017-01-23
Fix rendering of sub-headings

domcleal:
Spacing is required between the level three and four headings.


dLobatog:
Thanks @domcleal

PR #804, opened 2017-01-24
Add survey to frontpage

GregSutcliffe:
Same as for the Birthday stuff, looks OK in my browser.


dLobatog:
Looking good on f25 Firefox 50 and Chrome - merging.

PR #810, opened 2017-02-03
Add foreman-tasks manual

ares:
@iNecas or @adamruzicka could you please check that I'm not making up something? :-)


plugins/foreman_tasks/0.8/index.md
```diff
@@ -0,0 +1,302 @@
...
+
+## 3.4 Recurring logics
+
+Some tasks can be set to repeat based on regular basis. Recurring logic is the definition of such repeating. Note that the task is a single run, recurring logic creates new task instance for every iteration. Existing recurring logics can be found under Monitor -> Recurring Logics. Following list explains the logic attributes.
+
+- **Cron line** - recurring representation in cron-like syntax 
+- **Action** - human readable name of the task that is being repeated 
+- **Current iteration** - how many times has the task already run  
+- **Iteration limit** - if the task has been configured to run specific number of times, the number is displayed here 
+- **Repeat until** - another way of specifying when the reoccurring should stop is configuring a deadline, it can be combined with iteration limit in which case the first that happens stops the repetition
```
adamruzicka: s/reoccuring/recurring

plugins/foreman_tasks/0.8/index.md
```diff
@@ -0,0 +1,302 @@
...
+If you want to share the the history of tasks with someone else who does not have access to your Foreman instance, there's a tool to do the export. This is especially useful when developers asks you for debugging some debugging data. To do the export, use to following rake task provided by the plugin
+
+    foreman-rake foreman_tasks:export_tasks
+
+This will create gzipped tar file in temp directory containing html page for each exported task. The path to the file is printed after the export is finished. These data contains all tasks there are present in database. To limit what exact tasks should be part of the export, use following variables when running the task.
+
+- **TASK_SEARCH** - scoped search filter (example: 'label = "Actions::Foreman::Host::ImportFacts"')
+- **TASK_DAYS** - number of days to go back
+
+To explicitly name the export, you can use **TASK_FILE** variable. The following example would export all facts importing tasks from last week into file /tmp/last_week_impotring.tar.gz
```
adamruzicka: s/impotring/importing

plugins/foreman_tasks/0.8/index.md
```diff
@@ -0,0 +1,302 @@
...
+        :actions:
+          - :name: Actions::Foreman::Host::ImportFacts
+            :after: 10d
+
+The foreman_tasks:cleanup script also accepts additional parameters to specify the search criteria for the cleanup manually:
+
+- **TASK_SEARCH** - scoped search filter (example: 'label = "Actions::Foreman::Host::ImportFacts"')
+- **AFTER** - delete tasks created after AFTER period. Expected format is a number followed by the time unit (s, h, m, y), such as 10d for 10 days (applicable only when the TASK_SEARCH option is specified)
+- **STATES** - comma separated list of task states to touch with the cleanup, by default only stopped tasks are affected (applicable only when the TASK_SEARCH option is specified)
+- **NOOP** - set to "true" if the task should not actuall perform the deletion, only report the actions the script would perform
```
adamruzicka: s/actuall/actually

adamruzicka (changes requested):
Very well written, +1. However I noticed some small typos in the second half

ares:
thanks @adamruzicka I fixed what you commented on

plugins/foreman_tasks/0.8/index.md
```diff
@@ -0,0 +1,302 @@
...
+- **End at** - indicates the time when the task stopped
+- **Start before** - indicates the deadline after which the task should not start if it hasn't started yet
+
+- **Progress** - the task may consist of multiple steps or sub-tasks, the progress is calculated based on percentage of finished steps/sub-tasks
+
+Note that you get more information by hovering some of the attribute values.
+
+### 3.2.1 Results and State related Q&A
+
+- Q: Why I see paused task?
```
ohadlevy: Why do I see .. ?

plugins/foreman_tasks/0.8/index.md
```diff
@@ -0,0 +1,302 @@
...
+- A: Most common cause is that the task was programmed to pause on errors and some error occurred. On the other hand, there might be tasks that are paused because they are waiting to be resumed by some external service when it's ready. To find out more information, check if you see *Errors* field on task page or red box under the Errors tab. If you see something, the pause was very likely caused by an error.
```
ohadlevy: IMHO swap the condition, e.g.
an error has occurred and it was programmed to pause on error

plugins/foreman_tasks/0.8/index.md
```diff
@@ -0,0 +1,302 @@
...
+
+- Q: What is the pending result exactly?
+- A: While the task is in progress we can't tell what is the result, in such case the result is *pending*
```
ohadlevy: can you add examples of why you wont know it?

plugins/foreman_tasks/0.8/index.md
```diff
@@ -0,0 +1,302 @@
...
+
+- Q: Where do I find the error that I should report if the task failed?
+- A: The best place to look is the task detail page. In the first tab, you will see the human readable message based on the exception that was raised. The format depends on author of the task. If the task is based on Dynflow, you can find more information by navigating to the Errors tab on the same page. If you report an issue, make sure you attach the content of this tab with other logs. Note that the input and output fields may contain sensitive data so make sure you clean all such data while uploading it.
+
+- Q: There is a task with state *paused* and result *error*, what does that mean?
+- A: This indicates that the task was paused because of some error occurred. In task detail you should see question mark at *Result* field but in table, we display the value matching to what the result of the task would be, if we stoped it right now. Thanks to this you can also easily search tasks that needs attention by filtering on result value. The same applies for warning result. The pending value is displayed only if we don't have any information yet.
```
ohadlevy: then what is the difference to the above statement about pause? as you just said its mostly paused due to an error, which other state exists and why?

ares: I think the task might be paused even if no error occurred so you would see paused task with success state. @iNecas or @adamruzicka do you have any example? I think "Listen on candlepin events" was such task but I think the state was renamed there to be less confusing? The generic use case is that the task triggers some external event and pauses itself. Pause actually means it's waiting to be resumed by some external system, e.g. after the external event finished.

Anyway this Q is about specific combination of paused task with error result. As it says, in task detail, we denote this task result as pending since the task hasn't finished yet, but in task list we display the "current status". I'd prefer to always display pending until the task is stopped but it was not accepted. The reason was that this information might be useful even before the task finished.

ares:
@ohadlevy I tried to explain what I meant and updated the rest, could you please take another look if it makes more sense now?

ares:
@ohadlevy @adamruzicka ping

plugins/foreman_tasks/0.8/index.md
```diff
@@ -0,0 +1,302 @@
...
+
+    foreman-rake foreman_tasks:export_tasks TASK_SEARCH='label = "Actions::Foreman::Host::ImportFacts"' \
+      TASK_FILE=/tmp/last_week_importing.tar.gz TASK_DAYS=7
+
+If you prefer CSV format instead of HTML pages, you use use **TASK_FORMAT** variable and change to format to csv. Only html and csv formats are available at the moment. The CSV format does not generate the gzipped tarball but plain CSV file that contains limited information of matched tasks. The information includes task id, state, type, label, result, parent_task_id, started_at, ended_at attributes.
+
+
+# 4. Developer guide
+
+We don't have the developer guide yet. At the moment we recommend to read the code in the [repository](https://github.com/theforeman/foreman-taskshttps://github.com/theforeman/foreman-tasks). Please consider contributing.
```
dLobatog: The link is wrongly formatted here. I would rather recommend using the ActiveJob Dynflow interface instead, which is fully documented by Rails and has examples all over. There's nothing special we do, except that I tweaked it to accept `def humanized_*` methods to play well with tasks.

dLobatog: The link is wrongly formatted here. I would rather recommend using the ActiveJob Dynflow interface instead, which is fully documented by Rails and has examples all over. There's nothing special we do, except that I tweaked it to accept `def humanized_*` methods to play well with tasks.

plugins/foreman_tasks/0.8/index.md
adamruzicka: The "Listen on candlepin events" task should always be in the `running` state.

It sounds like you're confusing tasks' `paused` state with Dynflow steps' `suspended`. A step can be `suspended` which means it is waiting for an external event. When a task/execution plan is `paused` it means it encountered an error which it can't solve on its own and needs the user to either cancel it or fix the issue and resume it.

Also `s/stoped/stopped/`

ares: Ok, thanks for clarification. I updated Q&A section to mention only error as a cause for task to be paused

plugins/foreman_tasks/0.8/index.md
ares: Thanks, fixed. I think ActiveJob way should be mentioned here but I think there are still cases where you want to use advanced features of foreman-tasks, such as locking, that requires you to write the task manually. So I think the developer guide still makes sense covering stuff like locking, humanized_output etc. Anyway, I'd appreciate if you put together examples with ActiveJob API to this manual once it's ready.

ares:
I updated the PR, please take another look. I think the manual can be continuously improved and this could hopefully be already a baseline for anyone who'd like to contribute with even single sentence.

plugins/foreman_tasks/0.8/index.md
```diff
@@ -0,0 +1,302 @@
...
+
+- Q: There is a task with state *paused* and result *error*, what does that mean?
+- A: This indicates that the task was paused because of some error occurred. In task detail you should see question mark at *Result* field but in table, we display the value matching to what the result of the task would be, if we stopped it right now. Thanks to this you can also easily search tasks that needs attention by filtering on result value. The same applies for warning result. The pending value is displayed only if we don't have any information yet.  
+
+### 3.2.2 Other task data
+
+There are more tabs in task detail page. They are mostly used during debugging.
+ 
+- **Running steps** - you can see steps actively being performed for a running task, after the task is stopped this tab is empty
```
adamruzicka: missing period at EOL

plugins/foreman_tasks/0.8/index.md
```diff
@@ -0,0 +1,302 @@
...
+- **Locks** - tasks can lock resources preventing other tasks to access them. There are two types of locks, *exclusive* and *non-exclusive* sometimes called *link*. If the author of the task locks some resource e.g. bank account, other task trying to lock the same bank account will fail with error that the resource is locked. If the tasks supports resuming, you can wait for the first task to finish and them resume the failed one. If there's no other conflicting task, it should finish, moving it's state to stopped and setting state to warning. The warning is there since there was some error during the task execution. Non-exclusive locks are only links to resources. They do not block any access and serve only for referring purpose.
+- **Raw**- This tab is mostly for developers who debug errors. You can find internal information such as uniq ID of the task, task type denoted as *Label*, duration of the execution, input and output in machine readable form. If the task mirrors external task in some external service, you can find id of this external task here as well.
```
adamruzicka: `s/uniq/unique/`

plugins/foreman_tasks/0.8/index.md
```diff
@@ -0,0 +1,302 @@
...
+- **VERBOSE** - set to "true" for more verbose output
+- **BATCH_SIZE** - the size of batches the tasks get processed in (1000 by default)
+
+To see the current configuration (what actions get cleaned automatically and what is their after period), this script can be used:
+
+    foreman-rake foreman_tasks:cleanup:config
+
+## 3.8 Exporting
+
+If you want to share the the history of tasks with someone else who does not have access to your Foreman instance, there's a tool to do the export. This is especially useful when developers asks you for debugging some debugging data. To do the export, use to following rake task provided by the plugin
```
adamruzicka: Too much `debugging` on this line I'd say

plugins/foreman_tasks/0.8/index.md
```diff
@@ -0,0 +1,302 @@
...
+
+    foreman-rake foreman_tasks:export_tasks
+
+This will create gzipped tar file in temp directory containing html page for each exported task. The path to the file is printed after the export is finished. These data contains all tasks there are present in database. To limit what exact tasks should be part of the export, use following variables when running the task.
```
adamruzicka: > These data contains all tasks there are present in database.

This isn't exactly true. Running just `foreman-rake foreman_tasks:export_tasks` will export all the tasks started in the last 7 days and all unsuccessful ones started in last 60 days (see [here](https://github.com/theforeman/foreman-tasks/blob/master/lib/foreman_tasks/tasks/export_tasks.rake#L231))

plugins/foreman_tasks/0.8/index.md
```diff
@@ -0,0 +1,302 @@
...
+
+- **TASK_SEARCH** - scoped search filter (example: 'label = "Actions::Foreman::Host::ImportFacts"')
+- **TASK_DAYS** - number of days to go back
+
+To explicitly name the export, you can use **TASK_FILE** variable. The following example would export all facts importing tasks from last week into file /tmp/last_week_importing.tar.gz
+
+    foreman-rake foreman_tasks:export_tasks TASK_SEARCH='label = "Actions::Foreman::Host::ImportFacts"' \
+      TASK_FILE=/tmp/last_week_importing.tar.gz TASK_DAYS=7
+
+If you prefer CSV format instead of HTML pages, you use use **TASK_FORMAT** variable and change to format to csv. Only html and csv formats are available at the moment. The CSV format does not generate the gzipped tarball but plain CSV file that contains limited information of matched tasks. The information includes task id, state, type, label, result, parent_task_id, started_at, ended_at attributes.
```
adamruzicka: `s/, ended_at/ and ended_at/` would sound more natural

adamruzicka:
Did another read through, left some inline comments

ares:
Thanks @adamruzicka, please take another look, all comments should be fixed. Also after this is merged, we should probably copy the manual to 0.9 and do following updates there.

adamruzicka:
Looks good to me

adamruzicka:
Could anyone please merge this? I'd do it myself but I don't have write access to this repository.

iNecas:
I don't have either, perhaps @dLobatog could help?

mmoll:
merged, díky všetkým!

lzap:
Excellent work! Kudos.

PR #812, opened 2017-02-07
Add a backup advise to quickstart install

aptituz:
This commit adds a hint to install instructions that
advises a systems configuration backup for installations on existing
systems.

I feel that advisories is necessary because of puppetlabs/apache module usage :). Not entirely happing with the wording, but I'm unsure what assumptions the installer revolves around. Is is generally based on the assumption that it's running on a fresh system? If so, the hint should probably read way different.

### DIFF
--- a/_includes/manuals/1.14/2.1_quickstart_installation.md
+++ b/_includes/manuals/1.14/2.1_quickstart_installation.md
@@ -1,4 +1,4 @@
-[The Foreman installer](/manuals/{{page.version}}/index.html#3.2ForemanInstaller) uses Puppet to install Foreman. This guide assumes that you have a newly installed operating system, on which the installer will setup Foreman, a Puppet master, and the [Smart Proxy](/manuals/{{page.version}}/index.html#4.3SmartProxies) by default.
+[The Foreman installer](/manuals/{{page.version}}/index.html#3.2ForemanInstaller) uses Puppet to install Foreman. This guide assumes that you have a newly installed operating system, on which the installer will setup Foreman, a Puppet master, and the [Smart Proxy](/manuals/{{page.version}}/index.html#4.3SmartProxies) by default. It's **not advisable** to follow the steps below on an existing system, since installer will affect configuration of several components.
 
 #### Select operating system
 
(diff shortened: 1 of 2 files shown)


_includes/manuals/1.14/2.1_quickstart_installation.md
```diff
@@ -200,10 +200,15 @@ apt-get update && apt-get -y install foreman-installer
...
 
+If you are planning an installation on an existing system, it's strongly advised to make
+ a backup of your system configuration. This especially applies to apache configuration,
+ since the used puppet module does remove existing configuration files.
+
 <div class="quickstart_os quickstart_os_ubuntu1404 quickstart_os_ubuntu1604 alert alert-info">
   Ensure that <code>ping $(hostname -f)</code> shows the real IP address, not 127.0.1.1.  Change or remove this entry from /etc/hosts if present.
 </div>
 
+
```
mmoll: superfluous newline

_includes/manuals/1.14/2.1_quickstart_installation.md
```diff
@@ -200,10 +200,15 @@ apt-get update && apt-get -y install foreman-installer
 
 #### Running the installer
 
+If you are planning an installation on an existing system, it's strongly advised to make
+ a backup of your system configuration. This especially applies to apache configuration,
+ since the used puppet module does remove existing configuration files.
```
mmoll: I'd rather add some outline, what configurations are overwritten and advise to use the installer only on a clean, single-purpose box.

mmoll (changes requested):

_includes/manuals/1.14/2.1_quickstart_installation.md
aptituz: This implies not installing Foreman on an existing system, since the installier is the recommended installations. Thats why I hesistated from writing it in First instance.
With regards to the Outline: I thought about but Apache is the only case I'm currently aware. Any other?

mmoll: depending on the type of install: at least also postgres/mysql config, puppet config, tftps/xinetd configs, BIND config, DHCPd config.

mmoll: aw, quickstart with defaults, so BIND and DHCPd is not touched

aptituz: Ok, thanks. I'll update the wording as soon as I find time for it.

aptituz:
@mmoll

I changed it slightly.

First, I added a terse hint to quickstart guide. Second I added a paragraph to the installation instructions itself.

The hint in the quickstart guide is terse and explicit, since the quickstart guide should probably stay short and because an implicit hint (like "use it on a clean system") might not have the wanted signaling effect.

In the installation instructions the hint is placed directly below the list of affected components. And since it's already there I refrained from documenting it in another place.

Please review again and tell me if if it's alright this way. :)

_includes/manuals/1.14/2.1_quickstart_installation.md
```diff
@@ -204,6 +204,7 @@ apt-get update && apt-get -y install foreman-installer
   Ensure that <code>ping $(hostname -f)</code> shows the real IP address, not 127.0.1.1.  Change or remove this entry from /etc/hosts if present.
 </div>
 
+
```
mmoll, line 207: superfluous newline

mmoll:
I think this is OK.

mmoll (approved):

domcleal:
Merged, thanks @aptituz.

PR #813, opened 2017-02-10
Fix #18394. There is no monthly sync plan (only hourly/daily/weekly).

galkindmitrii:
http://projects.theforeman.org/issues/18394


dLobatog:
Thanks @galkindmitrii

PR #814, opened 2017-02-13
Update puppet master API info

ares:
My observations from what I read in code.

### DIFF
--- a/_includes/manuals/1.14/4.3.6_smartproxy_puppet.md
+++ b/_includes/manuals/1.14/4.3.6_smartproxy_puppet.md
@@ -21,7 +21,7 @@ Replace `4.5.0` with the version of Puppet installed on the Puppet master, this 
 
 The proxy generates a list of all Puppet classes and their parameters of the Puppet modules inside the environments declared in puppet.conf. These are imported by Foreman to generate the list of classes, smart class parameters and environments that they belong to. The mechanism used depends on the version of Puppet (specified by `:puppet_version`):
 
-* On Puppet 4, the smart proxy will use the environments and resource_type Puppet Server APIs to list classes and parameters.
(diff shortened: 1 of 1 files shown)


_includes/manuals/1.14/4.3.6_smartproxy_puppet.md
```diff
@@ -21,7 +21,7 @@ Replace `4.5.0` with the version of Puppet installed on the Puppet master, this
 
 The proxy generates a list of all Puppet classes and their parameters of the Puppet modules inside the environments declared in puppet.conf. These are imported by Foreman to generate the list of classes, smart class parameters and environments that they belong to. The mechanism used depends on the version of Puppet (specified by `:puppet_version`):
 
-* On Puppet 4, the smart proxy will use the environments and resource_type Puppet Server APIs to list classes and parameters.
+* On Puppet 4, the smart proxy will use the environments and environment_classes Puppet Server APIs to list classes and parameters. If you're running Puppet Master or Puppet Server older than 4.4, the API is not available. To use resource_types API, set `:puppet_version` to 4.3 even if you have newer Puppet Master.
```
domcleal: 1. "the environments and environment_classes Puppet Server APIs" should state that it uses environment_classes _or_ resource_types, not only environment_classes.
1. Simply state that the environment_classes API is available and used on Puppet 4.4 or higher, and resource_types on 4.0 to 4.3.
1. I don't think the recommendation to set a different Puppet version to the running version belongs in the manual.

domcleal (changes requested):

_includes/manuals/1.14/4.3.6_smartproxy_puppet.md
ares: environment_classes is only available on puppet server, while resource_types is also on puppet master (afaiu the difference) which is why I tried to put the info there, anyway I simplified it as suggested, please take another look

domcleal:
Merged, thanks @ares.

PR #816, opened 2017-02-17
Remove note about therubyracer, no longer used (#16198)

domcleal:


mmoll:
merged, thanks @domcleal!

PR #817, opened 2017-02-17
Remove mention of master branch, only use develop

domcleal:


mmoll:
merged, thanks @domcleal!

PR #825, opened 2017-03-06
Users should install foreman-proxy-content

ehelms:


mmoll:
merged, thanks @ehelms!

PR #828, opened 2017-03-09
Fixes #18840 - fix deploy option on smart proxy content install

jlsherrill:


jlsherrill:
@ehelms if you could double check that this is correct

ehelms:
Thanks @jlsherrill

PR #830, opened 2017-03-14
Add doc for #17228 - global default templates

xprazak2:


ares (changes requested):
Thanks a lot for the documentation!

> You can build PXE default on TFTP proxy from Foreman UI (from 'Provisioning Templates' page)

I'd say parenthesis are not required. Maybe mentioning the button label would help. Also I think the term 'PXE default' would be worth of explaining, e.g. it's a default PXE configuration for hosts that were not recognized by their MAC so it would apply to hosts unknown to Foreman.

> When settings are empty, Foreman uses default values

Could you please list them here?

> Any template that is selected as a value for any of these settings will not be modified during Foreman update/upgrade.

Isn't that true only if template which is not shipped by Foreman is set?

xprazak2:
I made the modifications as suggested.

>>Any template that is selected as a value for any of these settings will not be modified during Foreman update/upgrade.

>Isn't that true only if template which is not shipped by Foreman is set?

It is true for all templates, it was the main motivation behind the change. If it is a template shipped by Foreman, then it is [skipped in seeds](https://github.com/theforeman/foreman/blob/develop/db/seeds.d/07-provisioning_templates.rb#L98). If it is user-defined, then we are not touching it anyway.

ares:
So if user choose seeded template we won't be able to change it during update? Shouldn't it be the other way around? We always update it and if user wants to use customized template, he/she clones it and set the clone as new default. The customized is the one that should not be updated.

xprazak2:
> So if user choose seeded template we won't be able to change it during update?

Yes.

> We always update it and if user wants to use customized template, he/she clones it and set the clone as new default. 

That would be the ideal scenario.

> The customized is the one that should not be updated.

If you make a clone and customize it, it will never be changed during updates, because it is a user-defined template, so you do not need to check. Or am I wrong?

The current version ensures that content of global default templates is always what user expects them to be. If we want to update seeded templates each time regardless of whether they are global default or not, we need to remove [a line from seeds](https://github.com/theforeman/foreman/blob/develop/db/seeds.d/07-provisioning_templates.rb#L98) and a corresponding [method](https://github.com/theforeman/foreman/blob/develop/app/models/provisioning_template.rb#L200)

ares:
The original motivation was that we can lock and update all seeded templates on Foreman update. I think we need to remove those lines.

ares:
it's being removed by https://github.com/theforeman/foreman/pull/4320/files#diff-151633105c29141fe3aac7603de185a5L98 anyway

ares:
@xprazak2 the templates locking PR was merged, could you please update the text? We lock even global templates now, the only criteria is whether they are shipped with Foreman or not.

xprazak2:
Updated, I removed text related to locking.

ares (approved):

ares:
:+1: grammar check done by native speaker would be appreciated before merging.

mmoll:
@GregSutcliffe could you have a look?

mmoll:
merged, díky @xprazak2!

Issue #467, opened 2015-11-02
Clean up html leftovers in old blogs


GregSutcliffe:
Some of the pandoc-converted blogs are a mess, we should clean them up a bit

rabajaj0509:
I wish to work on this bug :)

mmoll:
@rahulbajaj0509 just open a a pull request with the needed chnages 👍

PR #831, opened 2017-03-19
Clean up html leftovers in old blogs

rabajaj0509:
Fixes: https://github.com/theforeman/theforeman.org/issues/467


mmoll:
Merged, thanks @rahulbajaj0509!
Please note that your commit had "you@example.com" set as e-mail adress, I did override this.

PR #832, opened 2017-03-20
Fix the return button functionality in blog posts.

rabajaj0509:
At present, consider that i am on the last page and i am reading the first blog on that page. After reading the blog i wish to read the next blog on the same page, if i press "RETURN" button it will take me to the first page and i shall have to traverse again till the last page to read the blog.

This patch takes care of the above problem.


dLobatog:
Thanks @rahulbajaj0509

PR #834, opened 2017-03-25
use foreman-ruby on Debian for plugins

mmoll:


domcleal:
Merged, thanks @mmoll.

PR #838, opened 2017-03-28
Fix Katello Doc Menu to include 3.3

Klaas-:
I hope this is the menu for the quick jump between versions inside the katello docs :)


ehelms:
Thanks @Klaas- !

PR #829, opened 2017-03-14
Add openscap doc for report downloads and running scans from UI

xprazak2:

PR #839, opened 2017-03-29
Fix rex version in openscap docs

adamruzicka:
In https://github.com/theforeman/theforeman.org/pull/829 I mentioned the required fetaures mentioned in the docs will be released in rex 1.2.3, but we skipped that version and released 1.3.0 instead. Sorry for giving misleading information


ehelms:
Thanks @adamruzicka !

PR #842, opened 2017-03-30
Clone openscap for 0.7.0

ares:
The 0.7.0 did not introduce anything new but fix required for 1.15 and extending core roles. So to 0.7 is just copy of 0.6 (no need to review the text).


ehelms:
Thanks @ares

PR #843, opened 2017-03-31
blog: March newsletter

GregSutcliffe:
Just in time again! Think I got all the plugins and mailing list goodies, anything else get missed? The survey blog is linked but the URL should be correct once it's published.


mmoll:
merged, thanks @GregSutcliffe!

PR #845, opened 2017-03-31
Adding katello request annotations

jlsherrill:


jlsherrill:
This is the start of an effort to add annotated documentation between katello, pulp, and candlepin.  This is being generated from https://github.com/Katello/katello/pull/6716 and will likely expand over the next weeks.  Right now only two annotated workflows are included: Repository create and Repository sync. As more are added, each workflow will likely be split into different files and a menu will be added, but this is a start.

bbuckingham:
@jlsherrill, looks good.  ACK!

One enhancement that I think may be nice in the doc (given it's length), is to provide a summary listing of all of the requests at the top that provide links to the section of the document that contains the request details.

jlsherrill:
yes, for the next iteration i'm hoping to actually put each scenario on a different page

PR #851, opened 2017-04-07
Fix broken link in katello docs

Klaas-:
The link on how to create a manifest zip is broken, this should fix it


dLobatog:
Thanks @Klaas- !

PR #852, opened 2017-04-10
Add foreman_templates manual

ares:


plugins/foreman_templates/4.0/index.md
```diff
@@ -0,0 +1,185 @@
+---
+layout: plugin
+title: Foreman Templates 4.0 Manual
+version: 4.0
+---
+
+# 1. {{ page.title }}
+
+This plugin enables synchronization of provisioning templates, partition tables, job templates from external git repository and/or file system. It can be used both with the [community templates](https://github.com/theforeman/community-templates) repository or with any other source of templates that has the same directory structure. Please note that Foreman core provides tracking of changes, browsing the history and reverting natively, but if you need more advanced tools such as branching, code reviews etc. you might be interested in this plugin.
```
GregSutcliffe: "*and* job templates" will finish the list correctly, otherwise it's dangling

plugins/foreman_templates/4.0/index.md
```diff
@@ -0,0 +1,185 @@
...
+
+You can use the native git support for simple syncrhonization with remote git repositories, which is usually useful for syncing new template versions from community repository or for exporting local changes to contribute back. Or if you have more complicated workflow, you can point the plugin to work with local directory. This way, you can use any VCS (including git) to control the versioning of that directory and use this plugin to sync whenever the directory is at desired version.
```
GregSutcliffe: I think "use this plugin to sync the directory when required." might read better here, but it's minor.

plugins/foreman_templates/4.0/index.md
```diff
@@ -0,0 +1,185 @@
...
+
+# 2. Installation
+
+You can install the plugin using the foreman-installer (recommended way). To do that, run the installer with following argument
+
+    foreman-installer --enable-foreman-plugin-templates
+
+If installed foreman from git, you can simply add foreman-templates gem as one of the dependencies. Don't forget to run migrations and seed afterwards.
+
```
GregSutcliffe: No mention of install from packages here? IMO it's preferable to using the installer, but I know I'm in a minority :)

plugins/foreman_templates/4.0/index.md
```diff
@@ -0,0 +1,185 @@
...
+    <th>Meaning on exporting</th>
+  </tr>
+  <tr>
+    <td><i>associate</i></td>
+    <td>associate OS, Organization, Location based on metadata</td>
+    <td>N/A</td>
+  </tr>
+  <tr>
+    <td><i>branch</i></td>
+    <td colspan="2">git branch to read or write from/to (only for git url), default value is based on current Foreman version - e.g. develop or 1.15-stable</td>
```
GregSutcliffe: I'd say "only for git-based repositories" in the parentheses, and maybe put the "repo" option immediately before this one so it's easily referenced.

plugins/foreman_templates/4.0/index.md
```diff
@@ -0,0 +1,185 @@
...
+
+    foreman-installer --enable-foreman-plugin-templates
+
+If installed foreman from git, you can simply add foreman-templates gem as one of the dependencies. Don't forget to run migrations and seed afterwards.
+      
+# 3. Configuration
+
+After the installation there are new settings available. These are the default setting for both importing and exporting tasks. They can be overridden on each synchronization run and serve only as default so users don't have to specify the same value on every run. You will find them under Administer -> Settings -> TemplateSync. Only administrator can modify their values. Following table explains the attribute behavior. Note that some are only used on import or export actions.
+
+<table class="table table-bordered table-condensed table-striped">
```
GregSutcliffe, line 26: I see you're going for alphabetical setting order - but many options are rarely used, does it makes sense to put common options first?

plugins/foreman_templates/4.0/index.md
```diff
@@ -0,0 +1,185 @@
...
+    <td><i>verbose</i></td>
+    <td colspan="2">adds extra verbose messages to the action output</td>
+  </tr>
+</table>
+
+Settings above are preconfigured for importing from upstream community-templates repository.
+
+# 4. Usage (features description)
+
+Both importing and exporting is available as rake tasks. Rake tasks are triggered from terminal. If you installed the plugin using foreman-installer, or simply it's package based installation), you should be able to run rake using `foreman-rake` command. If you installed Foreman alternatively, e.g. cloned from git, you should likely use `RAILS_ENV=production bundle exec rake` ran from your Foreman directory. There are other ways to run the action discussed under API chapter.
```
GregSutcliffe: Package install mentioned here, so it should definitely go under Installation too (see earlier comment)

GregSutcliffe: s/If you installed Foreman alternatively, e.g. cloned from git, you should likely use/Source-based installs of Foreman will need to use Bundler to call Rake as normal, e.g./

GregSutcliffe: I'd put the API sentence on a new line/paragraph to make it stand out for those who want to know.

plugins/foreman_templates/4.0/index.md
```diff
@@ -0,0 +1,185 @@
...
+foreman-rake templates:import
+```
+
+This task based on $repo setting either uses git or scans local directory for templates to import. If you want to customize any of setting for a particular run, just specify it as extra argument.
+
+```sh
+foreman-rake templates:import repo=https://github.com/theforeman/community-templates
+```
+
+The example above would clone the repository to some temp directory. If `branch` was specified, it checkouts this git branch. Then if `dirname` was provided, it will "cd" into it. Finally is searches for all files in this directory and subdirectories that has `'.erb` suffix. Then the template is parsed and imported (see below for more details)
```
GregSutcliffe: s/it checkouts this git branch/it switches to this branch/

GregSutcliffe: s/Finally is/Finally it/

plugins/foreman_templates/4.0/index.md
```diff
@@ -0,0 +1,185 @@
...
+
+```sh
+foreman-rake templates:import repo=/var/lib/my_templates
+```
+
+Compared to previous example, this uses local directory. Therefore not git clone or branch checkout is done. If `dirname` is specified, the task goes into this subdirectory and searches for all files with '.erb' extension. Then the same parsing and importing process follows,
+
+The parsing process reads the content of the file. It expects the content of the template including a metadata. The metadata are an ERB comment on top of the file. It's in yaml format and specifies template name, kind, compatible operating systems, organizations, locations and possibly other. The template gets it's name based on information from metadata. Before the template is saved in database, `prefix` setting is applied to it's name. Then the `filter` regular expression is matched against the prefixed name. If the `filter` matches, templates is saved. If `negate` is set to `true`, the template would be saved only if `filter` regular expression didn't match the template name. If the template with a given name exists in database already, it's content is updated.
```
GregSutcliffe: "including a metadata block." or "including some template metadata" - your choice :)

GregSutcliffe: Is it worth linking to an example from community-templates to show a metadata block?

GregSutcliffe (changes requested):
Looks pretty good to me, thanks @ares! A few small typos and wording changes, that's all.

plugins/foreman_templates/4.0/index.md
sean797: Does it require the same directory structure? I thought it worked with the metadata and all files needed to be suffixed with `.erb`

plugins/foreman_templates/4.0/index.md
```diff
@@ -0,0 +1,185 @@
...
+ 
+# 4.1 Importing
+
+If you have all attributes configured via Settings you can simply run following command
+
+```sh
+foreman-rake templates:import
+```
+
+This task based on $repo setting either uses git or scans local directory for templates to import. If you want to customize any of setting for a particular run, just specify it as extra argument.
```
sean797: s/If you want to customize any of setting/If you want to customize any setting/

I'm not sure `of` makes sense here.

plugins/foreman_templates/4.0/index.md
```diff
@@ -0,0 +1,185 @@
...
+
+The parsing process reads the content of the file. It expects the content of the template including a metadata. The metadata are an ERB comment on top of the file. It's in yaml format and specifies template name, kind, compatible operating systems, organizations, locations and possibly other. The template gets it's name based on information from metadata. Before the template is saved in database, `prefix` setting is applied to it's name. Then the `filter` regular expression is matched against the prefixed name. If the `filter` matches, templates is saved. If `negate` is set to `true`, the template would be saved only if `filter` regular expression didn't match the template name. If the template with a given name exists in database already, it's content is updated.
+
+The result of the whole import is printed on the STDOUT. The output contains list of created/updated templates as well as applied diffs.
+
+Few more useful examples follow
+
+```sh
+# prefix all templates with "[comunity] " string to avoid overriding existing templates
+foreman-rake templates:import prefix='[community] '
```
sean797: Not required but it might be nice to show something like:

```sh
$ tree /tmp/my-repo/
/tmp/my-repo/
└── large-ptable.erb

0 directories, 1 file
$ foreman-rake templates:import repo=/tmp/my-repo/ prefix='[community] '
```
Would result in a template named `[community] large-ptable`.

plugins/foreman_templates/4.0/index.md
```diff
@@ -0,0 +1,185 @@
...
+
+# 4.2 Exporting
+
+If you want to edit templates directly in Foreman and then sync them back to external repositories, you need the exporting task. This plugin allows to export directly to git repository or local directory. For exporting to git repository, the user that's Foreman running under needs to have write (commit) access to this repository. If your settings is correctly configure, exporting can be triggered by running following command.
+
+```sh
+foreman-rake templates:export
+```
+
+If the `repo` is git repository, we clone it to some temp directory. Then we apply changes to template files and if there was a change, we automatically create a commit with generate message and push it back to origin. We do not delete files that are not present in Foreman database, so this should not be destructive. On the other hand if you want more control over the process or you want to use another SCM, local directory export is advised. To specify local directory, just set the repo accordingly
```
sean797: s/Then we apply changes to template files and if there was a change, we automatically create a commit with generate message and push it back to origin./Then we apply changes to template files and automatically create a commit with a generate message and push it back to origin./

Also

s/clone it to some temp directory/clone it to a temp directory/

Also

s/To specify local directory/To specify a local directory/

plugins/foreman_templates/4.0/index.md
ares: This is also the order we have in settings table. It's not a huge number so I'd keep it ordered alphabetically

sean797 (changes requested):

ares:
Thanks @GregSutcliffe, I think I addressed all your comments. Please take another look.

@sean797 same for you. Just one answer instead of a text change

> Does it require the same directory structure? I thought it worked with the metadata and all files needed to be suffixed with .erb

It does not required the same structure and it works as you said for import. But export creates the same structure the community-templates has, so if you want to use both direction sync, you should use this structure.

plugins/foreman_templates/4.0/index.md
ares: added as a new example with a comment

sean797 (approved):
LGTM @ares ! Thanks for doing this 👍

plugins/foreman_templates/4.0/index.md
GregSutcliffe: We'll have to agree to disagree on this, as I feel for docs the user should get the most relevant info first - but I don't feel strongly enough to insist on it :)

plugins/foreman_templates/4.0/index.md
```diff
@@ -0,0 +1,198 @@
...
+
+You can use the native git support for simple synchronization with remote git repositories, which is usually useful for syncing new template versions from community repository or for exporting local changes to contribute back. Or if you have more complicated workflow, you can use this plugin to sync the directory when required. This way, you can use any VCS (including git) to control the versioning of that directory and use this plugin to sync whenever the directory is at desired version.
+
+# 2. Installation
+
+You can install the plugin using the foreman-installer (recommended way). To do that, run the installer with following argument
+
+    foreman-installer --enable-foreman-plugin-templates
+
+If you installed the plugin using package directly, you should run the migrations and seed (just to be sure) and restart the Foreman.
```
GregSutcliffe: Can we either detail these steps or link to where they can be found? If this is the first plugin a user ever installs, they may not know how...

plugins/foreman_templates/4.0/index.md
```diff
@@ -0,0 +1,198 @@
...
+
+The parsing process reads the content of the file. It expects the content of the template including some template metadata. The metadata are an ERB comment on top of the file, for illustration see some template in [community-templates repository](https://github.com/theforeman/community-templates/blob/develop/provisioning_templates/provision/kickstart_default.erb). It's in yaml format and specifies template name, kind, compatible operating systems, organizations, locations and possibly other. The template gets it's name based on information from metadata. Before the template is saved in database, `prefix` setting is applied to it's name. Then the `filter` regular expression is matched against the prefixed name. If the `filter` matches, templates is saved. If `negate` is set to `true`, the template would be saved only if `filter` regular expression didn't match the template name. If the template with a given name exists in database already, it's content is updated.
+
+The result of the whole import is printed on the STDOUT. The output contains list of created/updated templates as well as applied diffs.
+
+Few more useful examples follow
```
GregSutcliffe: nitpick: *A* few more ...

plugins/foreman_templates/4.0/index.md
```diff
@@ -0,0 +1,198 @@
...
+
+
+## 7.2 Contributing
+
+Follow the [same process as Foreman](/contribute.html#SubmitPatches)
+for contributing.
+
+# 8. Links
+
+* foreman_templates plugin [https://github.com/theforeman/foreman_chef](https://github.com/theforeman/foreman_templates)
```
GregSutcliffe: Url is right, but text says Chef :P

GregSutcliffe:
@ares Almost there, a couple of tiny things :)

plugins/foreman_templates/4.0/index.md
ares: I would prefer not documenting it in details. foreman-installer is the recommended way. We also don't document you could do `git clone ...`

plugins/foreman_templates/4.0/index.md
ares: damn, I swear I grepped it :-)

ares:
@GregSutcliffe thanks, another round? :-)

plugins/foreman_templates/4.0/index.md
GregSutcliffe: I don't think we need details (they go out of date) but maybe a link to https://theforeman.org/plugins/#2.2Packageinstallation ?

GregSutcliffe (approved):

GregSutcliffe:
Merged, thanks @ares!

PR #855, opened 2017-04-12
Installer and Proxy options for 1.15

dLobatog:


domcleal:
This seems to be missing updated installer parameters and changes to the realm smart proxy configuration options.

PR #857, opened 2017-04-18
API documentation for 1.15

dLobatog:
The diff is way too long to examine, if you want to check out whether it's ok, clone the branch and test it locally. This is the output of the procedure to get the API docs on the README of this repo, nothing else added/removed.


domcleal:
Merged, thanks @dLobatog.

PR #858, opened 2017-04-19
update docs to use forklift instead of katello-deploy

evgeni:


dLobatog:
We should probably start thinking about how to phase out manuals for some versions - updating old ones is tedious. Thanks @evgeni

PR #859, opened 2017-04-19
use HTTPS for all GitHub links

evgeni:
also drop the comment one can use http to trick firewalls, GitHub does not offer plain http anymore (just redirects)


mmoll (approved):

mmoll:
merged, спасибо Женя!

PR #860, opened 2017-04-24
fix display name of 3.4-RC and set the version to 3.4 instead of nightly

evgeni:


jlsherrill:
ACK thanks @evgeni !!!

PR #862, opened 2017-04-24
Refs #18639 - default templates locking

ares:


ares:
Ideally this should get in before 1.15 release

ares:
1.15 is getting closer, anyone could review?

mmoll:
I think this is fine, merged, díky @ares!

PR #863, opened 2017-04-27
Added incremental backup example

lzap:
SSIA

PR #866, opened 2017-05-04
Clarifying incremental backup strategy and args

phess:
Adding clarifications around incremental backup strategies + fixing mentions to previous backup directories when using --incremental.


lzap:
Yeah indeed confusing part. I also created an example in here: https://github.com/theforeman/theforeman.org/pull/863/files

lzap:
Thanks that's a merge.

PR #867, opened 2017-05-08
Change 1.15 to be a stable release

dLobatog:
Only to be merged when 1.15 is out


dLobatog:
@domcleal I think this can be merged now, I'll send the announcement in a few moments

domcleal (changes requested):
`_includes/version.html` and `_includes/latest_news.html` need to be updated too.

GregSutcliffe:
Obsoleted by #871

PR #870, opened 2017-05-12
Add installing yum-utils to katello installation

johnpmitsch:
We specify using yum-config-manager for RHEL installations,
but that won't be installed by default, so we should ensure
the yum-utils package is on the user's system before
asking them to use yum-config-manager.


jlsherrill:
ACK thanks @johnpmitsch !

PR #889, opened 2017-06-02
Add foreman_virt_who_configure plugin

ares:


plugins/foreman_virt_who_configure/0.1/index.md
```diff
@@ -0,0 +1,173 @@
...
+
+    yum install rubygem-foreman_virt_who_configure
+    foreman-rake db:migrate
+    foreman-rake db:seed
+
+Plugin package lives in [Katello repositories](https://fedorapeople.org/groups/katello/releases/yum/).
+
+If installed foreman from git, you can simply add foreman-virt_who_configure gem as one of the dependencies. Don't forget to run migrations and seed afterwards and obviously restart the Foreman. You still need to have Katello installed.
+
+Note for Satellite 6 users, the plugin is installed out of the box and no extra action is required.
```
lzap, line 23: So does it install for S6 but not for Katello? I'd rather see this to be very same, but not the right place here. Just making comment, won't block this :-)

plugins/foreman_virt_who_configure/0.1/index.md
```diff
@@ -0,0 +1,173 @@
...
+
+Note for Satellite 6 users, the plugin is installed out of the box and no extra action is required.  
+
+# 3. Configuration
+
+There's nothing to configure for this plugin yet. See the next chapter.
+
+# 4. Usage (features description)
+
+The plugin adds a new link "Virt-who configurations" under Infrastructure menu. This page lists all virt-who configurations that were created in Foreman. To create a new configuration, click Create Config button. Once provide all information, the configuration is registered and you can deploy it. Each configuration provides a configuration script that you find after clicking on its name. The configuration script should be run on host where you want to deploy virt-who instance at, e.g. the same host where Foreman runs. Existing virt-who configurations can be deleted when no longer required. Plugin also provides monitoring widget for dashboard. Please see chapters below for more details on every feature.
```
lzap: Once *you* provide all information ...

The configuration script should be run on *a* host where ...

plugins/foreman_virt_who_configure/0.1/index.md
```diff
@@ -0,0 +1,173 @@
...
+ 
+## 4.1 Creating
+
+In order to create new virt-who configuration, navigate to Insrastructure -> Virt-who configurations and click Create Config button. This takes you to form which guides you through what we need to know in order to successfully configure virt-who. Most of fields provide some inline help, please click the help icon for more information.
+
+Note that every virt-who configuration must exist in an organization. If you haven't selected any organization, you'll see Owner field in the form where you need to choose an Organization. If you already are in some organization context, the Owner field is not displayed and the virt-who configuration will be automatically created in this Organization scope.
+  
+Default values which are prefilled are recommended values which are usually not needed to be changed. This applies to especially to Interval, Hypervisor ID and Satellite server FQDN. The Satellite server FQDN is where the virt-who instance will send reports to so make sure this FQDN is reachable from the host, where the virt-who instance will run. The default value is determined based on what Setting foreman_url is configured to.
```
lzap, line 39: s/Satellite/Foreman/g

plugins/foreman_virt_who_configure/0.1/index.md
```diff
@@ -0,0 +1,173 @@
...
+
+Once you submit all information, the configuration script is rendered.
```
lzap: Rendered where? I'd like to know more info how this works. Does it create configuration files under `/etc/virt-who`? Can you explain somewhere?

plugins/foreman_virt_who_configure/0.1/index.md
```diff
@@ -0,0 +1,173 @@
...
+  
+Default values which are prefilled are recommended values which are usually not needed to be changed. This applies to especially to Interval, Hypervisor ID and Satellite server FQDN. The Satellite server FQDN is where the virt-who instance will send reports to so make sure this FQDN is reachable from the host, where the virt-who instance will run. The default value is determined based on what Setting foreman_url is configured to.
+
+Once you submit all information, the configuration script is rendered. 
+
+## 4.2 Deploying
+
+To find the configuration script for a virt-who configuration, click it's name on virt-who configurations list page. You will see the page with brief information about how to deploy the configuration script and then the script itself.
+ 
+There are multiple ways to do the deployment. Only recent version of virt-who is supported. The scripts does the check and prints warning if unsupported version is detected. Note for Satellite 6 users, the supported version is always in Satellite Tools repository, make sure the host where you run the script has access to it or you have supported virt-who version available.
```
lzap: Note for *RHEL* users...

plugins/foreman_virt_who_configure/0.1/index.md
```diff
@@ -0,0 +1,173 @@
...
+
+### 4.2.1 Copy and paste
+
+You can simply copy and paste the content of the script to some file on host where you want to deploy it. Make sure the file is only readable by root, it contains sensitive information. Also make sure, you clean clipboard afterwards. Then simply run the script e.g. like this
+
+```sh
+bash /root/my_script
+```
```
lzap, line 55: I am still little bit fuzzy about what the script contains, can you show an example in the docs? That would be very helpful. First time reader can't really tell what it actually does.

plugins/foreman_virt_who_configure/0.1/index.md
```diff
@@ -0,0 +1,173 @@
...
+
+### 4.2.2 Downloading the script
+
+You can download the script using "Download the script" button. If you're running this on some workstation, make sure you store it on secured location and transfer it to target host by secure transport layer (e.g. ssh). The script contains sensitive information. After you copy it to secure location on target, run the script. You might need to make it executable first. Following commands illustrate the process.
```
lzap, line 59: If you're running this on some workstation - I think it's irrelevant, you *always* store it on a secure location :-)

plugins/foreman_virt_who_configure/0.1/index.md
```diff
@@ -0,0 +1,173 @@
...
+```sh
+scp /home/user/deploy_virt_who_config_4.sh root@satellite.example.com:/root/
+ssh root@satellite.example.com
+chmod u+x /root/deploy_virt_who_config_4.sh
+/root/deploy_virt_who_config_4.sh
+```
+
+### 4.2.3 Using API
+
+The plugin comes with REST API that can be used. You can use API endpoint for downloading the script directly on the host, where you want to deploy the virt-who. Following curl command would download it. Make sure you keep it in secure place.
```
lzap: *a* REST API...

use *the* API endpoint...

plugins/foreman_virt_who_configure/0.1/index.md
```diff
@@ -0,0 +1,173 @@
...
+
+* **No Report Yet** - this means, no virt-who report was recieved by Foreman yet, either the configuration was not deployed yet or there was some error during the deployment or virt-who can't connect to Foreman
+* **OK** - we received the report within the scheduled interval, everything is OK
+* **No Change** - virt-who does not send any report when it does not find any change on hypervisor, if we don't receive any report during the scheduled interval, we consider the configuration as unchanged. This could also happen if virt-who fails to upload reports for whatever reason, if you know there was new VM added on the hypervisor but you see the configuration in __No Change__ state, check that virt-who is running and its logs.
+ 
+There's also a widget dasboard that you can watch. It contains a table with number of configurations per each state. If you click on status name, you'll see all configurations in this state. The widget also lists last 3 configuration in __No Change__ state as this might or might not be problematic.
```
lzap: also a *dashboard widget*....

last *three configurations*

plugins/foreman_virt_who_configure/0.1/index.md
```diff
@@ -0,0 +1,173 @@
...
+ 
+On configurations lists page you can also search based on the state. The search syntax can be used
```
lzap: On *the configuration* lists page...

plugins/foreman_virt_who_configure/0.1/index.md
```diff
@@ -0,0 +1,173 @@
...
+* **Virt-who Reporter** - this is a role that contains minimal set of permissions for virt-who to upload the report, it can be used if you configure virt-who manually and want to use user that has locked down account
+* **Virt-who Viewer** - this user can see all configurations including their configuration scripts, which means viewers could still deploy the virt-who instances for existing virt-who configurations. 
+
+Every virt-who configuration create a service user which is used for virt-who instance authentication. This means every virt-who configuration uses a different account. The service user login is _virt_who_reporter_$id_ and has randomly generated password. The user account is assigned _Virt-who Reporter_ role. The service user can't login to Foreman and can only be used for virt-who reporting. 
+
+# 5. API
+
+All functionality that is provided by WebUI can be also achieved by REST API. To find information about the available API endpoints, visit `your.foreman.com/apidoc`. In addition to common 7 routes for managing virt-who configurations, we have one endpoint for generating the configuration script.
```
lzap: *seven*

plugins/foreman_virt_who_configure/0.1/index.md
```diff
@@ -0,0 +1,173 @@
...
+
+Every virt-who configuration create a service user which is used for virt-who instance authentication. This means every virt-who configuration uses a different account. The service user login is _virt_who_reporter_$id_ and has randomly generated password. The user account is assigned _Virt-who Reporter_ role. The service user can't login to Foreman and can only be used for virt-who reporting. 
+
+# 5. API
+
+All functionality that is provided by WebUI can be also achieved by REST API. To find information about the available API endpoints, visit `your.foreman.com/apidoc`. In addition to common 7 routes for managing virt-who configurations, we have one endpoint for generating the configuration script. 
+
+# 6. Hammer
+
+The Hammer plugin is available too. To install it, install the Hammer plugin package from katello repostories. Note for Satellite 6 users, you don't need to take any action, as long as you have satellite-cli package installed, the plugin is configured automatically.
```
lzap: *Katello* instead Satellite 6

lzap (changes requested):
Some typos, overall too much *you* is used, we should avoid this in general.

plugins/foreman_virt_who_configure/0.1/index.md
ares: I'll ask Katello guys if they would accept this, but it sounds to me like making have foreman_discovery as a dependency of Foreman...

plugins/foreman_virt_who_configure/0.1/index.md
ares: it's actually Satellite, virt-who uses Satellite 5 and Satellite 6 terms in everywhere including man pages, so I kept this configuration attribute with "Satellite in it's name too", even though it can be Foreman with Katello URL as well.

plugins/foreman_virt_who_configure/0.1/index.md
ares: Changed to "Once you submit all information, the configuration script is rendered and displayed on the screen.", the deployment process is described below

plugins/foreman_virt_who_configure/0.1/index.md
ares: good catch

plugins/foreman_virt_who_configure/0.1/index.md
ares: I expanded the end of 4.1 with "The script in general contains commands to install virt-who, create it's configuration under /etc/virt-who.d/ and starting the service.". Script is too long to put it here as an example. I generated one for you [here](https://pastebin.com/tsuTAZwS)

plugins/foreman_virt_who_configure/0.1/index.md
ares: I tend to save files (logs, images) under /tmp since otherwise I forgot to clean them up form my home :-) I think it does not hurt to mention.

plugins/foreman_virt_who_configure/0.1/index.md
ares: This applies only to Satellite 6, I'd like to avoid confusion and was hoping that we can mention the difference here. I don't think there's similar package upstream.

ares:
Thanks a lot @lzap, I tried to address all your comments, please take another look.

plugins/foreman_virt_who_configure/0.1/index.md
lzap: Hmm I'd rather change it here and there, that's not correct. We can do it later if you prefer.

ares: I tend to agree but it would mean a long period for transition, the string is also used in virt-who arguments IIRC so users would need to migrate their configuration files. We'll see, if it causes to much confusion, we could simplify that to "Server FQDN" or something similar.

PR #891, opened 2017-06-07
Fixes #17776 - add katello apidocs

jlsherrill:


ehelms:
My word, this is big. I think maybe we should trim out the `v1` APIs that are present and consider not including any Foreman APIs? I know we override some but I am not as clear how that comes out in the docs.

jlsherrill:
We can trim out v1 as we don't add to or modify those, however we have to include the foreman apis as add to and modify those.

jlsherrill:
removed v1 apis

ehelms (approved):

PR #894, opened 2017-06-12
Add birthday event @ NETWAYS

dgoetz:
This commit adds our birthday event to the event page. Feel free to close this in favour of a separate birthday event page if there will be one for this year.


GregSutcliffe:
Merged, thanks @dgoetz ! I'll sort out a proper birthday page this week hopefully, but this should be present anyway.

PR #895, opened 2017-06-12
Added warning

charlescva:
Added warning about installing over existing foreman deployments.  As of version 1.15 (Foreman), attempting to install katello will break several components, autoremoving packages that are no longer needed will result in removal of several foreman-installer dependencies as well as break PuppetCA configurations.


charlescva:
This should probably be merged with the other stable releases as well as NIGHTLY.

dLobatog (approved):
Thanks @charlescva , I will put it on 3.3 and 3.4 upon merge.

dLobatog:
Merged as b5ff295ea0a75010f6b7be59693a5f73fc706496

PR #896, opened 2017-06-14
move back last step for signing katello rpms

jlsherrill:


dLobatog (approved):
Looks pretty similar to http://projects.theforeman.org/projects/foreman/wiki/Release_Process, maybe we want to automate & merge efforts at some point, thanks @jlsherrill

PR #897, opened 2017-06-16
add user ssh key documentation

timogoebel:
This adds a small paragraph about the user ssh key feature added in 1.15.

### DIFF
--- a/_includes/manuals/1.15/4.1.6_user_management.md
+++ b/_includes/manuals/1.15/4.1.6_user_management.md
@@ -1,0 +1,7 @@
+
+Foreman is all about hosts and users interacting with these hosts.
+
+#### SSH Keys
+
+Each Foreman user can have multiple SSH keys assigned when editing a user. These keys alone do not serve any purpose, but are available for use in provisioning templates and can be accessed via ENC data. They provide an easy way to manage users and login ssh keys on hosts without the need for LDAP.
+If you want users to be able to login to a host using the data provided in Foreman, you need to include the `create_users` snippet in your provisioning template. There is a [puppet module](https://github.com/ekohl/puppet-foreman_simple_user) available to keep user data in sync with Foreman and your hosts.
--- a/manuals/1.15/index.md
+++ b/manuals/1.15/index.md
@@ -79,6 +79,8 @@ previous_version: "1.14"
 {%include manuals/1.15/4.1.4_auditing.md %}
 ### 4.1.5 Searching
 {%include manuals/1.15/4.1.5_searching.md %}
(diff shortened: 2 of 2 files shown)


timogoebel:
cc: @ares

_includes/manuals/1.15/4.1.6_user_management.md
```diff
@@ -0,0 +1,7 @@
+
+Foreman is all about hosts and users interacting with these hosts.
+
+#### SSH Keys
+
+Each Foreman user can have multiple SSH keys assigned when editing a user. These keys alone do not serve any purpose, but are available for use in provisioning templates and can be accessed via ENC data. They provide an easy way to manage users and login ssh keys on hosts without the need for LDAP.
+If you want users to be able to login to a host using the data provided in Foreman, you need to include the `create_users` snippet in your provisioning template. There is a [puppet module](https://github.com/ekohl/puppet-foreman_simple_user) available to keep user data in sync with Foreman and your hosts.
```
ares, line 7: @ekohl how much is the puppet module supported by you? Would you be up for moving under theforeman org?

ares:
LGTM :+1: , waiting on answer from Ewoud

_includes/manuals/1.15/4.1.6_user_management.md
ekohl: Definitely, after a review from others and modulesync then we can move it to theforeman org. After publishing it to the forge it should also be easier to use.

ekohl (approved):
Looks good. I'm not sure if we should wait to move the puppet module or just publish this now.

_includes/manuals/1.15/4.1.6_user_management.md
timogoebel: As I read the code when you visited Germany, I'd consider the review done. :-)

timogoebel:
> I'm not sure if we should wait to move the puppet module or just publish this now.

Let's publish this now. We can easily improve this or change the link later.

_includes/manuals/1.15/4.1.6_user_management.md
ekohl: @mmoll I'd like to get the module moved and released on the forge but might not have time for it until next week. Think we could work on that then?

mmoll: I can't help with the move to the github org (@domcleal? @ohadlevy?), but with modulesync and forge things.

ares:
Sure, I was curious if that's somewhere on the roadmap :-) Thanks @timogoebel, merging!

PR #898, opened 2017-06-19
Enable binary callback on Ansible < 2.2 too

dLobatog:
As evidenced in
https://github.com/theforeman/foreman_ansible/issues/86#issuecomment-309139251
the docs are confusing and may lead to think bin_ansible_callbacks =
True is not needed on Ansible <= 2.2


agx:
@dLobatog looks good to me!

ares:
Thank you @dLobatog, merging

PR #905, opened 2017-07-07
Fix spelling of "Administer"

arjenz:
Fix spelling of Adminster-> Administer (only in the 1.15 tree).


ares:
Thanks @arjenz, merging!

PR #907, opened 2017-07-10
Add foreman_openscap 0.8

ares:


ares:
copy of 0.7, since 0.8 has been released already

plugins/foreman_openscap/0.8/index.md
```diff
@@ -0,0 +1,430 @@
+---
+layout: plugin
+title: Foreman OpenSCAP manual
+images: /plugins/foreman_openscap/0.6
+version: 0.6.0
```
lzap: Is this the correct version (0.8)?

plugins/foreman_openscap/0.8/index.md
```diff
@@ -0,0 +1,430 @@
...
+    <td>0.5.0</td>
+    <td>0.1.2</td>
+  </tr>
+  <tr>
+    <td>>= 1.13</td>
+    <td>0.6.0</td>
+    <td>0.6.0</td>
+    <td>0.2.0</td>
+  </tr>
+</table>
```
lzap: Missing compatibility info.

lzap (changes requested):
It's a good idea to make a copy in a separate commit to see the changes. Can you point on the paragraphs you have rewritten?

Would be a good idea to put images which are common into a common folder so we don't need to copy them every single release.

plugins/foreman_openscap/0.8/index.md
ares: fixed, also fixed in 0.7 which I took for the copy

plugins/foreman_openscap/0.8/index.md
ares: added

ares:
Thanks @lzap, I added fixes in second commit. First commit is 1:1 clone of 0.7. I'd prefer keeping images separate in case we need to modify them. Other plugins seems to do it the same way. We can always deduplicate if it takes too much space.

lzap:
Ok I am merging, haven't read all the text, don't know what changed.

PR #909, opened 2017-07-11
Fix plugin installation doc for Debian-based sys

eLvErDe:
On debian you cannot run "bundle update something" without running at least "bundle install" once.


mmoll:
Sorry for the delay, merged, merci/danke @eLvErDe!

PR #914, opened 2017-07-17
Saltstack 

sbernhard:
- Use new puppet path
- Add hint for adding system user

### DIFF
--- a/plugins/foreman_salt/7.0/index.md
+++ b/plugins/foreman_salt/7.0/index.md
@@ -154,9 +154,9 @@ In `/etc/salt/foreman.yaml`, make the following changes:
     :proto: https
     :host: foreman.example.com
     :port: 443
-    :ssl_ca: /var/lib/puppet/ssl/certs/ca.pem
-    :ssl_key: /var/lib/puppet/ssl/private_keys/foreman.example.com.pem
-    :ssl_cert: /var/lib/puppet/ssl/certs/foreman.example.com.pem
+    :ssl_ca: "/etc/puppetlabs/puppet/ssl/ssl_ca.pem"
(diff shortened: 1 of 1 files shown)


GregSutcliffe:
Heh, I was setting up Salt-Api just last night :)

Looks good to me - I actually couldn't get SSL to work on the API (mismatch SSL versions I *think*) so I went with:

```
host: 127.0.0.1
disable_ssl: true
```
and `http` in the foreman-proxy settings. Do you think we should add that as an option for the docs?

sbernhard:
Hi, 

well, if it's working: sure. If its documented it doesn't mean that we prefer this way as it's a little bit insecure :) 

Best regards, 
Bernhard 


Von: "Greg Sutcliffe" <notifications@github.com> 
An: "theforeman/theforeman.org" <theforeman.org@noreply.github.com> 
CC: "Bernhard Suttner" <bernhard.suttner@atix.de>, "Author" <author@noreply.github.com> 
Gesendet: Dienstag, 18. Juli 2017 14:44:47 
Betreff: Re: [theforeman/theforeman.org] Saltstack (#914) 



Heh, I was setting up Salt-Api just last night :) 

Looks good to me - I actually couldn't get SSL to work on the API (mismatch SSL versions I think ) so I went with: 
host: 127.0.0.1
disable_ssl: true 


and http in the foreman-proxy settings. Do you think we should add that as an option for the docs?

GregSutcliffe:
Sweet, I'll send a separate patch for that, if you want to review it. This looks good to me though. @stbenjam I know you're busy, so I'll merge this one, but shout if you want me to give you time to review Salt stuff in general.

PR #915, opened 2017-07-17
Add docs for openscap clean_messages rake task

xprazak2:


ares (commented):
Could you add information about what version of plugins were affected? And that running this can help only once?

ares:
Thanks @xprazak2, merging!

PR #922, opened 2017-07-21
Add common pitfalls for foreman_templates

ares:


plugins/foreman_templates/5.0/index.md
```diff
@@ -84,6 +84,31 @@ After the installation there are new settings available. These are the default s
 
 Settings above are preconfigured for importing from upstream community-templates repository.
 
+## 3.1 Example of configuration github repository
+
+Let's assume you have a github account with login `template_master`. Here are the steps how to configure foreman_templates so that you can import and export templates between Foreman and your github repository. Let's assume the repository is called `templaes` and it has branch called `master`.
```
GregSutcliffe: typo in `templaes`?

plugins/foreman_templates/5.0/index.md
```diff
@@ -84,6 +84,31 @@ After the installation there are new settings available. These are the default s
...
+
+1. First we'll configure settings. Set a `branch` setting to `master`, `prefix` setting to empty string and `repo` to `ssh://git@github.com/template_master/templates`.
+2. We'll need to import github SSH public key to foreman user, run `sudo -u foreman ssh github.com`, verify and save the key. You'll probably see permission denied message since the ssh connection will not succeed but that's expected.
```
GregSutcliffe: Probably better to call this the `GitHub SSH host key` so as not to confuse with the public part of the personal key.

GregSutcliffe: also, `accept` is probably better than `import`

plugins/foreman_templates/5.0/index.md
```diff
@@ -84,6 +84,31 @@ After the installation there are new settings available. These are the default s
...
+3. Create a ssh keypair that foreman will use to communicate with github. Run `sudo -u foreman ssh-keygen`. Don't specify any passphrase.
```
GregSutcliffe: `An SSH keypair`

GregSutcliffe:
Ugh, should have started a review, sorry for spam. Anyway, mostly good, few comments, and also generally `s/github/GitHub/` :)

ares:
Thanks @GregSutcliffe for quick look, would you mind doing a second review?

GregSutcliffe:
Looks pretty good now, some minor capitalization issues but I can overlook that :)

Are we suggesting to only use the API for import/export now? The Rake task is certainly less typing...

ares:
I think the API is a better way, it allows you to specify the user and can be triggered remotely. We only need to add hammer plugin to make is convenient as foreman-rake.

GregSutcliffe:
Fair enough, the remote trigger is a fair point. Thanks!

PR #925, opened 2017-07-26
Fix update cmd for katello smart proxies

Klaas-:
the two options need an argument to work

### DIFF
--- a/plugins/katello/3.4/upgrade/smart_proxy.md
+++ b/plugins/katello/3.4/upgrade/smart_proxy.md
@@ -69,7 +69,7 @@ The installer with the --upgrade flag will run the right database migrations for
 {% highlight bash %}
 foreman-installer --scenario foreman-proxy-content --upgrade\
                   --foreman-proxy-content-certs-tar ~/myproxy.example.com-certs.tar\
-                  --certs-update-all --certs-regenerate --certs-deploy
+                  --certs-update-all --certs-regenerate true --certs-deploy true
 {% endhighlight %}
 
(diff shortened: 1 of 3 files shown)


plugins/katello/3.4/upgrade/smart_proxy.md
```diff
@@ -69,7 +69,7 @@ The installer with the --upgrade flag will run the right database migrations for
 {% highlight bash %}
 foreman-installer --scenario foreman-proxy-content --upgrade\
                   --foreman-proxy-content-certs-tar ~/myproxy.example.com-certs.tar\
-                  --certs-update-all --certs-regenerate --certs-deploy
+                  --certs-update-all --certs-regenerate true --certs-deploy true
```
ohadlevy, line 72: is this a bug in the installer ^?

Klaas-: I think its because certs-update-all is added here https://github.com/Katello/katello-installer/blob/master/hooks/boot/20-certs_update.rb and the other two are options from puppet-certs module, so I'm not sure if its a bug or if its supposed to be like this in kafo; in any case I just noticed its not working like it says it should work in docs ;)

ekohl:
@Klaas- I think this is correct but there's a merge conflict. Can you rebase so we can merge this?

Klaas-:
@ekohl done, also added it to 3.5 docs, those didn't exist in July :)

ekohl (approved):

dLobatog (approved):
Had to use this myself to regenerate certs.. thanks @Klaas- !

PR #930, opened 2017-08-11
Added better way of adding a subnet

lzap:
Final fix. I am removing the puppet hack since it does not work for me.


GregSutcliffe:
Thanks for the update!

PR #931, opened 2017-08-14
Add "true" to --certs-regenerate

coffmant:
Without this boolean specified, the smart-proxy upgrade fails with "Parameter certs-regenerate is missing a value on the command line"


GregSutcliffe:
Thanks @coffmant!

@jlsherrill @ehelms can you ACK?

ehelms (approved):

PR #933, opened 2017-08-16
Modify the katello upgrade steps.

bkearney:
Customers found issues when not rebooting the machine. The team discussed
that making the reboot step the last step would allow it to remain optional
and ensure that the testing group is verifying a single upgrade process.
This docs change brings upstream and downstream upgrade steps in line
with that discussion.

### DIFF
--- a/plugins/katello/nightly/upgrade/index.md
+++ b/plugins/katello/nightly/upgrade/index.md
@@ -32,8 +32,6 @@ Ensure your operating system is fully up-to-date:
 yum -y update
 {% endhighlight %}
 
-**NOTE**: If kernel packages are updated here (e.g. upgrading el 6.6 to 6.7), you must reboot and ensure the new kernel and SELinux policy is loaded before upgrading Katello.
-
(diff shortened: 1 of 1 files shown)


GregSutcliffe:
LGTM

sean797:
ACK from me as well. @GregSutcliffe can you merge please?

evgeni:
LGTM too, but I would also add that `katello-service stop` has to be executed before doing `yum update`. (does not have to be in this PR, but I thought it would fit in here)

ekohl:
Related: does RH have the equivalent of Debians ```/run/reboot-required``` marker file to detect if a reboot is needed?

GregSutcliffe:
I can merge (thanks for the patience :P) but it seems we need a minor edit as per @evgeni's comments? Would people prefer a merge and a new commit for that?

ekohl:
I'd prefer a merge now since this in itself is good.

GregSutcliffe:
Next PR! Ty @bkearney

PR #934, opened 2017-08-18
Fixes #20335 - fix glossary link

johnpmitsch:


GregSutcliffe:
@ehelms @jlsherrill ack? looks fine to me but I don't like touching Katello docs without confirmation :P

sean797 (changes requested):
Looks right.. But we need to do the same thing for 2.4  -> 3.5 versions.

johnpmitsch:
@sean797 thanks, updated

sean797 (approved):

PR #939, opened 2017-08-25
Add script for checking gem owners

GregSutcliffe:
As per discussion on IRC as to where to home this, I've added it here. Usage is simple enough, add "--official" to the call to only match gems with "theforeman" somewhere in their URIs. Sample output is about to be sent to the dev list for previewing.

### DIFF
--- a/scripts/check_plugin_authors.rb
+++ b/scripts/check_plugin_authors.rb
@@ -1,0 +1,43 @@
+#!/usr/bin/env ruby
+#
+# Script to check which plugins have a single author
+# Copyright: Red Hat 2017
+# LICENSE: GPLv3
+
+require 'json'
+require 'rest-client'
+
+def get_gem_names(search)
+  page=1
+  gems=[]
+  url="https://rubygems.org/api/v1/search.json?query=#{search}"
+
+  response = RestClient.get('https://rubygems.org') # test connection and initialize "response"
+
+  $stderr.puts "Searching for gems matching '#{search}'"
+  while response.body != "[]"
+    $stderr.puts "Page #{page}"
+    response = RestClient.get(url+"&page=#{page}")
+    JSON.parse(response.body).each do |gem|
+      # Strip gems not under our namespace in some way - anything with "theforeman" in the URIs should do
+      regex = /(theforeman|katello)/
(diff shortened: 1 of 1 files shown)


scripts/check_plugin_authors.rb
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/ruby
```
ares: `#!/usr/bin/env ruby` perhaps?

scripts/check_plugin_authors.rb
```diff
@@ -0,0 +1,43 @@
...
+      next if ( ARGV[0] == "--official" and not gem.select { |k,v| k =~ /uri/}.values.join(" ") =~ regex )
+      gems << gem['name']
+    end
+    page += 1
+  end
+  gems.sort
+end
+
+foreman_gems = get_gem_names('foreman')
+proxy_gems   = get_gem_names('smart_proxy')
```
ares: maybe `smart-proxy` too?

GregSutcliffe: I ran that with no hits, but yeah, should add it to the search anyway

scripts/check_plugin_authors.rb
GregSutcliffe: Good catch, will fix later, gotta run :P

GregSutcliffe:
@ares - updated, sorry for the delay.

GregSutcliffe:
ping @ares?

scripts/check_plugin_authors.rb
```diff
@@ -0,0 +1,43 @@
...
+
+require 'json'
+require 'rest-client'
+
+def get_gem_names(search)
+  page=1
+  gems=[]
+  url="https://rubygems.org/api/v1/search.json?query=#{search}"
+
+  response = RestClient.get('https://google.com') # test connection and initialize "response"
```
ekohl: Wouldn't it be better to get rubygems.org?

GregSutcliffe: fair enough ;)

ares:
Looks good, thanks @GregSutcliffe and @ekohl, merging.

PR #947, opened 2017-09-18
image installs blog post

Raboo:
Ok, here is the re-post of my blog post about image installs.
Let me know if there is something I need to change or fix.

### DIFF
--- a/_posts/2017-09-19-linux-image-installs-for-bare-metal.md
+++ b/_posts/2017-09-19-linux-image-installs-for-bare-metal.md
@@ -1,7 +1,7 @@
 ---
 layout: post
 title: Linux image installs for Bare Metal
-date: 2017-09-19 01:33:07
+date: 2017-10-18 01:33:07
 author: Elias Abacioglu
 tags:
 - foreman
--- a/_posts/2017-09-19-linux-image-installs-for-bare-metal.md
+++ b/_posts/2017-09-19-linux-image-installs-for-bare-metal.md
@@ -1,0 +1,51 @@
+---
+layout: post
+title: Linux image installs for Bare Metal
+date: 2017-09-19 01:33:07
+author: Elias Abacioglu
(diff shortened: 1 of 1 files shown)


mmoll:
Content-wise this is good, I think. Maybe the division of the paragraphs and line breaks could be a bit tweaked, it looks a bit like copypasting text from a word document into wordpress (or something like that ;)) in my rendered version.

Raboo:
Well I'm not a seasoned blogger.. I had those extra line breaks to make it easier to read. But they are removed. Does this look better?

ohadlevy:
@Raboo nice :) /cc @lzap

lzap:
Yeah thanks. I clearly remember me saying by the beer that I will finish my POC patch "very soon". It's year ago now and it is still there. It now uses udpcast but I have a better idea now - incorporate remote execution into discovered hosts. Should not be hard, then you would initiate provisioning or partitioning via Foreman powerful template system. How is that? Do you use ReX?

Raboo:
It all sounds good as a goal. Would be a lot faster than having to reboot into the same image..
But lets not overextend :) Time is a big constrain, so small steps until we reach the goal.

I've been running baremetal image installs in production since shortly after CfgMgmtCamp 2017.
I haven't had much problems and it's a lot easier to maintain than those preseed.

@lzap what is ReX?

Hmm why are we discussing this in a PR?
Just ping me on IRC 😃

lzap:
It renders fine, I see a bunch of tags but they does not seem to be used at all.

lzap:
Thanks @Raboo

Raboo:
True, the tags don't show, don't know why..

GregSutcliffe:
Hmm, they used to be displayed, something must have gotten "tidied up" somewhere. I'll take a look when I have a moment :)

Thanks @Raboo btw!

PR #955, opened 2017-10-11
Fix class import filter regex

ekohl:
This made it work for me. I've only changed 1.15 and 1.16 since I confirmed it on 1.15. Might need to be fixed in older releases as well.


dLobatog (approved):
Thanks @ekohl

PR #963, opened 2017-11-01
Instruct to stop Katello before yum update

akofink:
Step 17 in the [downstream docs](https://access.redhat.com/documentation/en-us/red_hat_satellite/6.2/html/installation_guide/upgrading_satellite_server_and_capsule_server).

### DIFF
--- a/plugins/katello/3.5/upgrade/index.md
+++ b/plugins/katello/3.5/upgrade/index.md
@@ -48,6 +48,11 @@ Update the Foreman and Katello release packages:
 
 ## Step 4 - Update Packages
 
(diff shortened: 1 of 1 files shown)


chris1984 (approved):
LGTM

mmoll:
merged, thanks @akofink!

ekohl:
Would it make sense to do this in the 3.4 docs as well?

akofink:
I do believe this applies to 3.4. @ehelms?

ehelms:
It should but we haven't tested it. So I am tempted to not put it in the instructions.

PR #968, opened 2017-11-07
Add ldlinux.c32 into TFTP setup doc

xprazak2:
Setting up TFTP boot on Fedora 26,
a host was complaining about missing ldlinux.c32.
After I copied the file from syslinux into tftpboot,
host started provisioning as expected.


lzap:
Thanks this is a long overdue, required file to have.

ares:
should it be reflected also in our installer module? https://github.com/theforeman/puppet-foreman_proxy/blob/8f0848d27446bd65a3caa5953e9f7304d3964b70/manifests/params.pp#L28-L33 or will it bite us on some next rhel/centos version?

lzap:
I believe this is only required when you are installing on Fedora, thus this is not supported now. In the future, yeah.

PR #982, opened 2017-12-06
Update index.md

r3dphoenix:
Used the foreman.org site to install Katello on Centos7, I found it fails during dependancy check.
Determined the repo for foreman should be changed from 1.5  to  1.6  which resolves the installation issue.


ekohl (approved):
You're right, 3.5 is based on 1.16.

mmoll:
merged, thanks @r3dphoenix!

PR #984, opened 2017-12-10
blog: add post on Discourse migration

GregSutcliffe:
Basically the same text as the mail I just sent to the mailing lists - just for those who might only be on the rss feed, and for a convenient thing to link to elsewhere. I'll tweet a link to this once it's published.

Will do another PR for adding Discourse links to the website itself.

### DIFF
--- a/_includes/alerts.html
+++ b/_includes/alerts.html
@@ -1,5 +1,7 @@
-<!-- <div class='tease'>      -->
-<!--   <p>                    -->
-<!--     something important! -->
-<!--   </p>                   -->
-<!-- </div>                   -->
+ <div class='tease'>
+   <p>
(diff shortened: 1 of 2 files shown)


GregSutcliffe:
Does anyone think I should also put this in the frontpage alertbox?

ares:
> Does anyone think I should also put this in the frontpage alertbox?

that can't hurt

GregSutcliffe:
@ares done, sorry for the delay. Updated the date on the post.

ares:
LGTM, the alert is on the bottom and not too visible but if that's what you had in mind, I guess it can be merged

GregSutcliffe:
@ares it appears at the top of the frontpage on my local copy. I'll go ahead and merge since we're into the weekend,will send another PR if it doesn#'t work live FSR. ty!

PR #986, opened 2017-12-11
Foreman version for Katello 3.5 is 1.16

coffmant:


akofink (approved):
ACK from me

GregSutcliffe:
Ty!

ekohl:
Given we also had https://github.com/theforeman/theforeman.org/pull/982 - is this something we should put in our release process somewhere?

akofink:
I agree a list of files to change would be nice. The current process is basically to grep for 3.4 (and 1.15, which I missed).

PR #987, opened 2017-12-13
Install python-django with installation

johnpmitsch:
This is a workaround for an issue in pulp that is currently
being resolved, but is not in our 3.5 release yet.


chris1984 (approved):

akofink (approved):
LGTM - Should we put a comment or something to make sure we remove it later?

johnpmitsch:
@akofink I'm not sure what good a comment will do since this page generally doesn't get updated between releases, it technically won't hurt anything if it's still there after things are fixed

jlsherrill:
+1 if its there even after the issue is fixed, it shouldn't cause an issue

PR #993, opened 2017-12-30
Change mailing list references for Discourse links [do not merge]

GregSutcliffe:
Preparation for tomorrow night's migration. This changes most references to the mailing list (I only altered manuals back as far as 1.14, everything older there is historical). I also updated the Katello build notes since they reference the dev list quite a bit

I've put Discourse on the frontpage and moved the Contribute page to the docs landing page (which is also slightly altered to draw attention to the manual, release notes, and contribution pages). It's not ideal but it'll do until we can sort out a proper rebuild of the site ;)

This also reverts the alert on the frontpage about the migration, since the migration will be done once this is merged.

@tbrisker @orrabin since you may (possibly?) actually be around tomorrow, could you review? I'm happy to merge as part of the migration, but some eyes on this first would be good :)


documentation.md
```diff
@@ -31,18 +31,27 @@ version: '1.16'
 ---------------------------
 
 <div class='row'>
-  <div class='col-xs-3 center'>
+  <div class='col-xs-4 center'>
```
tbrisker, line 34: I think making part of the page 3 columns and part 4 doesn't look too good, I'd prefer having all icons on the same grid (but maybe that's just my OCD)

support.md
```diff
@@ -40,38 +40,34 @@ The issue tracker (Redmine) should be used to check for existing issues or to fi
...
 
 * Don't hijack threads - start new threads for new questions
-* Don't attach or paste in huge logs or attachments - upload them to a website
-  and include links in your post
-* Let us know if your question is answered - Replying to the thread with
-  *[SOLVED]* in the title lets others know that you no longer need help, and
-  helps others to find solutions when searching the archives
+* Don't attach or paste in huge attachments - upload them to a website and include links in your post
+* Let us know if your question is answered - the solution can be marked using
+  the thread tools, lets others know that you no longer need help, and helps
```
tbrisker: maybe: `the thread tools. This lets others...`?

tbrisker (commented):
Looks good in general, a couple of minor comments inline.

support.md
GregSutcliffe: Yeah, will fix on merge, ty

documentation.md
GregSutcliffe: The intention is to draw attention to (IMO) the three important icons on that page, given there are a lot now. Also 11 icons is a prime number, so there's no neat way to get them all on a single grid :)

GregSutcliffe: I'm going to merge as is, just to get this done, but I'll happily take a new PR for a better layout.

PR #1004, opened 2018-01-16
fix incorrect foreman answer file

sean797:


_includes/manuals/1.16/5.8.1_database.md
```diff
@@ -1,2 +1,2 @@
 
-All Foreman instances in a cluster/group must point to the same database. This can be done during the initial installation (through flags or altering `foreman_installer_answers.yaml`) or by directly altering `/etc/foreman/database.yaml` and pointing the correct environment (usually production) to your Foreman DB, then restarting Foreman.
+All Foreman instances in a cluster/group must point to the same database. This can be done during the initial installation (through flags or altering `/etc/foreman-installer/scenarios.d/foreman-answers.yaml`) or by directly altering `/etc/foreman/database.yaml` and pointing the correct environment (usually production) to your Foreman DB, then restarting Foreman.
```
ekohl: I'm not sure we should recommend changing the answers file at all - flags should be much more stable and work on multiple scenario's.

sean797: I'm not sure it's fair to say the installer flag are more stable, since they are created based on the parameters name, just like the parameters in the answer file.

I like to manually edit the answer file when I'm changing multiple values as I find that easier, maybe that's just me.

ekohl: There's a difference between advanced users who know what they're doing and recommending this to users. The filename can be different on katello for example. There the flag would work but these instructions wouldn't.

sean797: Sure, i get what you saying 👍 updated. Thanks @ekohl

_includes/manuals/1.16/5.8.1_database.md
```diff
@@ -1,2 +1,2 @@
 
-All Foreman instances in a cluster/group must point to the same database. This can be done during the initial installation (through flags or altering `foreman_installer_answers.yaml`) or by directly altering `/etc/foreman/database.yaml` and pointing the correct environment (usually production) to your Foreman DB, then restarting Foreman.
+All Foreman instances in a cluster/group must point to the same database. This can be done during the initial installation using foreman-installer or by directly altering `/etc/foreman/database.yaml` and pointing the correct environment (usually production) to your Foreman DB, then restarting Foreman.
```
ekohl: It can also be done by rerunning the installer with new arguments.

ekohl (changes requested):

sean797:
Updated, thanks @ekohl 👍

ekohl (approved):

dLobatog (approved):
Thanks @sean797

PR #1013, opened 2018-02-07
Adding Discovery 10 and 11 docs

lzap:
Looks like I forgot to add 10.0 when I released it.

This patch also adds 11.0 docs, there is unreleased warning box for now.

Also adds a paragraph about KMS driver issue in separate commit


ares:
Thanks @lzap, merging.

PR #1018, opened 2018-02-17
Fix missing Katello 3.6 sidebar

ehelms:


ehelms:
@mmoll this should fix the Jekyll error

mmoll:
@ehelms indeed it does, thanks!

PR #1021, opened 2018-02-21
use postgresql service instead of rh-postgresql96-postgresql

evgeni:
it's provided by the syspath packages anyways, and makes our life easier


ekohl (approved):

lzap:
WFM thanks.

PR #1023, opened 2018-02-22
Add shared-modules section to handbook->javascript

sharvit:
Add documentation about shared node-modules and the `config/webpack.vendor.js`

### DIFF
--- a/handbook.md
+++ b/handbook.md
@@ -152,6 +152,12 @@ We use eslint to enforce linting rules. Run eslint by typing `npm run lint` in t
 * Pollute the global namespace. If, for some reason, you must expose a function/value globally, use the ```window.tfm``` object as implemented in ```bundle.js```.
 * Remove global functions from existing code without deprecating them first. Deprecation can be done by calling the `tfm.tools.deprecate` function.
 
+#### Shared modules
+Usually, when we add a new node-module, we want it to be shared with plugins (Unless we have a reason not to share it).
+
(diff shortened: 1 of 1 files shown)


handbook.md
```diff
@@ -152,6 +152,12 @@ We use eslint to enforce linting rules. Run eslint by typing `npm run lint` in t
 * Pollute the global namespace. If, for some reason, you must expose a function/value globally, use the ```window.tfm``` object as implemented in ```bundle.js```.
 * Remove global functions from existing code without deprecating them first. Deprecation can be done by calling the `tfm.tools.deprecate` function.
 
+#### Shared modules
+Usually, when we add a new node-module, we want it to be shared with plugins (Unless we have a reason not share it).
```
dLobatog: `a reason not to share it`

handbook.md
```diff
@@ -152,6 +152,12 @@ We use eslint to enforce linting rules. Run eslint by typing `npm run lint` in t
...
+
+In order to do so, we let [webpack](https://webpack.js.org/) to create a `vendor.js` file with the relevant modules.
```
dLobatog: `we let webpack create`

dLobatog (changes requested):
Great addition, thanks @sharvit

sharvit:
Thanks @dLobatog, updated!

dLobatog (approved):
Merging! :medal_military: @sharvit

PR #1025, opened 2018-02-24
add travis testing

mmoll:


mmoll:
This at least does ensure that the site can be built.
Enabling travis on this repo would need to be done by somebody with the respective permissions.

bastilian:
But the site would be already deployed?

If we'd want to prevent a broken site, we could switch to master as the default branch and setup travis to push back to `gh-pages`.

I have a site that has that setup (basically) via [rake tasks](https://github.com/rossconf/rossconf.io/blob/master/Rakefile). The `:after_success` task adds the credentials to push, the `:deploy` task runs middlemans deploy, which builds and commits the site to `gh-pages` to push.

mmoll:
This would be only to test pull requests. I think the situation where one does push a green PR and the site building on Jenkins then fails is very rare. Also, if the build on jenkins fails, nothing will get deployed.

lzap (approved):
WFM I wish Travis could upload generated site somewhere. It does support this, but only to S3.

bastilian:
@lzap Travis can be bent to push or upload the site to any target. not out of the box maybe, but a shell script after success can do the trick

lzap:
@mmoll got time to look into this? If Travis also uploads generated site to some scratch directory, we could actually read the resulting texts rendered! That would be super awesome. If not, I am good to merge if there are no objections.

mmoll:
I didn't really have time, yet... What space would be push the generated pages to?

lzap:
Yeah that's the question, depends on technical limitations. I assume we'd need to have credentials here, that's a problem. Perhaps we won't find any place. I am fine merging this anyway.

lzap:
Let's merge it to see how it works.

mmoll:
Travis still needs to get enabled now.... @GregSutcliffe @ehelms @ohadlevy ...

ehelms:
Done!

PR #1033, opened 2018-03-01
How to set up Ansible SSH keys

dLobatog:


plugins/foreman_ansible/1.x/index.md
```diff
@@ -87,8 +89,22 @@ The project is not packaged, but you can find the sources here at [theforeman/fo
...
+
+```
+sudo mkdir /usr/share/foreman-proxy/.ansible
+sudo chown foreman-proxy.foreman-proxy /usr/share/foreman-proxy/.ansible
+sudo chown foreman-proxy.foreman-proxy /usr/share/foreman-proxy
+```
+
+You may need do the same for the `foreman` user, so that it can run Ansible directly from the Foreman host without a Foreman Proxy.
+
+If you have Foreman Remote Execution already, it is likely you have a ssh key in `/usr/share/foreman-proxy/.ssh/id_rsa_foreman_proxy`. You may copy it to `/usr/share/foreman-proxy/.ssh/id_rsa` so that Ansible is able to use it (ensure that the file is readable by the `foreman-proxy` user)
```
ares: does ansible require it to be named `id_rsa`? can't it also use `id_rsa_foreman_proxy`? and it might be useful to mentione that for ansible, the key can't be protected with passphrase.

dLobatog: I tried using `ansible_private_key_file=/usr/share/foreman-proxy/.ssh/id_rsa_foreman_proxy` on `/usr/share/foreman-proxy/.ansible.cfg`, to no avail, if you know how to make that work that'd be a good thing to put in https://github.com/theforeman/puppet-foreman_proxy/blob/master/templates/plugin/ansible.cfg.erb

plugins/foreman_ansible/1.x/index.md
```diff
@@ -87,8 +89,22 @@ The project is not packaged, but you can find the sources here at [theforeman/fo
 
 In Foreman, you should add whatever Ansible hosts you want to submit facts from to the setting `trusted_puppetmaster_hosts`. Change it at Administer > Settings, Auth tab.
 
+Ansible will run using the SSH key of the smart-proxy or Foreman. These keys are automatically set by Foreman Remote Execution, so you may see them at /usr/share/foreman-proxy/.ssh . The foreman-proxy user needs write access to '/usr/share/foreman-proxy/.ansible' at least to create the '.ansible' local directory. We are working on making the instller automatically do this, but if it didn't do it.
```
ekohl: Can we move that to ```/etc/foreman-proxy/ssh``` and only install a symlink in ```/usr/share/foreman-proxy```?

plugins/foreman_ansible/1.x/index.md
ares: I thought ansible would use any ssh key available to the user, like ssh does but I could be wrong, the option accroding to man should be `key_file` or private_key` so I'd try `ansible_private_key=...`

ares: and looking at forklift's ansible.cfg, I see `private_key_file = $HOME/.vagrant.d/insecure_private_key` so I guess it should be possible

plugins/foreman_ansible/1.x/index.md
dLobatog: We can do that, but for 1.16 users this is not an option

plugins/foreman_ansible/1.x/index.md
dLobatog: :man_facepalming: I just saw we're sending a `ansible_private_key_file = nil` on the inventory when the setting is empty. Therefore it just points to `~.ssh/id_rsa` by default. When I change `ansible_private_key_file` at the setting or at the host level, it's fine.

I should be documenting this here - for some reason the setting does not show up in Settings > Ansible, local blip or are you unable to see the setting too @ares ?

ares: I don't see it either even though I see the definition in ./app/models/setting/ansible.rb

dLobatog: https://github.com/theforeman/foreman_ansible/pull/142

dLobatog (commented):
@ares @ekohl Anything blocking this? I've updated the section to include a line about `ansible_private_key_file`

ares (approved):
leaving for @ekohl to ack, if he does not add comment I'll merge after a while

plugins/foreman_ansible/1.x/index.md
```diff
@@ -46,12 +46,14 @@ For Ansible versions older than 2.2:
...
 You can configure it via the following environment variables:
 
-* FOREMAN_URL: the URL of your Foreman installation (default "http://localhost:3000")
+* FOREMAN_URL: the URL of your Foreman installation (default "http://localhost:3000") - change it to your Foreman URL, e.g: "https://myforeman.mylocal.lan", it should not be port 3000.
 * FOREMAN_SSL_CERT: The public key when using SSL client certificates (default "/etc/foreman/client_cert.pem")
 * FOREMAN_SSL_KEY: The private key when using SSL client certificates (default  "/etc/foreman/client_key.pem")
 * FOREMAN_SSL_VERIFY: wether to verify SSL certificates. Use *False*
   to disable certificate checks. You can also set it to CA bundle (default is "True").
 
+If you are using the Katello plugin, the SSL certificate and key are set at `/etc/httpd/conf.d/05-katello-ssl.conf`. SSLCertificateFile, SSLCertificateKeyFile contain the values for FOREMAN_SSL_CERT and FOREMAN_SSL_KEY. By default these are at `/etc/pki/katello/certs/katello-default-ca.crt` and `/etc/pki/katello/private/katello-default-ca.key`.
```
ekohl: Are you sure this is true? On my katello nightly box `/etc/httpd/conf.d/05-katello-ssl.conf` doesn't exist. AFAIK katello uses the same apache file, but different values.

plugins/foreman_ansible/1.x/index.md
ekohl: I don't like putting promises of a change in documentation. Either refer to a redmine issue or leave it out IMHO.

plugins/foreman_ansible/1.x/index.md
```diff
@@ -87,8 +89,24 @@ The project is not packaged, but you can find the sources here at [theforeman/fo
 
 In Foreman, you should add whatever Ansible hosts you want to submit facts from to the setting `trusted_puppetmaster_hosts`. Change it at Administer > Settings, Auth tab.
 
+Ansible will run using the SSH key of the smart-proxy or Foreman. These keys are automatically set by Foreman Remote Execution, so you may see them at /usr/share/foreman-proxy/.ssh . The foreman-proxy user needs write access to '/usr/share/foreman-proxy/.ansible' at least to create the '.ansible' local directory. We are working on making the instller automatically do this, but if it didn't do it.
+
+```
+sudo mkdir /usr/share/foreman-proxy/.ansible
+sudo chown foreman-proxy.foreman-proxy /usr/share/foreman-proxy/.ansible
+sudo chown foreman-proxy.foreman-proxy /usr/share/foreman-proxy
```
ekohl: Is this line actually needed? From a security perspective it undesirable and a yum upgrade might just reset it for you (since it's not a config file).

plugins/foreman_ansible/1.x/index.md
```diff
@@ -87,8 +89,24 @@ The project is not packaged, but you can find the sources here at [theforeman/fo
...
+
+The SSH key used for the execution can be set by the `ansible_private_key_file` setting (find it under Administer > Settings > Ansible tab). If you leave that field empty, `/usr/share/foreman/.ssh/id_rsa` or `/usr/share/foreman-proxy/.ssh/id_rsa` will be used by default.
+
+If you have Foreman Remote Execution already, it is likely you have a ssh key in `/usr/share/foreman-proxy/.ssh/id_rsa_foreman_proxy`. You may copy it to `/usr/share/foreman-proxy/.ssh/id_rsa` so that Ansible is able to use it (ensure that the file is readable by the `foreman-proxy` user)
+
 If the Foreman setting `create_new_host_when_facts_are_uploaded` (Puppet tab) is true, and $HOSTNAME doesn't exist in Foreman, it will autocreate that host in Foreman. If it already exists, it will update the facts.
 
+If you want to use Ansible and submit facts/reports to Foreman, through the callback, you should add whatever hosts (again, except Smart Proxies) you want to submit facts **from** to the setting `trusted_puppetmaster_hosts`. Change it at Administer > Settings, Auth tab. e.g: If you're running Ansible from host 'A', which SSHs into host 'B', you need to add host 'A'
```
ekohl: s/whatever/the/

plugins/foreman_ansible/1.x/index.md
```diff
@@ -87,8 +89,24 @@ The project is not packaged, but you can find the sources here at [theforeman/fo
...
 
+Ansible will run using the SSH key of the smart-proxy or Foreman. These keys are automatically set by Foreman Remote Execution, so you may see them at /usr/share/foreman-proxy/.ssh . The foreman-proxy user needs write access to '/usr/share/foreman-proxy/.ansible' at least to create the '.ansible' local directory. We are working on making the instller automatically do this, but if it didn't do it.
+
+```
+sudo mkdir /usr/share/foreman-proxy/.ansible
+sudo chown foreman-proxy.foreman-proxy /usr/share/foreman-proxy/.ansible
+sudo chown foreman-proxy.foreman-proxy /usr/share/foreman-proxy
+```
+
+You may need do the same for the `foreman` user, so that it can run Ansible directly from the Foreman host without a Foreman Proxy.
```
ekohl: Personally I dislike this general design pattern that we can run it both with and without a proxy. It complicates the stack and IMHO we should avoid ever implementing it in the first place.

ekohl (changes requested):

plugins/foreman_ansible/1.x/index.md
```diff
@@ -46,12 +46,16 @@ For Ansible versions older than 2.2:
...
+* FOREMAN_URL: the URL of your Foreman installation (default "http://localhost:3000") - change it to your Foreman URL, e.g: "https://myforeman.mylocal.lan", it should not be port 3000.
 * FOREMAN_SSL_CERT: The public key when using SSL client certificates (default "/etc/foreman/client_cert.pem")
 * FOREMAN_SSL_KEY: The private key when using SSL client certificates (default  "/etc/foreman/client_key.pem")
 * FOREMAN_SSL_VERIFY: wether to verify SSL certificates. Use *False*
   to disable certificate checks. You can also set it to CA bundle (default is "True").
 
+If you're using Foreman without Katello, the SSL certificate and key are set at `/etc/httpd/conf.d/foreman.conf`. SSLCertificateFile, SSLCertificateKeyFile contain the values for FOREMAN_SSL_CERT and FOREMAN_SSL_KEY.
+
+If you are using the Katello plugin, the SSL certificate and key are set at `/etc/httpd/conf.d/05-foreman-ssl.conf`. SSLCertificateFile, SSLCertificateKeyFile contain the values for FOREMAN_SSL_CERT and FOREMAN_SSL_KEY. By default these are at `/etc/pki/katello/certs/katello-apache.crt` and `/etc/pki/katello/private/katello-apache.key`.
```
dLobatog, line 57: I checked on 1.16 and the keys are at `/etc/httpd/conf.d/foreman.conf` and `/etc/httpd/conf.d/05-foreman.conf` respectively, let me know if you think that is wrong and should instead point to some other file

plugins/foreman_ansible/1.x/index.md
```diff
@@ -87,8 +91,23 @@ The project is not packaged, but you can find the sources here at [theforeman/fo
 
 In Foreman, you should add whatever Ansible hosts you want to submit facts from to the setting `trusted_puppetmaster_hosts`. Change it at Administer > Settings, Auth tab.
 
+Ansible will run using the SSH key of the smart-proxy or Foreman. These keys are automatically set by Foreman Remote Execution, so you may see them at /usr/share/foreman-proxy/.ssh . The foreman-proxy user needs write access to '/usr/share/foreman-proxy/.ansible' at least to create the '.ansible' local directory.
+
+```
+sudo mkdir /usr/share/foreman-proxy/.ansible
+sudo chown foreman-proxy.foreman-proxy /usr/share/foreman-proxy/.ansible
```
dLobatog, line 98: Removed /usr/share/foreman-proxy from here

plugins/foreman_ansible/1.x/index.md
```diff
@@ -87,8 +91,23 @@ The project is not packaged, but you can find the sources here at [theforeman/fo
...
+```
+
+You may need do the same for the `foreman` user, so that it can run Ansible directly from the Foreman host without a Foreman Proxy.
+
+The SSH key used for the execution can be set by the `ansible_private_key_file` setting (find it under Administer > Settings > Ansible tab). If you leave that field empty, `/usr/share/foreman/.ssh/id_rsa` or `/usr/share/foreman-proxy/.ssh/id_rsa` will be used by default.
```
dLobatog, line 103: Removed the promise to fix this in the installer

plugins/foreman_ansible/1.x/index.md
```diff
@@ -87,8 +91,23 @@ The project is not packaged, but you can find the sources here at [theforeman/fo
...
+
+If you have Foreman Remote Execution already, it is likely you have a ssh key in `/usr/share/foreman-proxy/.ssh/id_rsa_foreman_proxy`. You may copy it to `/usr/share/foreman-proxy/.ssh/id_rsa` so that Ansible is able to use it (ensure that the file is readable by the `foreman-proxy` user), or change the Setting `ansible_private_key_file` to point to it.
```
dLobatog, line 105: I added the last bit about using the setting

dLobatog (commented):
@ekohl Updated according to your comments, let me know what you think

dLobatog (commented):
@ares @ekohl anything missing here?

ares:
quick read -> good from my side, waiting on @ekohl to confirm

plugins/foreman_ansible/1.x/index.md
ekohl: It should *always* be ```05-foreman-ssl.conf```. Both with and without katello since puppet-foreman sets that up. In the katello scenario we just change the default answers from puppet locations to katello locations.

That said, I think in the katello scenario the correct certificate is the one we install for puppet server to talk back to foreman. It's guaranteed to be signed by the correct CA (remember there can be custom certificates signed by a diffferent CA). It's also the same function: report callbacks. See https://github.com/theforeman/puppet-certs#certificates-overview

Maybe @stbenjam has some good insight here, but I think it would make sense to implement this in the installer.

dLobatog:
https://github.com/theforeman/puppet-foreman_proxy/pull/415

plugins/foreman_ansible/2.x/index.md
```diff
@@ -91,10 +95,22 @@ Check Ansible's [dynamic inventory documentation](http://docs.ansible.com/ansibl
...
+
+```
+sudo mkdir /usr/share/foreman-proxy/.ansible
+sudo chown foreman-proxy.foreman-proxy /usr/share/foreman-proxy/.ansible
+```
+You may need do the same for the `foreman` user, so that it can run Ansible directly from the Foreman host without a Foreman Proxy.
+
+The SSH key used for the execution can be set by the `ansible_private_key_file` setting (find it under Administer > Settings > Ansible tab). If you leave that field empty, `/usr/share/foreman/.ssh/id_rsa` or `/usr/share/foreman-proxy/.ssh/id_rsa` will be used by default.
+
+If you have Foreman Remote Execution already, it is likely you have a ssh key in `/usr/share/foreman-proxy/.ssh/id_rsa_foreman_proxy`. You may copy it to `/usr/share/foreman-proxy/.ssh/id_rsa` so that Ansible is able to use it (ensure that the file is readable by the `foreman-proxy` user), or change the Setting `ansible_private_key_file` to point to it.
```
bastilian: A link to the [Remote Execution Manual](https://www.theforeman.org/plugins/foreman_remote_execution/1.3/index.html#1.ForemanRemoteExecution1.3Manual) here or above would make sense.

ekohl:
@dLobatog Right now I don't know where I could set these variables so I can just run ```ansible``` without having to think about setting those variables. Is there a way to configure this?

dLobatog (commented):
@ekohl Which variables, `FOREMAN_HOST` & friends (used by the Ansible callback)? You may use `/etc/environment` if you just want to call `ansible` from your CLI (explained in line 36 of the new docs / 32 in the old docs)

ekohl:
With the packaging change in nightly and https://github.com/theforeman/puppet-foreman_proxy/pull/424 I think we should have a decent out of the box experience. The known hosts is still a problem and even though we have `host_key_checking = False` in the default config, it appears to be ignored.

ekohl (changes requested):
Could you revisit this? I think we should be promoting `ansible.cfg` over environment variables. In nightly we should now set this up automatically though a warning that 2.5 is broken would be good.

GregSutcliffe:
Bump, are we still needing this? Seems like yes ;)

ekohl:
I think it was replaced by https://github.com/theforeman/theforeman.org/pull/1093

dLobatog (commented):
Yup, not needed anymore I would say.

PR #1037, opened 2018-03-07
Added post about filters

lzap:
SSIA

Please *fix* typos, don't comment them. There are *many*.

### DIFF
--- a/_posts/2018-03-07-foreman-117-fact-filtering.md
+++ b/_posts/2018-03-07-foreman-117-fact-filtering.md
@@ -91,3 +91,8 @@ be used as well):
 
     records = FactName.where(query).delete_all
     puts "Removed #{records} fact names"
+
+Please note that the settings for "Exclude pattern for facts stored in Foreman" and "Ignore
+interfaces with matching identifier" have some additional logic to simplify regex writing. For
+example, if "exclude pattern for facts" contains `foo01`, it will match `interfaces::foo01::mac`.
+Typically you will not want to copy/paste from one field to another.
--- a/_posts/2018-03-07-foreman-117-fact-filtering.md
+++ b/_posts/2018-03-07-foreman-117-fact-filtering.md
@@ -1,0 +1,93 @@
+---
+layout: post
+title: Foreman 1.17 fact filtering
+date: 2018-03-07 14:27:47
+author: Foreman
+tags:
+- foreman
+---
+
+New version of Foreman comes with handy feature - fact and NIC filtering. When
+Foreman manages hypervisors or container hosts with many bridge or virtual
+devices, fact names are too often updated and when under heavy load, database
+locks or rollbacks can happen as fact upload is one of the most common API
+calls. Also for each individual NIC reported, Foreman creates NIC entry and by
+design it never deletes those from the database. This can lead to situations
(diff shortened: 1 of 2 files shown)


beav:
post looks good from a content perspective, did not update for typo yet

lzap:
Thanks, I was typing really quick... Hope it explains a lot.

lzap:
Thanks for the typos!

_posts/2018-03-07-foreman-117-fact-filtering.md
```diff
@@ -91,3 +91,8 @@ be used as well):
 
     records = FactName.where(query).delete_all
     puts "Removed #{records} fact names"
+
+Please note that the settings for "Exclude pattern for facts stored in Foreman" and "Ignore
+interfaces with matching identifier" have some additional logic to simplify regex writing. For
+example, if "exclude pattern for facts" contains `foo01`, it will match `interfaces::foo01::mac`.
+Typically you will not want to copy/paste from one field to another.
```
ShimShtein, line 98: Actually it tries to equalize the semantics for both settings: if you want to get rid of all interfaces that start with `veth`, it means that you also don't want facts that are under `veth` in facts subtree (for hierarchical facts). The regex that is generated under the hood, tries to match common hierarchy delimiters to use as start and finish for fact name matching: we know that RHSM facts are using `::` do delimit fact names, so the regex will be "starts with ::, contains the string `veth` and delimited by ::"; we also know that puppet delimits its hierarchy with `_`, so for that sort of facts, we will add "satrts with _, contains the string `veth` and delimited by _" regex to the search.
Both cases express user's intent to remove `veth*` facts, so I do think that copy and paste is acceptable here. I would say that facts filter should be a superset of the interfaces filter, since the user will want additional facts be filtered out, like blockdevices e.t.c.

beav: The issue I hit with copy/paste was `*veth*`, which is allowed by "exclude pattern" but not "ignore interfaces".

ShimShtein: Ah, OK.

beav:
@lzap ACK from me except for the last paragraph, which I can't ack because I wrote it :memo:

_posts/2018-03-07-foreman-117-fact-filtering.md
```diff
@@ -0,0 +1,98 @@
...
+* vlinuxbr*
+* vovsbr
+
+All network interfaces detected by Foreman matching any of these patterns are
+silently ignored and never added into Foreman inventory. This setting solves part one of
+the problem. Now let's focus on what's new in Foreman 1.17.
+
+The new setting is named "Exclude pattern for facts stored in Foreman" and
+works similarly, but for fact names. This prevents Foreman from performing
+unnecessary updates in fact names database table which can easly become a
```
ShimShtein: which can eas**i**ly become (missing "i")

ShimShtein (commented):
Small typo, but besides it :+1:

beav:
@lzap article is ready to go at this point

lzap:
Thanks squashed, merged.

PR #1044, opened 2018-03-20
Link to the puppetfile-pinner script

ekohl:
I'm not sure where this script should live so this is partially a RFC. It's also not tested with all possible options, but it works for what we have in our current Puppetfile.

### DIFF
--- a/plugins/katello/developers.md
+++ b/plugins/katello/developers.md
@@ -289,14 +289,7 @@ git push
 
 ### Step 7. Pin katello-installer puppet modules for release
 
-On the katello-installer's KATELLO-X.Y branch, we need to pin the modules to a minor release.
-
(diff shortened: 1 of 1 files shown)


ehelms:
At first pass, this sounds like something that should be run as part of the release process by the Rakefile in the installer?

ekohl:
@ehelms I thought that too. Should we duplicate it in both foreman-installer and katello-installer?

ehelms:
For now I think it has to live in both.

ekohl:
Will update after https://github.com/theforeman/foreman-installer/pull/266 is merged and a similar PR is merged to katello-installer.

ekohl:
The PRs to both installers have now been merged and this can be merged as well.

PR #1052, opened 2018-04-04
Base structure for 1.18 manual

xprazak2:
Initial commit for 1.18 manual. It contains copied previous version stripped of release notes.


tbrisker (approved):
Thanks @xprazak2, I guess now that it's branched we should also have the manual ready so we can start updating it.

PR #1058, opened 2018-04-11
Update Katello Docs for 3.6 GA

zjhuntin:


ehelms:
Here needed as well -- https://github.com/theforeman/theforeman.org/blob/fa9c60889616445e8cef85866096a4bcf8538bd7/_layouts/plugins/katello/documentation.html#L3

ehelms (approved):

PR #1068, opened 2018-04-29
drop obsolete OSes and Ruby versions from plugins doc

evgeni:


plugins/index.md
```diff
@@ -319,27 +297,15 @@ deb http://deb.theforeman.org/ plugins {{page.version}}
...
   <p>
     The naming of packages is as follows:
   </p>
 
   <ul>
-    <li>Foreman plugin packages have a <code>rubygem-foreman_</code> codefix.  Adapt any instructions to suit.</li>
+    <li>Packages for Foreman have a <code>tfm-rubygem-foreman_</code> codefix.</li>
     <li>Smart proxy packages have a <code>rubygem-smart_proxy_</code> codefix.</li>
-    <li>Hammer CLI packages have a <code>rubygem-hammer_cli_</code> codefix.</li>
+    <li>Hammer CLI packages have a <code>tfm-rubygem-hammer_cli_</code> codefix.</li>
```
mmoll: "codefix" should really be "prefix", I presume?

evgeni: maybe? is that a `s/pre/code/g` went wrong? :)

mmoll: Now that you say it, yes, probably. :)

evgeni: No, that was added like this in 7b9f7390b1 :face_with_head_bandage:

evgeni: Shall we do a `s/code/pre/g` on the whole file? Can't find the word `codefix` in any dictionary. Could also do in a separate PR (which technically would be cleaner).

mmoll: `<code>` is OK, no? So a second commit with `s/codefix/prefix/g` would be nice :heart:

evgeni: Yeah, `<code>` is fine, just wanted to tease you ;)
Now thinking about it, given this is markdown, we can also use backticks, which are more readable IMHO. I'll add two commits when back at the laptop.

evgeni: Mhh, nope, the rest is HTML to, so let's keep it that way for now. Pushed the `s/codefix/prefix/g` change.

mmoll:
merged, спасибо Женя!

PR #1073, opened 2018-05-02
Load-Balanced Smart Proxies with HAProxy

evgeni:


beav:
I made a couple of tiny updates, but ACK from me otherwise.

Someone less familiar with the topic should review this as well to ensure it is clear.

_posts/2018-05-02-load-balanced-smart-proxies-with-haproxy.md
```diff
@@ -0,0 +1,138 @@
...
+* two or more machines for the Smart Proxy cluster
+* a load balancer - you can use a machine with HAProxy if you don't already have a load balancer
+* a common name in the DNS that points to the load balancer (we'll be using `smartproxy.example.test`)
+
+## Installing the first Smart Proxy
+### Generating Katello certificates
+When calling `foreman-proxy-certs-generate`, we have to pass `--foreman-proxy-cname smartproxy.example.test` to add the load-balanced name as an accepted `subjectAltName` to the list of names in the generated certificates. This allows clients to connect to each member of the smartproxy cluster as `smartproxy.example.test`
+
+### Preparing `custom-hiera.yaml`
+We need to configure Pulp to serve redirects in the streamer code to the name of the load-balancer, but this option is not exposed in the installer. Thus edit `/etc/foreman-installer/custom-hiera.yaml` and add the following line at the end:
```
aptituz: As one who is not so much into the details of Pulp, I stumble across the term "streamer code". Maybe it would be helpful to eloborate on why that's necessary or point to appropriate docs.

_posts/2018-05-02-load-balanced-smart-proxies-with-haproxy.md
```diff
@@ -0,0 +1,138 @@
...
+This step is identical to the steps used to generate the certificates for the first smart proxy. Simply use `foreman-proxy-certs-generate` as described above.
+
+### Generating Puppet certificates
+On the first smart proxy, generate the Puppet certificates by running
+
+```shell
+puppet cert generate proxy02.example.test --dns_alt_names=smartproxy.example.test
+```
+
+Now copy `/etc/puppetlabs/puppet/ssl/*/proxy02.example.test.pem` and `/etc/puppetlabs/puppet/ssl/certs/ca.pem` to the second smart proxy and ensure the permissions are identical to the ones on the first host. The `puppet` user is created after installing the `puppetserver` package, so you might want to do that first. Also don't forget to check the SELinux contexts of the files you copied over (hint: `ls -lZ`) :)
```
aptituz: Probably worth to add a step for the installation of the puppetserver instead of pointing out that it should have happened at this point? :)

_posts/2018-05-02-load-balanced-smart-proxies-with-haproxy.md
```diff
@@ -0,0 +1,138 @@
...
+As only the first Smart Proxy will host a PuppetCA, we point the Puppet agent to it explicitly during the installation and also configure the Puppet certificate to include the `smartproxy.example.test` name:
+```
+--puppet-dns-alt-names smartproxy.example.test
+--puppet-ca-server proxy01.example.test
+--foreman-proxy-puppetca false
+--puppet-server-ca false
+```
+## Installing HAProxy
+
+As HAProxy will listen on ports not "owned" by it, SELinux will be default prevent that. For the sake of this post, we'll run the whole machine in permissive mode, but it's an open TODO item to investigate how to properly confine HAProxy. Also note that if you are using a different load balancer technology, you can use this configuration as reference.
```
aptituz: s/be default/by default/

aptituz: The requirement to run the machine in permissive mode is IMHO worth a note in the caveats section, because it might be something for people to take into consideration before setting this up.

_posts/2018-05-02-load-balanced-smart-proxies-with-haproxy.md
```diff
@@ -0,0 +1,138 @@
...
+
+[You can find an example `haproxy.cfg` in the `forklift` repository.](https://github.com/theforeman/forklift/blob/master/roles/haproxy/templates/haproxy.cfg.j2)
+
+## Configuring Clients
+
+As far as clients are concerned, they should only know about `smartproxy.example.test`, and should never need to connect directly to a smartproxy. That has to be configured for Puppet, goferd, and for RHSM/Yum.
+
+### Using katello-client-bootstrap
+
+[katello-client-bootstrap](https://github.com/Katello/katello-client-bootstrap) is designed to configure a machine to be used with an exiting Katello environment. It also can configure Puppet properly for us, and supports [setting a separate Puppet CA and Puppet CA port in current Git](https://github.com/Katello/katello-client-bootstrap/pull/250).
```
aptituz: s/also can/can also/ and I think the comma before and is not right

_posts/2018-05-02-load-balanced-smart-proxies-with-haproxy.md
```diff
@@ -0,0 +1,138 @@
...
+
+To attach clients to the load-balanced proxy, we just have to pass `--server smartproxy.example.test` and `--puppet-ca-port 8141`, `bootstrap.py` will take care of the rest.
+
+### Manual registration
+
+Instead of using `bootstrap.py`, the client can also be registered manually. However, we recommend using `bootstrap.py` since it handles these extra manual steps for you.
+
+After installing `katello-ca-consumer` RPM from `http://smartproxy.example.test/pub/katello-ca-consumer-latest.noarch.rpm` `/etc/rhsm/rhsm.conf` and `/etc/yum.repos.d/redhat.repo` will contain the hostname of whichever proxy served the initial request and not `smartproxy.example.net` as we'd like it. Thus, when registering the system with `subscription-manager`, we have to pass `--serverurl=https://smartproxy.example.test:8443/rhsm --baseurl=https://smartproxy.example.test/pulp/repos` to update those two config files.
+
+Also Puppet needs to be configured to talk to the split-out PuppetCA, so the `puppet.conf` needs the following lines:
```
aptituz: Wait ... to refer to the "split-out PuppetCA" is kind of irritating me. Is it relevant, that its split out at this point? 

I guess not, because the haproxy configuration handles forwarding to the right server, so I'd write that PuppetCA has to be configured to use the shared hostname, but not explicitly refer that detail of the balancer setup.

Besides that just a wording suggestion: s/Also/Additionally/ ... but "Also" is probably okay.

_posts/2018-05-02-load-balanced-smart-proxies-with-haproxy.md
```diff
@@ -0,0 +1,138 @@
...
+server = smartproxy.example.test
+ca_server = smartproxy.example.test
+ca_port = 8141
+```
+
+## Forklift
+As usual, there is [support for the loadbalanced scenario in `forklift`](https://github.com/theforeman/forklift/blob/master/pipelines/pipeline_katello_lbproxy.yml).
+
+## Help! Something doesn't work!
+The load balancer adds additional complexity to a Foreman/Katello installation. If a client is having trouble using one of the services provided by the smart proxies, a typical first debugging step is to point the client directly to one of the smart proxies via a `/etc/hosts` entry, which bypasses the load balancer. If the issue persists, then you know that the issue is with that smart proxy. Each smart proxy can be checked individually for functionality with this method.
```
aptituz: Is it worth nothing that pointing the client directly to one of the smart proxies will be subject to restrictions if the smart proxy they point it too is not the first one?

aptituz (commented):
Hi,

I added some review comments as one who is not that much into the details. Have not actually *tried* if the instructions work, though, but from a technical point of view that seems reasonable to me.

Best Regards,
Patrick

_posts/2018-05-02-load-balanced-smart-proxies-with-haproxy.md
evgeni: Well, it needs to talk to the dedicated port, as that is what is forwarded to proxy01 only. I rephrased that sentence, does it make more sense now?

_posts/2018-05-02-load-balanced-smart-proxies-with-haproxy.md
evgeni: @beav this paragraph was initially written by you, so what do you think?

I'd add something like "As the PuppetCA functionality is only available on the first smart proxy, it can only be tested when Puppet is configured to use port 8140 on the first smart proxy, or port 8141 on the load-balancer for PuppetCA requests."

_posts/2018-05-02-load-balanced-smart-proxies-with-haproxy.md
aptituz: Fair enough, I somehow missed (to keep in mind) the fact that PuppetCA is running on a different port. I would still suggest a rephrasing. Maybe something like "Since our PuppetCA is reachable on a dedicated port, we additionally need to adjust the puppet.conf to use that port and the shared hostname for certificate requests."

_posts/2018-05-02-load-balanced-smart-proxies-with-haproxy.md
aptituz: For me that addition sounds reasonable.

_posts/2018-05-02-load-balanced-smart-proxies-with-haproxy.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Requirements
+* a working Foreman/Katello installation (we don't need to do any changes here)
+* two or more machines for the Smart Proxy cluster
+* a load balancer - you can use a machine with HAProxy if you don't already have a load balancer
+* a common name in the DNS that points to the load balancer (we'll be using `smartproxy.example.test`)
+
+## Installing the first Smart Proxy
+### Generating Katello certificates
+When calling `foreman-proxy-certs-generate`, we have to pass `--foreman-proxy-cname smartproxy.example.test` to add the load-balanced name as an accepted `subjectAltName` to the list of names in the generated certificates. This allows clients to connect to each member of the smartproxy cluster as `smartproxy.example.test`
```
ekohl: I always prefer example.{com,net,org} because they're mentioned in an RFC for this purpose.

_posts/2018-05-02-load-balanced-smart-proxies-with-haproxy.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Preparing `custom-hiera.yaml`
+We need to configure Pulp to serve redirects in the streamer code to the name of the load-balancer, but this option is not exposed in the Installer. Thus edit `/etc/foreman-installer/custom-hiera.yaml` and add the following line at the end:
+
+```yaml
+pulp::lazy_redirect_host: smartproxy.example.test
+```
+### Running the installer
+As only the first Smart Proxy will host a PuppetCA, we configure that explicitly during the installation and also configure the Puppet certificate to include the `smartproxy.example.test` name:
+```
```
ekohl: Would it help to add `[root@smartproxy01] # foreman-installer` as the first line? Perhaps also backslashes so it can be copy-pasted?

_posts/2018-05-02-load-balanced-smart-proxies-with-haproxy.md
```diff
@@ -0,0 +1,138 @@
...
+This step is identical to the steps used to generate the certificates for the first smart proxy. Simply use `foreman-proxy-certs-generate` as described above.
+
+### Generating Puppet certificates
+On the first smart proxy, generate the Puppet certificates by running
+
+```shell
+puppet cert generate proxy02.example.test --dns_alt_names=smartproxy.example.test
+```
+
+Now copy `/etc/puppetlabs/puppet/ssl/*/proxy02.example.test.pem` and `/etc/puppetlabs/puppet/ssl/certs/ca.pem` to the second smart proxy and ensure the permissions are identical to the ones on the first host. The `puppet` user is created after installing the `puppetserver` package, so you might want to do that first. Also don't forget to check the SELinux contexts of the files you copied over (hint: `ls -lZ`) :)
```
ekohl: Should proxy02 point to proxy01 as a CA and have its own CA functionality disabled? I think you can also put the alt name in puppet.conf and then just request as normal. It does mean you need to sign the cert on proxy01, but then you don't have to transfer files and worry about permissions. Probably also works better with renewals.

_posts/2018-05-02-load-balanced-smart-proxies-with-haproxy.md
evgeni: Yes and no, as that's not the complete installer call, but we could do
```
[root@smartproxy01] # foreman-installer \
…
<the above with backslashes>
```

_posts/2018-05-02-load-balanced-smart-proxies-with-haproxy.md
evgeni: Aye, updated to include something very similar to your line :)

_posts/2018-05-02-load-balanced-smart-proxies-with-haproxy.md
```diff
@@ -0,0 +1,151 @@
...
+### Generating Puppet certificates
+On the first smart proxy, generate the Puppet certificates by running
+
+```shell
+puppet cert generate proxy02.example.net --dns_alt_names=smartproxy.example.net
+```
+
+Install the `puppetserver` package on the second smart proxy, to create the `puppet` user and the directory structure in `/etc/puppetlabs`.
+
+Now copy `/etc/puppetlabs/puppet/ssl/*/proxy02.example.net.pem` and `/etc/puppetlabs/puppet/ssl/certs/ca.pem` to the second smart proxy and ensure ownership and permissions are identical to those the first host. Also don't forget to check the SELinux contexts of the files you copied over (hint: `ls -lZ`) :)
```
ekohl, line 78: I realize now you're already passing in the right vars in the installer. Not sure if it works, but I think you can add `proxy02.example.net` to `/etc/puppetlabs/puppet/autosign.conf` on `proxy01.example.net` and I think the installer should get the cert from its CA. Not sure though.

_posts/2018-05-02-load-balanced-smart-proxies-with-haproxy.md
```diff
@@ -0,0 +1,151 @@
...
+--foreman-proxy-puppetca false \
+--puppet-server-ca false
+```
+## Installing HAProxy
+
+Please note that if you are using a different load balancer technology, you can use this configuration as reference.
+
+Configure SELinux to run in `permissive` mode.
+
+The HAProxy setup itself is super simple, just load-balance a bunch of ports in TCP mode round-robin across all the Smart Proxies. TCP mode is required because the load balancer will not be doing TLS termination. There are just two exceptions to that: port 443 should use `source` not `roundrobin` to not confuse Yum, and the additional port for the PuppetCA (8141) is forwarded only to the first smart proxy.
```
ekohl: I read somewhere that you should avoid describing things as simple. It can make beginners feel stupid when they think it's not simple. Maybe it's better to use `straight forward`?

_posts/2018-05-02-load-balanced-smart-proxies-with-haproxy.md
```diff
@@ -0,0 +1,151 @@
...
+| Anaconda   | 8000 | TCP  | `roundrobin` | port 8000 on all proxies|
+| HTTPS      | 443  | TCP  | `source`     | port 443 on all proxies |
+| RHSM       | 8443 | TCP  | `roundrobin` | port 8443 on all proxies |
+| AMQP       | 5647 | TCP  | `roundrobin` | port 5647 on all proxies |
+| Puppet     | 8140 | TCP  | `roundrobin` | port 8140 on all proxies |
+| PuppetCA   | 8141 | TCP  | `roundrobin` | port 8140 on the first proxy |
+| SmartProxy | 9090 | TCP  | `roundrobin` | port 9090 on all proxies |
+| Docker     | 5000 | TCP  | `roundrobin` | port 5000 on all proxies |
+
+[You can find an example `haproxy.cfg` in the `forklift` repository.](https://github.com/theforeman/forklift/blob/master/roles/haproxy/templates/haproxy.cfg.j2)
```
ekohl, line 115: I wonder if linking to a persistent location (i.e. the file in a specific commit) is better (so the link won't die) or worse (won't receive updates).

_posts/2018-05-02-load-balanced-smart-proxies-with-haproxy.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Configuring Clients
+
+As far as clients are concerned, they should only know about `smartproxy.example.net`, and should never need to connect directly to a smartproxy. That has to be configured for Puppet, goferd, and for RHSM/Yum.
+
+### Using katello-client-bootstrap
+
+[katello-client-bootstrap](https://github.com/Katello/katello-client-bootstrap) is designed to configure a machine to be used with an exiting Katello environment. It can also configure Puppet properly for us and supports [setting a separate Puppet CA and Puppet CA port in current Git](https://github.com/Katello/katello-client-bootstrap/pull/250).
+
+To attach clients to the load-balanced proxy, we just have to pass `--server smartproxy.example.net` and `--puppet-ca-port 8141`, `bootstrap.py` will take care of the rest.
```
ekohl, line 125: Did you mean `--server` instead of `--puppet-server` here?

_posts/2018-05-02-load-balanced-smart-proxies-with-haproxy.md
```diff
@@ -0,0 +1,151 @@
...
+Instead of using `bootstrap.py`, the client can also be registered manually. However, we recommend using `bootstrap.py` since it handles these extra manual steps for you.
+
+After installing `katello-ca-consumer` RPM from `http://smartproxy.example.net/pub/katello-ca-consumer-latest.noarch.rpm` `/etc/rhsm/rhsm.conf` and `/etc/yum.repos.d/redhat.repo` will contain the hostname of whichever proxy served the initial request and not `smartproxy.example.net` as we'd like it. Thus, when registering the system with `subscription-manager`, we have to pass `--serverurl=https://smartproxy.example.net:8443/rhsm --baseurl=https://smartproxy.example.net/pulp/repos` to update those two config files.
+
+Since PuppetCA is reachable on a dedicated port, all Puppet agents need to be configured to talk to `smartproxy.example.net` on port 8140 for Puppet requests and port 8141 for PuppetCA requests.
+This can be achieved by modifying `puppet.conf` to include the following lines:
+
+```
+server = smartproxy.example.net
+ca_server = smartproxy.example.net
```
ekohl, line 138: Technically I think this line is redundant but doesn't hurt either.

ekohl (commented):
IIRC haproxy also had a status page where you could see the health of the backends. That might be something useful to point to and verify it's balancing.

_posts/2018-05-02-load-balanced-smart-proxies-with-haproxy.md
evgeni: I tried that while working on the forklift changes, and it somehow did not work. No idea why, but proxy02 never did a CA request to proxy01. Might re-test it at some point, but the above works.

_posts/2018-05-02-load-balanced-smart-proxies-with-haproxy.md
evgeni: `--server` is correct, as that will use that name for all things registration (API, katello-ca-consumer RPM, puppet, puppetCA), whereas `--puppet-server` is only to set a puppet server that is *different* from the one passed in `--server`

_posts/2018-05-02-load-balanced-smart-proxies-with-haproxy.md
ekohl: I think it does make sense, but we could have a look at supporting something like it. For now this works indeed :)

_posts/2018-05-02-load-balanced-smart-proxies-with-haproxy.md
evgeni: Yeah, but I kinda preferred being explicit here.

evgeni:
I think we have addressed all the comments so far.

ekohl (approved):

PR #1078, opened 2018-05-07
Update supported Puppet versions

tbrisker:
Following the discussion in https://community.theforeman.org/t/reduce-supported-puppet-versions/8948

### DIFF
--- a/_includes/manuals/1.18/3.1.3_puppet_versions.md
+++ b/_includes/manuals/1.18/3.1.3_puppet_versions.md
@@ -1,4 +1,5 @@
 Foreman integrates with Puppet and Facter in a few places, but generally using a recent, stable version will be fine.  The exact versions of Puppet, Puppet Server and Facter that Foreman is compatible with are listed below.
+Foreman 1.18 is the last version that will support Puppet < 4.0. Report/fact processing and ENC with older versions may work in Foreman 1.19, but this workflow will not be tested.
 
 <table class="table table-bordered table-condensed">
   <tr>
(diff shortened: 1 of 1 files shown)


_includes/manuals/1.18/3.1.3_puppet_versions.md
```diff
@@ -9,42 +10,14 @@ Foreman integrates with Puppet and Facter in a few places, but generally using a
...
-  <tr>
-    <td>3.0.x - 3.5.x</td>
-    <td>Not supported</td>
-    <td>Supported</td>
-    <td>Supported</td>
-    <td>Supported</td>
-  </tr>
-  <tr>
-    <td>3.6.x - 3.8.x</td>
+    <td>3.x</td>
```
ekohl: This implies we support any puppet version. I'd suggest we make this 3.8.x and mark anything < 3.8 as untested.

_includes/manuals/1.18/3.1.3_puppet_versions.md
```diff
@@ -1,4 +1,5 @@
 Foreman integrates with Puppet and Facter in a few places, but generally using a recent, stable version will be fine.  The exact versions of Puppet, Puppet Server and Facter that Foreman is compatible with are listed below.
+Foreman 1.18 is the last version that will support Puppet < 4.0.
```
timogoebel: 👎 
Can't we make this `< 3.8`?

GregSutcliffe: Even Puppet 4 will be end of life later this year. Can we stick to supporting just one out-of-date version? :P

ekohl: Given our installer requires puppet 4 since a few versions, I'm inclined to say we no longer fully support `< 4.0`. We have partial support for older versions for now.

GregSutcliffe: Can we clarify the doc then? If we're leaving in the report importers etc then we should probably say that the functionality will remain but is unsupported / untested on < 3.8

tbrisker:
updated to address the comments, @ekohl @GregSutcliffe @timogoebel what do you think about this?

GregSutcliffe:
:+1: reads well. We'll need to ensure that warning goes to the the release highlights too...

GregSutcliffe (approved):

timogoebel (approved):
Works for me. :-)

ekohl (approved):
I think it's ok.

mmoll:
merged, thanks @tbrisker!

PR #1079, opened 2018-05-08
Fixes #23542 - fix up katello client docs

jlsherrill:


stbenjam (changes requested):
Fedora docs are broken on Katello 3.6 and 3.5, could we fix those too?

jlsherrill:
@stbenjam updated!

stbenjam (approved):

PR #1083, opened 2018-05-12
Clarify and update Contribute setup

lonnv:
Updated the localhost port.
And moved up the instruction to copy the `config/datatabase.yml.example` file for clarification.


tbrisker (approved):
Thanks @lonnv !

PR #1088, opened 2018-05-17
Fix 1.17 manual

juliantodt:
The Foreman 1.17 manual is currently broken because of a not starting code-box in 4.4.4, see:
foreman org missing5
This causes paragraphs 4.4.5-5.1 to not be displayed correctly (and also missing in the navigation on the left).
This PR fixes this and lets the manual be displayed correctly again.
theforeman org missing5 fixed


leewaa:
Nice catch 😄

mmoll:
merged, danke @juliantodt!

PR #1091, opened 2018-05-22
Katello: extra flags for changelog generator

jturel:
@ekohl FYI ... adding flags mentioned here: https://github.com/theforeman/puppet-pulp/pull/316


beav (approved):

ekohl (approved):
I can't merge but this is correct.

PR #1102, opened 2018-06-01
Add Katello with remote database docs

johnpmitsch:
Katello can be installed with remote databases,
this documentation details that process. This is moved
from the
[foreman-maintain wiki](https://github.com/theforeman/foreman_maintain/wiki)
where it temporarily lived.

### DIFF
--- a/plugins/katello/nightly/user_guide/remote_databases/index.md
+++ b/plugins/katello/nightly/user_guide/remote_databases/index.md
@@ -99,8 +99,8 @@ Warning: This is just minimal testing setup which is not suitable for production
 Assume our Mongo server has hostname `mongo.example.com`.
 Install and enable Mongo server
 ``` 
-yum -y localinstall https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
-yum install -y mongodb mongodb-server
+yum install -y centos-release-scl
+yum install -y rh-mongodb34-syspaths
 ```
 Enable authentication in `/etc/mongod.conf`
 ```
--- a/plugins/katello/nightly/user_guide/remote_databases/index.md
+++ b/plugins/katello/nightly/user_guide/remote_databases/index.md
@@ -1,0 +1,270 @@
+---
+layout: plugins/katello/documentation
+title: Use Katello with Remote Databases
+version: nightly
+---
+
+# Setup Katello with remote databases
+
+Katello can be installed with remote databases for both postgresql and mongo.
+
+## High level
+There are two ways to deploy Katello with remote databases:
+
+1. [Fresh install](plugins/katello/{{ page.version }}/user_guide/remote_databases/index.html#fresh-install)
+    * prepare Postgres server with databases for Foreman and Candlepin and dedicated users owning them
+    * prepare Mongo DB with user owning the pulp_database
+    * prepare box where the Katello will be installed and make sure the databases are accessible from the box
+    * run foreman-installer with right parameters pointing to the databases
+
+2. [Migration of existing Katello installation](plugins/katello/{{ page.version }}/user_guide/remote_databases/index.html#migration-of-existing-katello)
+    * prepare Postgres server with databases for Foreman a Candlepin and dedicated users owning them
+    * prepare Mongo DB with user owning the pulp_databse
+    * make sure the databases are accessible from the box where Katello is installed
+    * shut down the services except the dbs you want to move (mongod, postgresql)
+    * dump the DBs
+    * restore the DBs on remote servers
+    * run foreman-installer with right parameters pointing to the databases. It re-configures the databases and start all the services with new DB locations
+
+## Prepare remote Postgres
+GOAL: To use remote Postgres database with Katello we have to:
+* be able to access the databases from katello box
+* the database user we use to connect to the database needs to own the database, i.e. it can create, alter and delete the tables, indexes and constraints. Note it is not required to be able to create the database itself.
+
+### Install Postgres
+Warning: This is just minimal testing setup which is not suitable for production.
+
+Assume our postgres server has hostname `postgres.example.com`.
+
+```
+  yum -y localinstall https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+  yum install postgresql-server postgresql-contrib
+  postgresql-setup initdb
+  systemctl start postgresql
+  systemctl enable postgresql
+```
+
+Now we need to make Postgres listen to inbound connections, please adjust these parameters to your own networking and security requirements. 
+
+Edit `/var/lib/pgsql/data/postgresql.conf`. Uncomment `listen_address` and modify its value to look like: 
+```
+  listen_address = "*"
+```
+
+The next step we need to take is to add a proper client authentication for remote client to our postgres server. To achieve the same, edit `/var/lib/pgsql/data/pg_hba.conf`.
+Append the following line at the end of the file
+```
+  host          all          all          <katello.example.com ip>/24              md5
+```
+
+Now restart the postgres service for changes to take effect
+```
+  systemctl restart postgresql
+```
+
+### Create the databases
+Switch the user role to postgres and start postgres client
+```
+  su - postgres -c psql
+```
+
+Once inside the client, we need to create two databases and dedicated roles, one for foreman and one for candlepin
+```
+  CREATE USER "foreman" WITH PASSWORD '<FOREMAN_PASSWORD>';
+  CREATE USER "candlepin" WITH PASSWORD '<CANDLEPIN_PASSWORD>';
+  CREATE DATABASE foreman OWNER foreman;
+  CREATE DATABASE candlepin OWNER candlepin;
+```
+
+### Test it works
+From `katello.example.com` test the DB is accessible:
+```
+PGPASSWORD='<FOREMAN_PASSWORD>' psql -h postgres.example.com  -p 5432 -U foreman -d foreman -c "SELECT 1 as ping"
+```
+```
+PGPASSWORD='<CANDLEPIN_PASSWORD>' psql -h postgres.example.com  -p 5432 -U candlepin -d candlepin -c "SELECT 1 as ping"
+```
+If there are no errors we are done with database preparation. 
+
+## Prepare remote Mongo
+GOAL: To use remote Mongo database with Katello we have to:
+* be able to access the databases from katello box
+* the database user we use to connect to the database needs to own the database
+
+### Install Mongo DB
+Warning: This is just minimal testing setup which is not suitable for production.
+
+Assume our Mongo server has hostname `mongo.example.com`.
+Install and enable Mongo server
+``` 
+  yum -y localinstall https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+  yum install -y mongodb mongodb-server
+```
+Enable authentication in `/etc/mongod.conf`
+```
+auth=true
+```
+
+Enable and start the service
+```
+  systemctl start mongod
+  systemctl enable mongod
+```
+
+### Create Pulp user and database 
+```
+  mongo admin -u admin -p admin --eval "db.createUser({user:'pulp',pwd:'<PULP_PASSWORD>',roles:[{role:'dbOwner', db:'pulp_database'},{ role: 'readWrite', db: 'pulp_database'}]})"
+
+```
+
+### Test it works
+From `katello.example.com` test the mongo DB is accessible:
+```
+  mongo --host mongo.example.com -u pulp -p <PULP_PASSWORD> --port 27017 --eval 'ping:1' pulp_database
+```
+If there are no errors we are done with database preparation. 
+
+## Fresh install
+
+### Install katello package
+We assume the box where the Katello server will be installed has hostname `katello.example.com`.
+
+Follow the [documentation](plugins/katello/{{ page.version }}/installation/index.html) to install the `katello` package and **do not run foreman-installer.** We need to use the remote database flags with the installer. Use the following steps once the katello rpm is installed.
+
+
+### Prepare remote databases
+Follow the instructions to [prepare remote mongo](plugins/katello/{{ page.version }}/user_guide/remote_databases/index.html#prepare-remote-mongo) and [prepare remote postgres](plugins/katello/{{ page.version }}/user_guide/remote_databases/index.html#prepare-remote-postgres) to make the remote database servers ready for installation.
+
+### Run the installer
+To install and configure Katello we just need to run 
+```
+  foreman-installer --scenario katello \
+    --foreman-db-host postgres.example.com \ 
+    --foreman-db-password <FOREMAN_PASSWORD> \
+    --foreman-db-database foreman \
+    --katello-candlepin-db-host postgres.example.com \
+    --katello-candlepin-db-name candlepin \
+    --katello-candlepin-db-password <CANDLEPN_PASSWORD> \
+    --katello-candlepin-manage-db false \
+    --katello-pulp-db-username pulp \ 
+    --katello-pulp-db-password <PULP_PASSWORD> \ 
+    --katello-pulp-db-seeds “mongo.example.com:27017” \
+    --katello-pulp-db-name pulp_database
+```
+
+Note: for more related options and tips on SSL configuration see [Full list of options](plugins/katello/{{ page.version }}/user_guide/remote_databases/index.html#full-list-of-remote-database-related-options-in-the-installer)
+
+## Migration of existing Katello
+We assume that Katello was installed and is running on `katello.example.com`.
+
+### Prepare remote databases
+Follow the instructions to [prepare remote mongo](plugins/katello/{{ page.version }}/user_guide/remote_databases/index.html#prepare-remote-mongo) and [prepare remote postgres](plugins/katello/{{ page.version }}/user_guide/remote_databases/index.html#prepare-remote-postgres) to make the remote database servers ready for migration.
+
+### Stop the Katello server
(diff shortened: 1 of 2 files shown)


plugins/katello/nightly/user_guide/remote_databases/index.md
```diff
@@ -0,0 +1,270 @@
...
+### Install Postgres
+Warning: This is just minimal testing setup which is not suitable for production.
+
+Assume our postgres server has hostname `postgres.example.com`.
+
+```
+  yum -y localinstall https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+  yum install postgresql-server postgresql-contrib
+  postgresql-setup initdb
+  systemctl start postgresql
```
ekohl: This can be removed by using `systemctl enable --now postgresql`

plugins/katello/nightly/user_guide/remote_databases/index.md
```diff
@@ -0,0 +1,270 @@
...
+* be able to access the databases from katello box
+* the database user we use to connect to the database needs to own the database, i.e. it can create, alter and delete the tables, indexes and constraints. Note it is not required to be able to create the database itself.
+
+### Install Postgres
+Warning: This is just minimal testing setup which is not suitable for production.
+
+Assume our postgres server has hostname `postgres.example.com`.
+
+```
+  yum -y localinstall https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
```
ekohl: Is this needed? Both `postgresql-server` and `postgresql-contrib` are in base.

plugins/katello/nightly/user_guide/remote_databases/index.md
```diff
@@ -0,0 +1,270 @@
...
+```
+
+### Create the databases
+Switch the user role to postgres and start postgres client
+```
+  su - postgres -c psql
+```
+
+Once inside the client, we need to create two databases and dedicated roles, one for foreman and one for candlepin
+```
```
ekohl: I think you can add sql here for syntax highlighting.

plugins/katello/nightly/user_guide/remote_databases/index.md
```diff
@@ -0,0 +1,270 @@
...
+* the database user we use to connect to the database needs to own the database, i.e. it can create, alter and delete the tables, indexes and constraints. Note it is not required to be able to create the database itself.
+
+### Install Postgres
+Warning: This is just minimal testing setup which is not suitable for production.
+
+Assume our postgres server has hostname `postgres.example.com`.
+
+```
+  yum -y localinstall https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+  yum install postgresql-server postgresql-contrib
```
ekohl: You can remove the indenting when using the backticks. This gives better results when copy-pasting to a shell.

plugins/katello/nightly/user_guide/remote_databases/index.md
```diff
@@ -0,0 +1,270 @@
...
+  yum install -y mongodb mongodb-server
+```
+Enable authentication in `/etc/mongod.conf`
+```
+auth=true
+```
+
+Enable and start the service
+```
+  systemctl start mongod
```
ekohl: Can be omitted by using `enable --now` as well.

plugins/katello/nightly/user_guide/remote_databases/index.md
```diff
@@ -0,0 +1,270 @@
...
+### Prepare remote databases
+Follow the instructions to [prepare remote mongo](plugins/katello/{{ page.version }}/user_guide/remote_databases/index.html#prepare-remote-mongo) and [prepare remote postgres](plugins/katello/{{ page.version }}/user_guide/remote_databases/index.html#prepare-remote-postgres) to make the remote database servers ready for installation.
+
+### Run the installer
+To install and configure Katello we just need to run 
+```
+  foreman-installer --scenario katello \
+    --foreman-db-host postgres.example.com \ 
+    --foreman-db-password <FOREMAN_PASSWORD> \
+    --foreman-db-database foreman \
```
ekohl: This is the default so it could be omitted. OTOH it can be good to be explicit.

plugins/katello/nightly/user_guide/remote_databases/index.md
```diff
@@ -0,0 +1,270 @@
...
+Follow the [documentation](plugins/katello/{{ page.version }}/installation/index.html) to install the `katello` package and **do not run foreman-installer.** We need to use the remote database flags with the installer. Use the following steps once the katello rpm is installed.
+
+
+### Prepare remote databases
+Follow the instructions to [prepare remote mongo](plugins/katello/{{ page.version }}/user_guide/remote_databases/index.html#prepare-remote-mongo) and [prepare remote postgres](plugins/katello/{{ page.version }}/user_guide/remote_databases/index.html#prepare-remote-postgres) to make the remote database servers ready for installation.
+
+### Run the installer
+To install and configure Katello we just need to run 
+```
+  foreman-installer --scenario katello \
```
ekohl: You probably also want `--foreman-db-manage false` so it doesn't install a local postgresql server. The same goes for `--katello-pulp-manage-db false`.

plugins/katello/nightly/user_guide/remote_databases/index.md
```diff
@@ -0,0 +1,270 @@
...
+    --foreman-db-host postgres.example.com \ 
+    --foreman-db-password <FOREMAN_PASSWORD> \
+    --foreman-db-database foreman \
+    --katello-candlepin-db-host postgres.example.com \
+    --katello-candlepin-db-name candlepin \
+    --katello-candlepin-db-password <CANDLEPN_PASSWORD> \
```
ekohl: CANDLEPN -> CANDLEPIN

plugins/katello/nightly/user_guide/remote_databases/index.md
```diff
@@ -0,0 +1,270 @@
...
+We assume that Katello was installed and is running on `katello.example.com`.
+
+### Prepare remote databases
+Follow the instructions to [prepare remote mongo](plugins/katello/{{ page.version }}/user_guide/remote_databases/index.html#prepare-remote-mongo) and [prepare remote postgres](plugins/katello/{{ page.version }}/user_guide/remote_databases/index.html#prepare-remote-postgres) to make the remote database servers ready for migration.
+
+### Stop the Katello server
+Stop the Katello related services to minimize risk of the data changes during the migration
+```
+  katello-service stop
+  systemctl start postgresql
```
ekohl: You can also use `systemctl start postgresql mongodb` so it happens in parallel or `katello-service stop --exclude postgresql,mongod` to prevent them from stopping in the first place.

plugins/katello/nightly/user_guide/remote_databases/index.md
```diff
@@ -0,0 +1,270 @@
...
+You can restore the SQL dumps to the remote databases from the katello system.
+
+```
+PGPASSWORD='<FOREMAN_PASSWORD>' pg_restore -h postgres.example.com -U foreman -d foreman < /tmp/migration_backup/foreman.dump
+PGPASSWORD='<CANDLEPIN_PASSWORD>' pg_restore -h postgres.example.com -U candlepin -d candlepin < /tmp/migration_backup/candlepin.dump
+mongorestore --host mongo.example.com --db pulp_database --username pulp --password <PULP_PASSWORD> /tmp/migration_backup/mongo_dump
+```
+Now the copy of the local database is also at the remote locations.
+
+### Update the configuration
```
ekohl, line 191: I don't really see why you rerun the installer now with additional options rather than doing it right the first time.

plugins/katello/nightly/user_guide/remote_databases/index.md
johnpmitsch: thanks, will keep them as separate steps in case they are stopped for some reason

plugins/katello/nightly/user_guide/remote_databases/index.md
johnpmitsch: This is for an existing installation of Katello, this doc is split into two parts, new installs and existing installs

johnpmitsch:
@ekohl appreciate the suggestions, I updated in a separate commit

plugins/katello/nightly/user_guide/remote_databases/index.md
ekohl: It might be good to write a few words about stopping databases on the foreman host when doing an upgrade. I'm also quite sure the installer call is now the same so you can also restructure the manual a bit to do it like this:
* Install postgres.example.com
* Stop services on foreman.example.com (skip if fresh install)
* Dump and restore postgres (skip if fresh install)
* Install mongo.example.com
* Dump & restore mongodb (skip if fresh install)
* Reconfigure foreman.example.com
* Stop DBs on foreman.example.com (skip if fresh install)

You can mention users can proceed with installing mongo while the postgres dump & restore is running.

However, I'll leave this up to you.

plugins/katello/nightly/user_guide/remote_databases/index.md
```diff
@@ -0,0 +1,270 @@
+---
+layout: plugins/katello/documentation
+title: Use Katello with Remote Databases
+version: nightly
+---
+
+# Setup Katello with remote databases
+
+Katello can be installed with remote databases for both postgresql and mongo.
```
ehelms: We should note for Mongo in a few places this only applies to the server and not external smart proxies.

johnpmitsch: Can you elaborate? I'm not sure I understand what you mean

ekohl: I think capsules could use a remote mongo (not 100% sure), but users shouldn't share their mongo db between the main server and their capsules I think.

ehelms: In conversations I've been involved, we'd been opting to support external MongoDB for the main server only. If that's changed, that's fine. Has that been tested if we want to support that?

plugins/katello/nightly/user_guide/remote_databases/index.md
```diff
@@ -0,0 +1,268 @@
...
+
+## High level
+There are two ways to deploy Katello with remote databases:
+
+1. [Fresh install](plugins/katello/{{ page.version }}/user_guide/remote_databases/index.html#fresh-install)
```
ehelms: Should we note here that it is not required to externalize both databases? They can be externalized separately, or only one of them.

plugins/katello/nightly/user_guide/remote_databases/index.md
johnpmitsch: I can put a disclaimer at the top saying these instructions are for the main katello server only

plugins/katello/nightly/user_guide/remote_databases/index.md
```diff
@@ -0,0 +1,268 @@
...
+* the database user we use to connect to the database needs to own the database
+
+### Install Mongo DB
+Warning: This is just minimal testing setup which is not suitable for production.
+
+Assume our Mongo server has hostname `mongo.example.com`.
+Install and enable Mongo server
+``` 
+yum -y localinstall https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+yum install -y mongodb mongodb-server
```
ehelms: We should avoid encouraging installation of EPEL MongoDB and use RHSCL examples since that is what we support internally.

plugins/katello/nightly/user_guide/remote_databases/index.md
```diff
@@ -0,0 +1,268 @@
+---
+layout: plugins/katello/documentation
+title: Use Katello with Remote Databases
+version: nightly
+---
+
+# Setup Katello with remote databases
```
ehelms: General info here would be good about how you can externalize 1 or both databases. And that both databases can live on the same remote host but we recommend splitting them given MongoDB likes to own memory.

plugins/katello/nightly/user_guide/remote_databases/index.md
```diff
@@ -0,0 +1,268 @@
...
+Stop the Katello related services to minimize risk of the data changes during the migration
+```
+katello-service stop
+systemctl start postgresql mongod
+```
+
+### Dump databases
+Dump the local databases
+```
+foreman-maintain backup online --skip-pulp-content --preserve-directory -y /tmp/migration_backup
```
ehelms, line 178: Should this not be an offline backup since we stopped all services?

plugins/katello/nightly/user_guide/remote_databases/index.md
```diff
@@ -0,0 +1,268 @@
...
+  --katello-pulp-db-username pulp \ 
+  --katello-pulp-db-password <PULP_PASSWORD> \ 
+  --katello-pulp-db-seeds “mongo.example.com:27017” \
+  --katello-pulp-db-name pulp_database
+  --katello-pulp-manage-db false
+```
+
+Note: for more related options and tips on SSL configuration see [Full list of options](plugins/katello/{{ page.version }}/user_guide/remote_databases/index.html#full-list-of-remote-database-related-options-in-the-installer)
+
+## Migration of existing Katello
```
ehelms: I think it would be good to warn users here that this requires a maintenance window, taking all services offline and depending on database size can mean a large maintenance window without services.

plugins/katello/nightly/user_guide/remote_databases/index.md
```diff
@@ -0,0 +1,268 @@
...
+  --katello-candlepin-db-name candlepin \
+  --katello-candlepin-db-password <CANDLEPN_PASSWORD> \
+  --katello-candlepin-manage-db false \
+  --katello-pulp-db-username pulp \ 
+  --katello-pulp-db-password <PULP_PASSWORD> \ 
+  --katello-pulp-db-seeds “mongo.example.com:27017” \
+  --katello-pulp-db-name pulp_database \
+  --katello-pulp-manage-db false
+```
+Installer also starts the services and everything should be up and ready at this point.
```
ehelms: Maybe say 'starts all services except for the databases`

ehelms:
This looks like a good first pass as this is a large and somewhat complicated procedure to document. I generally would prefer if we opted to talk about the Foreman server, or at least Foreman server with Katello to enable our move away from treating Katello as if it's own distribution.

plugins/katello/nightly/user_guide/remote_databases/index.md
johnpmitsch: How does one do that?

plugins/katello/nightly/user_guide/remote_databases/index.md
johnpmitsch: We need the SQL dumps from an online backup (at least that is how we've been testing this process). Foreman-maintain will start the necessary services needed for the online backup (and they are started in the previous step)

johnpmitsch:
@ehelms updated according to your comments in a separate commit

> This looks like a good first pass as this is a large and somewhat complicated procedure to document. I generally would prefer if we opted to talk about the Foreman server, or at least Foreman server with Katello to enable our move away from treating Katello as if it's own distribution.

I do agree. We documented this as the full Katello process since that has the most steps. I mostly just copied from the notes we kept in the [foreman-maintain wiki](https://github.com/theforeman/foreman_maintain/wiki/Setup-Katello-with-remote-databases). Splitting this up and testing in foreman will be more work and prolong getting this in the documentation. I would like to get this process in upstream docs in one shape or another since I think its lacking, and then split up as needed.

plugins/katello/nightly/user_guide/remote_databases/index.md
ehelms: Install centos-release-scl and then rh-mongodb34-syspaths

plugins/katello/nightly/user_guide/remote_databases/index.md
ehelms: I guess when I think of an online backup I think of all primary services running. If the webserver is down but databases are up, I don't consider that online given the application is not running and no modifications can be made to the database.

ehelms:
I think you misunderstood me. I am not asking to update these talks to talk about doing vanilla Foreman. I am asking to talk about Foreman as the server/application/main-entity. I want us all to talk about having Foreman or Foreman with X/Y/Z capabilities present. Does that make more sense?

johnpmitsch:
@ehelms I think so, do you mean instead of saying "Katello", we should say "Foreman server with Katello installed" or just "Foreman server" as to not refer to Katello as a separate application?

plugins/katello/nightly/user_guide/remote_databases/index.md
johnpmitsch: The purpose of the online backup is to get the SQL/mongo dumps, since offline will not provide them. We need to use the dumps to seed the remote dbs.  We instruct users to shut down non-database services so nothing is written to the databases at the time of the backup, ensuring an accurate backup.

I'm not sure what action you are looking for here? Can you elaborate on your concerns?

johnpmitsch:
@ehelms updated to install from mongo SCL and changed Katello references to "Foreman" with a disclaimer at the top that we are assuming Foreman has Katello installed.

I need some clarification here before I can update based on your comment, otherwise your comments should be addressed: https://github.com/theforeman/theforeman.org/pull/1102#discussion_r193174092

ehelms:
@johnpmitsch Do you want to copy this to the 3.7 doc sin this PR or as a follow up?

johnpmitsch:
@ehelms I'll do a follow-up

ehelms (approved):

ehelms:
@johnpmitsch Merged -- please do follow up, thanks!

PR #1100, opened 2018-06-01
Add release notes for 1.18

xprazak2:

PR #1107, opened 2018-06-11
Add 1.18 into doc menu

xprazak2:
Adds 1.18 into the versions menu thus making the RC1 publicly visible. Should be merged after #1106 and #1100


tbrisker (approved):
Thanks @xprazak2 !

PR #1110, opened 2018-06-13
Katello 3.7 API Docs

jturel:
Generated from the respective RC1 branches of foreman and katello


akofink:
plugins/katello/3.7/api/index.html shows: "Api documentation is not generated for nightly builds. Browse to https://HOSTNAME/apidoc/ for the local version of the api documentation on any installed foreman server."

~I think you need to update [this page](https://github.com/theforeman/theforeman.org/blob/gh-pages/plugins/katello/3.7/api/apidoc.html#L16).~

akofink:
Actually, I think you need to remove the apipie-cache directory. Look at (i.e.) 3.5 https://github.com/theforeman/theforeman.org/blob/gh-pages/plugins/katello/3.5/api/apidoc.html

jturel:
@akofink thanks for the close look. It wasn't quite what you mentioned but it gave me a clue. I need to go back and update our release docs since they're not quite right w.r.t generating the api docs.

akofink:
Hmm... is it working for you? I'm still getting "Api documentation is not generated for nightly builds. ..." when I visit http://lo:4000/plugins/katello/3.7/api/index.html

akofink:
Also, for the record, 3.6 apidocs are also broken atm. :(

jturel:
Yeah it works fine for me - maybe try restarting jekyll? (although you shouldnt have to in theory)

akofink (approved):
Ah! It's working after another fetch. :see_no_evil: ACK! Thanks @jturel!

PR #1112, opened 2018-06-14
Add a docker instructions for easy development/testing

akofink:
Use the instructions in the Dockerfile to test!

Note: `rm Gemfile.lock` first

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,9 @@ to learn and to become rapidly productive!
 
 Everything you push into the gh-pages branch goes live immediately!
 
-## Testing locally
+## Testing
+
+### Locally
 
 To test your changes locally use
 
@@ -18,6 +20,14 @@ to generate your site in the \_site directory, or
     # rake server
 
 To start Jekyll server locally.
+
+### With Docker
+
+```
+docker run -it --rm -v `pwd`:/srv/jekyll:Z -p 4000:4000 jekyll/jekyll jekyll serve --watch
+```
+
+Navigate to http://localhost:4000
 
 ## Contributing
 


ehelms:
Looks good but I am curious what this adds over simply adding to the README the command to run the Jekyll container directly? The intermediate container doesn't seem to add much value?

akofink:
:rofl: didn't think of that! I will simply add some docs to the Readme.

jturel:
Good point @ehelms. This did work nicely though :)

akofink:
Updated!

ehelms:
Looks good... final question, is there value in adding a Rake task to run what's in the README? Something like:

```
rake docker:serve
```

akofink:
No way! ;) If I use rake, I have to install ruby + gems, and that's the whole point of using Docker.

jturel:
Works for me - thanks!

jturel (approved):

ehelms:
Fair enough, you should only need ruby + rake but it was worth asking :)

ehelms (approved):

README.md
```diff
@@ -19,6 +19,14 @@ to generate your site in the \_site directory, or
 
 To start Jekyll server locally.
 
+## With Docker
```
sean797: Shouldn't this be a subheading of  `## Testing locally` above?

akofink: This is running *not* locally but in a container, so I don't think it makes sense under that header. Also this isn't for running the tests (though I guess you could), it's for running the jekyll app for manual testing.

sean797: right okay, then I think either of the following would make better sense.

## Testing 

### Locally

### With Docker

---------------------

## Testing locally

## Testing with Docker

akofink: But as I said, that header is for running the automated tests, not for running the app. If I were to move it under a top level header, I'd move it to the contributing section. Eh?

sean797: There are no automated tests. I'm fairly sure what you added is the same running `rake server`

akofink: Doh! You're right :) Ok, I'll put it under the testing section.

akofink:
Updated

sean797 (approved):

stbenjam (approved):
I think everyone's happy now, so merging.

PR #1129, opened 2018-07-10
refs #23851 - discovery: auto bond creation

timogoebel:


lzap:
Short enough :-) Thanks.

PR #1130, opened 2018-07-10
Creating nightly manual dir

lzap:
This is proposal to follow katello and discovery process: have "nightly" directory holding up-to-date docs and creating copy for each release when needed. Until now, we had to create copy first and then put changes as cherry picks - awful workflow.

@tbrisker @ewoud


tbrisker (approved):
I like this approach! It will make updating the manual much smoother I believe as changes in core can come with a matching change in the manual and not wait for the new version to be created.

ekohl:
I've been thinking about this and I think I like it, but I'm not sure it's the only way.

An alternative is to copy the docs directly after branching. That means that now develop is 1.20, we should already create the 1.20 docs to be. The downside of that is that it's less clear for users that we already have 1.20 docs when 1.19 hasn't even been released. This can probably we worked around by just using the nightly label but the 1.20 directory names. Not sure how much of a difference this really is though.

tbrisker:
Lets give this a try than and see how it goes.

PR #1131, opened 2018-07-11
Update release note script for new Redmine fields

GregSutcliffe:
This updates the script to use `fixed_version_id` instead of `release_id` (i.e. the Target Version field).

I also updated it so it works for any project and without having to know the version ID. For example:

```
[greg@opal]$ ./release_notes.rb foreman 1.18.0
#### API
* Some API parameters are wrongly documented as enum types ([#22753](https://projects.theforeman.org/issues/22753))
* Validation on Puppet Class ID When using API ([#22747](https://projects.theforeman.org/issues/22747))
* Extend apidoc with list of fields to use in a search ([#17964](https://projects.theforeman.org/issues/17964))
* Missing API for external auth sources ([#15286](https://projects.theforeman.org/issues/15286))
* Extract API v1 into a plugin ([#11389](https://projects.theforeman.org/issues/11389))

#### Audit Log
* audit:  undefined method `each' for "":String ([#23965](https://projects.theforeman.org/issues/23965))
* host global_status changes should not be audited ([#23913](https://projects.theforeman.org/issues/23913))
```

but also
```
[greg@opal]$ ./release_notes.rb salt 10.1.0
#### API
* Foreman Salt Plugin 10.0 upload-salt-reports error with Foreman 1.17 ([#23658](https://projects.theforeman.org/issues/23658))
```

works for Katello too (cc/ @akofink @jturel)
```
[greg@opal]$ ./release_notes.rb katello "Katello 3.7.0"
#### API
* Subscription quantity available should not show less than -1 ([#23535](https://projects.theforeman.org/issues/23535))
* Add available attribute to upstream subscriptions API ([#23533](https://projects.theforeman.org/issues/23533))
* NumberValidator should come from apipie-rails ([#23460](https://projects.theforeman.org/issues/23460))
* Add param to limit upstream subs index to attachable ones ([#23182](https://projects.theforeman.org/issues/23182))
* Set http proxy in a thread-safe way ([#23137](https://projects.theforeman.org/issues/23137))
* Add API to update upstream_entitlement quantities ([#23111](https://projects.theforeman.org/issues/23111))
```

I can *probably* auto close the version, but I've left that alone for now. Likewise the `modify_target_version` method seems unused at the moment, so I've left it alone.

### DIFF
--- a/scripts/release_notes.rb
+++ b/scripts/release_notes.rb
@@ -7,7 +7,6 @@
 require 'uri'
 require 'net/http'
 require 'json'
-require 'pry-byebug'
 
 URL = 'https://projects.theforeman.org'
 
@@ -17,15 +16,40 @@ URL = 'https://projects.theforeman.org'
 # Create new release via API call
 # Create new search by open + release field
 
-unless @current_release_id = ARGV[0]
-  puts "Usage: #{$0} RELEASE_ID"
-  puts "Hint: 1.16.0 is 240"
+unless ARGV.size == 2
+  puts "Usage: #{$0} PROJECT RELEASE_NAME"
+  puts "Example: #{$0} foreman 1.18.0"
+  exit 1
+end
+
+@project_name = ARGV[0]
+@current_release_name = ARGV[1]
+
+# Determine version
+def get_version_id
+  url = "#{URL}/projects/#{@project_name}/versions.json"
+  uri = URI(URI.escape(url))
+  response = Net::HTTP.get(uri)
+  result = JSON.parse(response)
+  return nil if result['versions'].nil?
+  result = result['versions'].detect{|v| v['name'] == @current_release_name }
+  return nil if result.nil?
+  result['id']
+rescue
+  puts "Error getting version list from #{url}"
+  exit 1
+end
+
+@current_release_id = get_version_id
+
+if @current_release_id.nil?
+  puts "Release #{@current_release_name} not found in project #{@project_name}"
   exit 1
 end
 
 # Move bugs with Release set to VERSION to NEXT-VERSION with status new
 def gather_issues(offset=0)
-  url = "#{URL}/projects/foreman/issues.json?status_id=closed&offset=#{offset}&limit=100&release_id=#{@current_release_id}"
+  url = "#{URL}/projects/#{@project_name}/issues.json?status_id=closed&offset=#{offset}&limit=100&fixed_version_id=#{@current_release_id}"
   uri = URI(URI.escape(url))
   response = Net::HTTP.get(uri)
 	result = JSON.parse(response)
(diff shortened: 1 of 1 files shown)


scripts/release_notes.rb
```diff
@@ -9,24 +9,45 @@
...
+
+@project_name = ARGV[0]
+@current_release_name = ARGV[1]
+
+# Determine version
+def get_version_id
+  url = "#{URL}/projects/#{@project_name}/versions.json"
+  uri = URI(URI.escape(url))
+  response = Net::HTTP.get(uri)
+	result = JSON.parse(response)
```
ekohl: Tabs vs spaces here

scripts/release_notes.rb
```diff
@@ -17,15 +17,37 @@
...
+  result['versions'].detect{|v| v['name'] == @current_release_name }['id']
```
ekohl: I think it's cleaner to check if the `result.detect` is nil than catching it. `rescue` could then be used to catch network errors.

Is there pagination in the versions?

akofink:
`require pry-byebug` but it's not in the Gemfile?

```
Traceback (most recent call last):
        2: from ./scripts/release_notes.rb:10:in `<main>'
        1: from /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:59:in `require'
/usr/share/rubygems/rubygems/core_ext/kernel_require.rb:59:in `require': cannot load such file -- pry-byebug (LoadError)
```

scripts/release_notes.rb
GregSutcliffe: Not that I can see, no. You're right about the rescue, will fix tomorrow

GregSutcliffe:
@akofink yeah I hit that too - the scripts folder is a bit of a dumping ground for useful thingies, rather than being an integrated part of the repo. I just gem-installed it ;)

ekohl:
> require pry-byebug but it's not in the Gemfile?

I suspect it's just a debug leftover and we should remove it.

GregSutcliffe:
@ekohl updated and dropped the debug gem too

ekohl:
We also need to change links at https://theforeman.org/manuals/1.17/index.html#Releasenotesfor1.17 (A full list of changes in 1.17.0 is available via [Redmine](http://projects.theforeman.org/rb/issues/backlog/product/foreman?release_id=296)).

Note that because of version unsharing you now need to loop over all projects. We also need to link to all subprojects on release note pages.

akofink (approved):
Works well for me with the following command:

```sh
docker run -it --rm -v `pwd`:/usr/src/app:Z  -w /usr/src/app ruby bundle && ./scripts/release_notes.rb katello 'Katello 3.7.0'
```

akofink:
And now it gives a helpful error message when the release doesn't exist:

```
Release 3.7 not found in project katello
```

:+1: it's an ACK from me! Thanks @GregSutcliffe!

jturel (approved):
I only tested cherry picking for the katello 3.7 release but it looks good :)

tbrisker:
Thanks @GregSutcliffe! I've incorporated this change in #1145 which I used to create the 1.19 release notes. Shall we merge it and rebase my pr on top of it or close this pr and include the change when the other one is merged?

tbrisker:
Closing in favor of #1145 which also includes these changes.

PR #1132, opened 2018-07-11
Remove OpenSourceCamp from alerts

ekohl:


GregSutcliffe:
Thanks for the heads up @ekohl, I'd forgotten about that. I think it's right to promote our birthday though, so I'll close this and merge a commit doing that instead  :+1:

PR #1134, opened 2018-07-12
Update Puppet releasing docs

ekohl:


dLobatog (approved):
Changelog task is merged, so this should be merged too, thanks @ekohl

PR #1137, opened 2018-07-12
Manual: Non-ENC-mode needs cronjob to push facts

ghost:
Give a hint, that for pushing facts to Foreman when not using the ENC functionality, a cronjob is needed therefore.


dLobatog (approved):
Thanks @wodry

PR #1138, opened 2018-07-13
Update Katello triage process after redmine update

akofink:
We've talked for a while now about changing this to only be one link. The Redmine update was a good catalyst to finally do it! In addition, I've talked with @jlsherrill about the changes in Redmine and what it means for our triage process, and this is what we've come up with.

### DIFF
--- a/plugins/katello/developers.md
+++ b/plugins/katello/developers.md
@@ -18,18 +18,15 @@ For general contribution information, see [here](/contribute.html).
 
 ## Triage Process Links
 
-The Katello bug triage process is designed to address Redmine issues that have been opened by the community and developers in order to target the releases that these issues will be included in. This should be done on a weekly basis and in the following order of items to be addressed:
+The Katello bug triage process is designed to address Redmine issues that have been opened by the community and developers in order to target the releases that these issues will be included in. This should be done on a weekly basis and in a top down order of items to be addressed:
 
-  * [Open Recycle Bin](http://projects.theforeman.org/projects/katello/issues?utf8=%E2%9C%93&set_filter=1&f%5B%5D=status_id&op%5Bstatus_id%5D=o&f%5B%5D=release_id&op%5Brelease_id%5D=%3D&v%5Brelease_id%5D%5B%5D=166&f%5B%5D=&c%5B%5D=tracker&c%5B%5D=status&c%5B%5D=priority&c%5B%5D=subject&c%5B%5D=author&c%5B%5D=assigned_to&c%5B%5D=updated_on&c%5B%5D=category&c%5B%5D=fixed_version&c%5B%5D=position&group_by=fixed_version)
-  * [Untriaged High Priority](http://projects.theforeman.org/projects/katello/issues?utf8=%E2%9C%93&set_filter=1&f%5B%5D=release_id&op%5Brelease_id%5D=%21*&f%5B%5D=subproject_id&op%5Bsubproject_id%5D=%21*&f%5B%5D=priority_id&op%5Bpriority_id%5D=%3D&v%5Bpriority_id%5D%5B%5D=5&v%5Bpriority_id%5D%5B%5D=6&v%5Bpriority_id%5D%5B%5D=7&f%5B%5D=&c%5B%5D=tracker&c%5B%5D=status&c%5B%5D=priority&c%5B%5D=subject&c%5B%5D=author&c%5B%5D=assigned_to&c%5B%5D=updated_on&c%5B%5D=category&c%5B%5D=fixed_version&group_by=)
-  * [Untriaged Closed](http://projects.theforeman.org/projects/katello/issues?utf8=%E2%9C%93&set_filter=1&f%5B%5D=release_id&op%5Brelease_id%5D=%21*&f%5B%5D=subproject_id&op%5Bsubproject_id%5D=%21*&f%5B%5D=status_id&op%5Bstatus_id%5D=c&f%5B%5D=&c%5B%5D=tracker&c%5B%5D=status&c%5B%5D=priority&c%5B%5D=subject&c%5B%5D=author&c%5B%5D=assigned_to&c%5B%5D=updated_on&c%5B%5D=category&c%5B%5D=fixed_version&group_by=)
-  * [Untriaged Open](http://projects.theforeman.org/projects/katello/issues?utf8=%E2%9C%93&set_filter=1&f%5B%5D=release_id&op%5Brelease_id%5D=%21*&f%5B%5D=subproject_id&op%5Bsubproject_id%5D=%21*&f%5B%5D=status_id&op%5Bstatus_id%5D=o&f%5B%5D=&c%5B%5D=tracker&c%5B%5D=status&c%5B%5D=priority&c%5B%5D=subject&c%5B%5D=author&c%5B%5D=assigned_to&c%5B%5D=updated_on&c%5B%5D=category&c%5B%5D=fixed_version&group_by=)
+  * [Untriaged](https://projects.theforeman.org/projects/katello/issues?query_id=54)
 
 ### Triage Runner
 
 The runner is the person in charge of following the procedures here and generally conducting the meeting.
 
-We follow a rotation with respect to who runs the triage session. You can see the rotation [here](http://projects.theforeman.org/projects/katello/wiki/Triage_Rotation).
+We follow a rotation with respect to who runs the triage session. You can see the rotation [here](http://projects.theforeman.org/projects/katello/wiki/Triage_Rotation). Feel free to sign up if you want to run triage some time!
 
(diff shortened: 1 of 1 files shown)


jlsherrill:
Mind updating the 'Release States' below? and all the stuff that references the recycle bin?  Maybe put in section about the need_info -> rejected workflow we discussed with some text we could copy/paste when rejecting an issue that hits the 2 week mark.

akofink:
Updated

plugins/katello/developers.md
```diff
@@ -45,53 +42,56 @@ There are a few important states that issues can be in that are worth knowing ab
...
+The triage runner starts the meeting by going through the issues in descending order. This is so that if time were to run out, the most critical categories of issues are triaged. The runner is in charge of ensuring the flow of the meeting and updating issues or ensuring that someone on the call takes responsibility for updating an issue. This latter part sometimes occurs if a duplicate needs to be looked up, or another developer on the call can explain/ask the user for more information more clearly than the runner.
 
 #### General Flow for New or Needs More Information
 
-The general flow is to open an issue and the runner to exam the issue, typically reading the general idea aloud. This allows for any developers with knowledge to chime in with any relevant info. The runner attempts to answers the following questions:
+The general flow is to open an issue and the runner to examine the issue, typically reading the issue number and general idea aloud. This allows for any developers with knowledge to chime in with any relevant info. The runner attempts to answers the following questions:
```
jturel: s/answers/answer

jturel (approved):
APT

(ack pending typo)

Thanks!

akofink:
Fixed the typo

plugins/katello/developers.md
```diff
@@ -45,53 +42,56 @@ There are a few important states that issues can be in that are worth knowing ab
...
+> Please reopen this issue with the requested information, and we will be happy to take a look.
 
 ### Release States
 
 There are a few important states the issue can be in with respect to the 'Release' field within Redmine. These release states have an impact not only a release itself but the triage process as a whole. A quick definition of each that will make more sense in the process section:
 
-  * Empty: No release has been set, this is the "untriaged" bucket
-  * Katello Backlog: Issues that have been put on the backlog and not scheduled for a release due to any number of reasons
-  * Katello Recycle Bin: Issues that have been rejected, or are a duplicate, or have been in 'Needs more information' for longer than a week
+  * Empty: No release has been set. Triaged issues with empty release states are implied backlog (open issues) or recycle bin (closed issues).
```
ehelms: Are Backlog and Recycle Bin gone?

plugins/katello/developers.md
```diff
@@ -45,53 +42,56 @@ There are a few important states that issues can be in that are worth knowing ab
...
+  * What release should the issue be targeted for? (e.g. next z-stream, current release, next release)
 
 ### Notes
 
 A few notes about some automatic operations that are performed by the prprocessor that help facilitate this workflow.
 
-  * All issues are set to an empty release, which again represents "untriaged"
-  * Any issues on the backlog that a PR is opened for has its release set to 'empty' to ensure that it flows through the triage process
-  * Developers that believe an issue should be considered for a release, should set the release to 'empty' and show up to triage to argue their case
+  * All issues are set to an empty release and untriaged
```
ehelms: Did the code get changed to account for this? I would also emphasize that "triaged" is a field vs. the concept of triage

plugins/katello/developers.md
```diff
@@ -45,53 +42,56 @@ There are a few important states that issues can be in that are worth knowing ab
...
+  * Any issues on the backlog that a PR is opened for has its release set to 'empty' and triaged to false to ensure that it flows through the triage process
+  * Developers that believe an issue should be considered for a release should set triaged to false and show up to triage to argue their case
```
ehelms: So, this goes back to the old way of doing triage that we had issues with (and hence the current process). What does the following mean now:

 * On a release, triage unset
 * Not on a release, triage set

The original problem we were solving with the current method (and simplifying) was that we did not need a triaged flag. Rather, release means triaged. I believe the only flaw we ever saw was that users could directly set issues on a release which did happen but rarer. I think given the two scenarios above, if we are going to use the triaged flag, the process needs to be modified to account for those states.

ehelms:
The PR title is a bit misleading since it does not simply update links, it changes the process (which can be OK but we should discuss).

plugins/katello/developers.md
akofink: Yes

akofink: Wait, no, they're still there, but we're not going to use them anymore.

plugins/katello/developers.md
akofink: > On a release, triage unset

This basically means a user set the target release (on accident or on purpose), almost like, "please put it on this release". When we see it in triage, we'll set the triage flag and update the target release (or leave it).

> Not on a release, triage set

If the issue is open, it's implicitly on the backlog. If it's closed, it's implicitly in the recycle bin. When issues move to ready for testing, the triage flag should be cleared by prprocessor.

plugins/katello/developers.md
akofink: We still need to update prprocessor.

plugins/katello/developers.md
ehelms: Can you explain why they are being dropped?

plugins/katello/developers.md
ehelms: I felt like the backlog was a nice and specific abstraction for putting things in that bucket that was obvious for users. It would read in the issue "Backlog" for example. Now I feel things are less obvious to users and to developers looking through issues.

I am still trying to wrap my head around what the "Triaged" flag now means. Triaged used to mean the team has reviewed and assigned an issue to a release. If a user wanted it re-considered they could unset the release or pop it off the backlog. 

The doc here says "Developers that believe" but what about users? :)

plugins/katello/developers.md
akofink: Because they are redundant information now that we have the triaged flag.

plugins/katello/developers.md
akofink: I agree with the "developers" part and will update it. We don't need to get rid of the backlog and recycle bin targets if you'd rather keep them and can convince @jlsherrill. I agree it makes it more obvious.

plugins/katello/developers.md
ehelms: So triaged, closed represents Recycle Bin? How are Recycle Bin candidate issues represented? 

The part I dislike is "implied backlog". I personally liked that there was no implication before. The backlog was obvious and we are losing that now. I feel that's less obvious to users and even developers.

plugins/katello/developers.md
ehelms: Hah, so @jlsherrill is the real culprit here? Let's see if pinging him a few times gets him to pop in here.

plugins/katello/developers.md
akofink: Yes, I agree it's less obvious. @jlsherrill !?!? ;)

plugins/katello/developers.md
johnpmitsch: I would just say "leave Triaged unchecked" instead of "set triage to false" as it is a passive action

johnpmitsch:
One small comment, but otherwise looks good. There are some good discussions here and maybe some broader questions, but I would advocate to merge this sooner rather than later and update the docs as we finalize our process as the links there now are broken

plugins/katello/developers.md
johnpmitsch: `All issues are set to an empty release and untriaged` when? May want to specify

plugins/katello/developers.md
```diff
@@ -45,53 +42,56 @@ There are a few important states that issues can be in that are worth knowing ab
...
 
 ### Notes
 
 A few notes about some automatic operations that are performed by the prprocessor that help facilitate this workflow.
 
-  * All issues are set to an empty release, which again represents "untriaged"
-  * Any issues on the backlog that a PR is opened for has its release set to 'empty' to ensure that it flows through the triage process
-  * Developers that believe an issue should be considered for a release, should set the release to 'empty' and show up to triage to argue their case
+  * All issues are set to an empty release and untriaged
+  * Any issues on the backlog that a PR is opened for has its release set to 'empty' and triaged to false to ensure that it flows through the triage process
```
johnpmitsch: I would spell out pull request

plugins/katello/developers.md
akofink: I think we should keep the recycle bin and backlog targets and use them like we have been so it's not confusing.

akofink:
Updated. Also, related prprocessor PR: https://github.com/theforeman/prprocessor/pull/117

johnpmitsch (approved):

jlsherrill (approved):

PR #1141, opened 2018-07-17
Katello 3.8 docs

sjha4:
Opening this PR to get Katello 3.8 docs in.

@jturel , @akofink .. Does this look ok to you?

I ran the tool_belt branch docs command and then basically this : grep -rl '3.7' ./ | xargs sed -i 's/3.7/3.8/g' for the files here in the instructions here:

 - Update the sidebar at repos/katello_3.8.0/theforeman.org/_includes/plugins/katello
 - Update the docs at repos/katello_3.8.0/theforeman.org/plugins/katello/3.8.
 - Add a link to 3.8 in repos/katello_3.8.0/theforeman.org/_layouts/plugins/katello/documentation.html.


akofink (commented):
- Add 3.8 to https://github.com/theforeman/theforeman.org/blob/gh-pages/plugins/katello/index.md

Other than that... are we supporting F28 clients? If so, we should add that to the client installation OS list. Oh, and you might should use 3.8 RC as the version name for now.

ehelms:
@akofink We are working on F28 support but have to fix Koji first.

sjha4:
Does this look good?

akofink (approved):
Yes! This is also something you can easily tweak later if we find we missed something.

PR #1144, opened 2018-07-23
Initial 1.19 Manual

tbrisker:
TODO:

- [x] Headline Features
- [x] Contributors
- [x] API docs

PR #1145, opened 2018-07-23
Update release notes script to new redmine releases

tbrisker:


scripts/release_notes.rb
```diff
@@ -7,25 +7,17 @@
...
   puts "Usage: #{$0} RELEASE_ID"
   puts "Hint: 1.16.0 is 240"
   exit 1
 end
 
-# Move bugs with Release set to VERSION to NEXT-VERSION with status new
-def gather_issues(offset=0)
-  url = "#{URL}/projects/foreman/issues.json?status_id=closed&offset=#{offset}&limit=100&release_id=#{@current_release_id}"
+def gather_issues(offset = 0, feature = true)
+  url = "#{URL}/projects/foreman/issues.json?status_id=closed&offset=#{offset}&limit=100&f[]=cf_12&op[cf_12]==&v[cf_12][]=#{@current_release_id}&f[]=tracker_id&op[tracker_id]=#{feature ? '=':'!'}&v[tracker_id][]=2"
```
ekohl: Right now you also have to loop over the projects because we no longer inherit versions.

ekohl (changes requested):
Please also have a look at https://github.com/theforeman/theforeman.org/pull/1131. It might be nicer to automatically get the version ID so you can just pass in release numbers (1.19.0).

scripts/release_notes.rb
tbrisker: Right, i'll fix this to handle multiple versions.

tbrisker:
I noticed #1131 only after sending this, I've rebased this PR on top of that.

scripts/release_notes.rb
```diff
@@ -7,57 +7,89 @@
...
+end
 
-  issues.each do |issue|
-	  puts "* #{issue['subject']} ([##{issue['id']}](#{URL}/issues/#{issue['id']}))"
+def releases_filter
+  @current_release_ids.map{|release| "v[cf_12][]=#{release}"}.join('&')
+end
+
+def gather_issues(offset = 0, feature = true)
+  url = "#{URL}/projects/foreman/issues.json?status_id=closed&offset=#{offset}&limit=100&f[]=cf_12&op[cf_12]==&#{releases_filter}&f[]=tracker_id&op[tracker_id]=#{feature ? '=':'!'}&v[tracker_id][]=2"
```
ekohl: Maybe we should store the parameters in a hash and use `URI.encode_www_form`. Probably easier to understand. You also have a `op[cf_12]==` but is that intended?

scripts/release_notes.rb
```diff
@@ -7,57 +7,89 @@
...
+  result = url_to_json(url)
+  if result['total_count'].to_i - offset - 100 <= 0
+    result['issues']
+  else
+    result['issues'] += gather_issues(offset + 100)
```
ekohl, line 58: You probably need to pass in `feature` here for proper recursion.

ekohl (changes requested):

scripts/release_notes.rb
tbrisker: `URI.encode_www_form` doesn't seem to handle arrays in the way rails expects them, but maybe i'm doing it wrong:
```
[5] pry(main)> URI.encode_www_form({f: ['cf_12', 'status'], op: {cf_12: '='}, v: {cf_12: [1,2]}})
=> "f=cf_12&f=status&op=%7B%3Acf_12%3D%3E%22%3D%22%7D&v=%7B%3Acf_12%3D%3E%5B1%2C+2%5D%7D"
```
`op[cf_12]==` is intended, but i think the second = should be escaped - it indicates which action you're matching on (= - is, ! - isn't etc)

tbrisker: actually i take it back, if i escape it first it doesn't work, guessing double escape is breaking it

scripts/release_notes.rb
```diff
@@ -7,57 +7,91 @@
...
-def modify_target_version!(issue_id, options)
-  uri = URI(URI.escape("#{URL}/issues/#{issue_id}.json"))
-  req = Net::HTTP::Put.new(
-		uri,
-		{ 'Content-Type' => 'application/json',
-		  'X-Redmine-API-Key' => File.read('~/.redminekey') }
-	)
-  Net::HTTP.new(uri.host, uri.port).start {|http| http.request(req) }
+def gather_issues(offset = 0, feature = true)
+  url = "#{URL}/projects/foreman/issues.json?status_id=closed&offset=#{offset}&limit=100&f[]=cf_12&op[cf_12]==&#{releases_filter}&f[]=tracker_id&op[tracker_id]=#{feature ? '=':'!'}&v[tracker_id][]=2"
```
ekohl: `foreman` should be `#{project}` which should also be a parameter.

ekohl: Or just use `/issues.json` since the version filter is probably enough.

tbrisker: right now every project is under foreman so this works, but i'll update it.

tbrisker:
Also updated the search to only take issues with status is "closed", not a "closed" status (which also took resolved, duplicate, etc)

scripts/release_notes.rb
```diff
@@ -7,57 +7,91 @@
...
-# Check out installer modules
-# Compare packages in nightly vs release and retag dependencies
-
-# Create new release via API call
-# Create new search by open + release field
-
-unless @current_release_id = ARGV[0]
-  puts "Usage: #{$0} RELEASE_ID"
-  puts "Hint: 1.16.0 is 240"
+unless ARGV.size == 2
```
ehelms, line 13: Got an example of what this file generates?

tbrisker: https://github.com/theforeman/theforeman.org/pull/1144/files#diff-66a06a0e26dd946fb2dfb39ace1e11c9

scripts/release_notes.rb
```diff
@@ -7,57 +7,91 @@
...
+
+puts "Copy and paste the following into the release notes section of the relevant manual:"
+
+puts "\n----------[Start of notes]----------\n\n"
+
+puts '### Features'
+grouped_issues
+puts
+
+puts '### Bug Fixes'
```
ekohl: This isn't just bug fixes. It also includes the tracker `Refactor`

scripts/release_notes.rb
```diff
@@ -7,57 +7,91 @@
...
 
-def modify_target_version!(issue_id, options)
-  uri = URI(URI.escape("#{URL}/issues/#{issue_id}.json"))
-  req = Net::HTTP::Put.new(
-		uri,
-		{ 'Content-Type' => 'application/json',
-		  'X-Redmine-API-Key' => File.read('~/.redminekey') }
-	)
-  Net::HTTP.new(uri.host, uri.port).start {|http| http.request(req) }
+def gather_issues(offset = 0, feature = true)
```
ekohl: I wonder if we should split this by `tracker` in the result rather than retrieve it twice.

scripts/release_notes.rb
```diff
@@ -7,57 +7,91 @@
...
+  url = "#{URL}/issues.json?f[]=status_id&op[status_id]==&v[status_id][]=5&offset=#{offset}&limit=100&f[]=cf_12&op[cf_12]==&#{releases_filter}&f[]=tracker_id&op[tracker_id]=#{feature ? '=':'!'}&v[tracker_id][]=2"
```
ekohl: I'm not sure we should only include by status_id closed. IMHO we should take any "closed" issues with `Fixed in` is this version. On duplicates we shouldn't set `Fixed in`.

tbrisker:
@ekohl - removed the separation by tracker type, if anyone cares about that they can check redmine. also updated to take all closed issues instead of only status=closed

tbrisker:
ping @ekohl ?

ekohl (approved):

PR #1155, opened 2018-07-27
Removed TEMPLATE_CONTEXT from logs

lzap:
This is for https://github.com/theforeman/foreman/pull/5683

### DIFF
--- a/_includes/manuals/nightly/7.2_debugging.md
+++ b/_includes/manuals/nightly/7.2_debugging.md
@@ -111,11 +111,6 @@ Here is the list of most important structured fields which can help with debuggi
     <td>Template name (blob logger)</td>
   </tr>
   <tr>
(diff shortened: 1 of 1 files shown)


tbrisker:
Thanks @lzap, is this the only change needed in the manual regarding the rendering change?

tbrisker:
@timogoebel same question :wink:

timogoebel:
Good catch.

I think this should be all as we mostly did a refactoring. What did change was, that you cannot use the template macros defined in https://github.com/theforeman/foreman/blob/develop/lib/foreman/renderer/scope/variables/base.rb in partition tables anymore. But this should not hurt.

lzap:
Yeah this field is not a big deal.

tbrisker (approved):
Thanks @lzap and @timogoebel !

PR #1156, opened 2018-07-28
Fix `single test` instructions

alexjfisher:
Thanks to @mmoll for help with this.


tbrisker (approved):
Thanks @alexjfisher !

PR #1158, opened 2018-07-30
Remove `password_hash` setting from the manual

tbrisker:
This setting has never actually been implemented. See
https://github.com/theforeman/foreman/pull/1736 for details.


ekohl (approved):
Interesting...

PR #1161, opened 2018-08-01
trusted_puppetmaster_hosts renamed to trusted_hosts

tbrisker:


_includes/manuals/1.18/1.2_release_notes.md
```diff
@@ -286,6 +286,8 @@ The following deprecations with target version 1.20 were added:
 
 ### Upgrade warnings
 
+The `trusted_puppetmaster_hosts` setting has been renamed to `trusted_hosts`.
+
 The following plugins are known to be incompatible with 1.18, their updated versions have not been released yet:
 
 * foreman_chef 0.7.0 and lower
```
ekohl, line 293: This can removed as well.

_includes/manuals/1.18/3.5.2_configuration_options.md
```diff
@@ -578,9 +578,9 @@ Default: The private key file used by puppet
 Time in minutes installation tokens should be valid for, 0 to disable token generation.
 Default: 360 (6 hours)
 
-##### trusted_puppetmaster_hosts
+##### trusted_hosts
```
ekohl, line 581: IMHO this is a badly named configuration option. It's very wide and could be interpreted as "fully trust this host". IMHO it should have been `trusted_configuration_management_hosts` or similar. Guess it's too late now.

ekohl (approved):

_includes/manuals/1.18/1.2_release_notes.md
tbrisker: ok, but i'd rather do that in a separate pr.

_includes/manuals/1.18/3.5.2_configuration_options.md
tbrisker: Welcome to open an issue and rename it in core, but for now the docs should match the code :)

_includes/manuals/1.18/1.2_release_notes.md
tbrisker: Also, are you sure? foreman_chef 0.7 seems to already include some fixes for 1.17 compat, but i'm not sure about 1.18 (cc: @ares)

ekohl: This wasn't actually true at the release. Perhaps during RC1 we didn't have 0.7.0 but somewhere during the development cycle we didn't have foreman_chef packages that were compatible with the current code. It was fixed in https://github.com/theforeman/foreman-packaging/pull/2610

_includes/manuals/1.18/3.5.2_configuration_options.md
ekohl: Hence my ACK but with comments :)

PR #1164, opened 2018-08-02
Remove reference to long gone mysql adapter

tbrisker:


ekohl (approved):
I wouldn't bother updating manuals prior to 1.15 for such a message but since you already did, it's probably fine to just merge this.

tbrisker:
@ekohl generally that makes sense but i'm using these changes to practice my vim and sed skills :sweat_smile:

PR #1165, opened 2018-08-02
Update 1.19 and nightly manual

tbrisker:
@ekohl - I think this should include all of the remaining comments from your review.

### DIFF
--- a/_includes/manuals/1.19/3.6_upgrade.md
+++ b/_includes/manuals/1.19/3.6_upgrade.md
@@ -42,10 +42,13 @@ Clean up the yum cache:
 
     yum clean all
 
-First upgrade or install the SCL release package to add rh-ruby24 and
-tfm-ror51 repositories:
+First install or upgrade the SCL release package to add the rh-ruby24 repository:
 
     yum install foreman-release-scl
+
+or:
+
+    yum upgrade foreman-release-scl
 
 Next upgrade all Foreman packages:
 
--- a/_includes/manuals/nightly/3.6_upgrade.md
+++ b/_includes/manuals/nightly/3.6_upgrade.md
@@ -42,10 +42,13 @@ Clean up the yum cache:
 
     yum clean all
 
-First upgrade or install the SCL release package to add rh-ruby24 and
-tfm-ror51 repositories:
+First install or upgrade the SCL release package to add the rh-ruby24 repository:
 
     yum install foreman-release-scl
+
+or:
+
+    yum upgrade foreman-release-scl
 
 Next upgrade all Foreman packages:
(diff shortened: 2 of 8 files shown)


_includes/manuals/1.19/3.6_upgrade.md
```diff
@@ -45,7 +45,7 @@ Clean up the yum cache:
 First upgrade or install the SCL release package to add rh-ruby24 and
 tfm-ror51 repositories:
```
ekohl: tfm-ror51 will no longer come from this repo since https://github.com/theforeman/foreman-packaging/commit/fc8687eb98f932d7ef1576d1a0ed93489bc17680

_includes/manuals/1.19/3.6_upgrade.md
```diff
@@ -45,7 +45,7 @@ Clean up the yum cache:
 First upgrade or install the SCL release package to add rh-ruby24 and
```
ekohl: This only upgrades. If the package is not yet installed then it'll do nothing:
```
[root@bd55df735b49 /]# yum upgrade foreman-release-scl
Loaded plugins: fastestmirror, ovl
Loading mirror speeds from cached hostfile
 * base: mirror.karneval.cz
 * extras: mirror.karneval.cz
 * updates: mirror.karneval.cz
Package(s) foreman-release-scl available, but not installed.
No packages marked for update
```

_includes/manuals/nightly/3.6_upgrade.md
```diff
@@ -45,7 +45,7 @@ Clean up the yum cache:
 First upgrade or install the SCL release package to add rh-ruby24 and
```
ekohl: Same comments as in 1.19

ekohl (changes requested):

_includes/manuals/1.19/3.6_upgrade.md
tbrisker: So should this step be completely removed?

ekohl: No, we still use `foreman-release-scl` to pull in `centos-release-scl` and `centos-release-scl-rh` (though we only need the latter I think) so it's still needed. We used to depend on `centos-release-scl` for `sclo-ror42` but we now host our own rails SCL.

tbrisker:
@ekohl addressed your comment

tbrisker:
ping @ekohl ?

ekohl (approved):
We probably need to go over all `/var/lib/puppet/ssl` paths but IMHO this good to go now.

PR #1175, opened 2018-08-17
Blog post on ESXi deployment through Foreman

amard33p:
This post aims to be the definitive guide for deploying ESXi through Foreman in both Legacy BIOS and UEFI modes.  
Request your reviews.


tbrisker:
@timogoebel - gihtub only lets me ask committers for reviews, but since you're familiar with vmware provisioning would be great if you can go over this post

timogoebel:
Nice, looks like we should try to add native support for this to Foreman. I've tried this once but failed to get it working.

_posts/2018-08-09-deploying-esxi-through-foreman.md
```diff
@@ -0,0 +1,402 @@
...
+  * **2.2.1:** Edit the file -   `/usr/share/foreman/app/models/concerns/pxe_loader_support.rb`
+    <pre><code>
+    def all_loaders_map(precision = 'x64')
+      {
+        "None" => "",
+        "PXELinux BIOS" => "pxelinux.0",
+        "PXELinux UEFI" => "pxelinux.efi",
+        <b>"ESXi PXELinux BIOS" => "pxelinux-3.86.0",</b> # Add this line
+        ...
+    </pre></code>
```
lzap: Feel free to file a PR to add this line. I'd not mention ESXi in the loader name tho to prevent user confusion.

_posts/2018-08-09-deploying-esxi-through-foreman.md
```diff
@@ -0,0 +1,402 @@
...
+    <img src="/static/images/blog_images/2018-08-16-deploying-esxi-through-foreman/esxi_installer_loading.png" width="400" />
+  </p>
+
+Now once installation is complete and the host reboots, it will try to PXEBoot and **should** be presented with the default PXE menu.
+<p float='left' align='center'>
+  <img src="/static/images/blog_images/2018-08-16-deploying-esxi-through-foreman/localboot.png" width="400" />
+</p>
+Upon timeout, the host should attempt a Localboot and boot to ESXi.
+
+However, since the host will boot into the pxelinux.0 v3.86 bootloader, without the supporting files of the same version (menu.c32 etc) the host will not be presented with this PXE menu and will never boot into ESXi.
```
lzap: Isn't better to put pxelinux v3 into a subdirectory like `pxelinux3/` with all the c32 files and create a relative symlink for `pxelinux3/pxelinux.cfg` pointing to the configuration directory one level above?

_posts/2018-08-09-deploying-esxi-through-foreman.md
```diff
@@ -0,0 +1,402 @@
...
+
+  * **1.3.2:** Create the Kickstart Template
+    * Click on **Create Template**
+    * Name - ESXi Minimal Kickstart
+    * A sample template:  
+      ```
+      vmaccepteula
+      keyboard 'US Default'
+      reboot
+      rootpw P@ssw0rd
```
ekohl: Would it be possible to use the password macro for this or does it have to be plain text?

lzap (approved):

_posts/2018-08-09-deploying-esxi-through-foreman.md
lzap: If this was named maybe `pxelinux-alt/` we could add a PXE loader named "PXELinux Alternative BIOS" -> `pxelinux-alt/pxelinux.0`. Symlink would do the job just fine and you could get rid of the clunky hook.

_posts/2018-08-09-deploying-esxi-through-foreman.md
amard33p: What would you like to call it? "PXELinux v3.86"?

amard33p: Also what about the EFI bootloader entry? "ESXi mboot UEFI"? Do you want that changed as well?

_posts/2018-08-09-deploying-esxi-through-foreman.md
```diff
@@ -0,0 +1,402 @@
...
+
+```sh
+cd /usr/share/foreman/config/hooks
+
+mkdir -p host/managed/before_provision
+
+cd host/managed/before_provision
+
+# Symlink the hook_functions file
+ln -s /usr/share/foreman/vendor/ruby/2.3.0/gems/foreman_hooks-0.3.14/examples/bash/hook_functions.sh .
```
ekohl: This path is dependent on the OS and package version. I think that on RH we place it `/usr/share/doc`. Maybe it's better to copy it?

_posts/2018-08-09-deploying-esxi-through-foreman.md
lzap: I'd suggest naming this "PXELinux Alt BIOS" -> `pxelinux.alt/pxelinux.0` because with symlink you could avoid the incompatibility with modules.

_posts/2018-08-09-deploying-esxi-through-foreman.md
amard33p: Removed plaintext password and used the macro now.

_posts/2018-08-09-deploying-esxi-through-foreman.md
amard33p: Have added a note at the top that warns about this. Otherwise a lot of paths needed to be changed.

_posts/2018-08-09-deploying-esxi-through-foreman.md
amard33p: I had kept it this way to allow for similarity between BIOS and UEFI. But now I realize that it was complicating an otherwise easy task. Have incorporated the changes you requested.

lzap:
Well 386 is a bit unfortunate naming (Intel 386 my first AMD procesor came to my mind first) but hey, that's fine :-)

lzap:
However I still think that the foreman hook is not necessary. If I understand correctly, VMWare bootloaders for both BIOS and UEFI is just syslinux and foreman can write regular syslinux configuration. If you put both EFI and BIOS files into proper directory and force it to load pxelinux.cfg/01-MAC which is written by foreman it should work.

However if you want to keep the current solution we are happy for new blogpost and I am good to merge.

amard33p:
> If you put both EFI and BIOS files into proper directory and force it to load pxelinux.cfg/01-MAC which is written by foreman it should work.  

From Page 11 of the VMWare official PXEboot [guide](https://www.vmware.com/content/dam/digitalmarketing/vmware/en/pdf/techpaper/vsphere-esxi-vcenter-server-60-pxe-boot-esxi.pdf), the 01-MAC dir needs to be under tftproot...not under pxelinux.cfg/. This is what forced me to use hook scripts.

lzap:
Thanks for the post!

amard33p:
Thank you very much for your help!

PR #1176, opened 2018-08-21
Update foreman-ansible configuration documentation

bastilian:
This updates the documentation to reflect the change to rely on REX settings for SSH connections.


tbrisker (approved):
Thanks @bastilian !

PR #1180, opened 2018-08-23
Remove clearing sessions command after upgrade

xprazak2:


ekohl:
Why exactly is this no longer needed?

xprazak2:
The command to clear sessions is no longer available in Rails 5.x
https://community.theforeman.org/t/documentation-update-needed/10868

ekohl:
I think clearing the sessions is still a good thing. In https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html they mention `db:sessions:clear` as a replacement. Should we use that instead?

xprazak2:
Yes, we probably should. Updated.

ekohl (approved):

mmoll:
merged, díky @xprazak2!

PR #1189, opened 2018-09-06
ignore .jekyll-metadata

bjozet:
As per https://jekyllrb.com/docs/structure/:

"This helps Jekyll keep track of which files have not been modified
since the site was last built, and which files will need to be
regenerated on the next build. This file will not be included in the
generated site. It’s probably a good idea to add this to your
.gitignore file."

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ _site
 .ruby-gemset
 .bundle
 Gemfile.lock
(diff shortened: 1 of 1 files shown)


tbrisker (approved):
Thanks @bjozet !

mmoll:
merged, tack @bjozet!

PR #1194, opened 2018-09-21
Update 3.6_Update.md with findings

petervreeburg:
During upgrade of 1.18 to 1.19, some issues did pop up, described in https://community.theforeman.org/t/issues-during-upgrade-1-18-2-to-1-19-following-the-upgrade-guide/11158
Added some small notes to make sure future upgrades may not suffer from this issue.

### DIFF
--- a/_includes/manuals/1.19/3.6_upgrade.md
+++ b/_includes/manuals/1.19/3.6_upgrade.md
@@ -74,7 +74,8 @@ number from the previous release or "stable" to `{{ page.version }}`:
     deb http://deb.theforeman.org/ stretch {{ page.version }}
     deb http://deb.theforeman.org/ plugins {{ page.version }}
 
-Next upgrade all Foreman packages:
+Next upgrade all Foreman packages.
+During the upgrade it's suggested to update the /etc/foreman/database.yml to the package version. If you used foreman-installer to setup your existing Foreman instance, make sure not to upgrade this file with the version from the package.
 
     apt-get update
     apt-get --only-upgrade install ruby\* foreman\*
@@ -126,7 +127,7 @@ You should compile i18n strings and precompile assets now:
 ##### Step 3 (A) - Database migration and cleanup
 
 The database should be migrated already, but you can make sure by executing the
-migration script again, it should produce no errors or output:
+migration script again, it should produce no errors or output. If you used foreman-installer to setup your existing Foreman instance, you can skip this step and jump over to step 3(C):
 
     foreman-rake db:migrate
     foreman-rake db:seed


_includes/manuals/1.19/3.6_upgrade.md
```diff
@@ -74,7 +74,8 @@ number from the previous release or "stable" to `{{ page.version }}`:
     deb http://deb.theforeman.org/ stretch {{ page.version }}
     deb http://deb.theforeman.org/ plugins {{ page.version }}
 
-Next upgrade all Foreman packages:
+Next upgrade all Foreman packages.
+During the upgrade it's suggested to update the /etc/foreman/database.yml to the package version. If you used foreman-installer to setup your existing Foreman instance, make sure not to upgrade this file with the version from the package.
```
dLobatog, line 78: How does one `update the /etc/foreman/database.yml to the package version.`?

_includes/manuals/1.19/3.6_upgrade.md
```diff
@@ -126,7 +127,7 @@ You should compile i18n strings and precompile assets now:
 ##### Step 3 (A) - Database migration and cleanup
 
 The database should be migrated already, but you can make sure by executing the
-migration script again, it should produce no errors or output:
+migration script again, it should produce no errors or output. If you used foreman-installer to setup your existing Foreman instance, you can skip this step and jump over to step 3(C):
```
dLobatog, line 130: I think the reason we included this section is just to make sure any errors during db migration are obvious - even if it means the user has to try to repeat the migration process after the installer does it.

dLobatog (commented):
Thanks @petervreeburg !

_includes/manuals/1.19/3.6_upgrade.md
ekohl: `apt upgrade` will ask you what to do with config files. This is very similar to running `rpmconf` on RH-based systems but apt will ask you during the upgrade.

ekohl: I'd say in general you probably don't want to replace this, but there can be relevant changes; like recommending the increasing the default pool size. Not sure how to phrase this though.

_includes/manuals/1.19/3.6_upgrade.md
ekohl: The installer doesn't clear the cache and existing sessions so totally jumping over it isn't correct. Running migrations and seeding doesn't hurt so I wouldn't skip it.

petervreeburg: That must have been my unfamiliarity (is that a word ?) with the complete solution.
Since the earlier change in the apt install step will make sure the database engine underneath the hood is not changes, I would think those steps will also not cause any problems.
Then.. would it be better to : 
a ) Remove this addition in step 3A all together
b ) Move those commands to after running the foreman-installer (in case the foreman-installer was used to setup...etc)
Please advise.

ekohl:
@tbrisker this is an old PR (and my apologies for it getting lost in my mailbox), but we should incorporate the changes into the 2.0 manual.

lzap:
@ekohl can you review if this patch was actually fixed by the "Rewrite upgrade instructions"? I am not sure what the intention was.

PR #1196, opened 2018-09-25
Enable incremental flag for faster builds

dLobatog:
This little thing has been there since Jekyll 3.0. Given our build time for the whole website on my laptop is 120 sec, the performance gains are quite considerable. It takes around 20s to regenerate a page of the manual, which is much nicer.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -12,7 +12,7 @@ end
 
 desc 'Start server'
 task :server do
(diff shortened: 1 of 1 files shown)


ekohl (approved):

lzap:
Yeah this was slow as hell, thanks :-) Yet 20secs for a page is terrible.

PR #1198, opened 2018-09-27
Fixes #24898 - added missing discovery docs

lzap:
I've forgot to update those during releases.


chris1984 (approved):
@lzap looks good to me

lzap:
@tbrisker or @GregSutcliffe

tbrisker (approved):
Didn't review in depth the entire text but I assume it's ok, if there are any issues with it we can always iterate.

PR #1200, opened 2018-10-03
Add docs for #24259

xprazak2:


ares (approved):
IMHO good to merge, grammar check from native speaker would be nice though, @GregSutcliffe would you find a minute?

tbrisker:
We branched the docs for 1.20 already, after @GregSutcliffe confirms please also copy the text to the 1.20 file

GregSutcliffe:
Seems OK. I'm not clear why we have a role that can grant itself a privilege escalation by design (might as well just make people an admin?) but that's probably out of scope for this change :)

ares:
@GregSutcliffe the reason is we want to have built-in role just for managing the instance, hiding unnecessary menus. That user can't escalate admin checkbox though. Thanks for checking, @xprazak2 could you please apply the same to 1.20?

The easiest thing that worked for me elsewhere:
```
git diff HEAD^ > change.diff
vim change.diff
:%s/nightly/1.20/g
git apply change.diff
rm change.diff
git commit -a --amend; git push -f
```

tbrisker:
In this case even simpler `cp _includes/manuals/{nightly,1.20}/4.1.2_roles_and_permissions.md` will do :wink:

xprazak2:
Updated.

tbrisker (approved):
Thanks @xprazak2

PR #1205, opened 2018-10-12
Typo in apidoc

pipopopo:
Typo in apidoc for host_tracer

### DIFF
--- a/api/1.15/apidoc/v2/host_tracer.html
+++ b/api/1.15/apidoc/v2/host_tracer.html
@@ -57,7 +57,7 @@
             GET /api/hosts/:host_id/tracer
           </a>
           <br>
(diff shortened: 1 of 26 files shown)


ekohl:
Looks like it was fixed in https://github.com/Katello/katello/commit/c3eb1961ad1598f0d098924fe7901b155d6579b2 which was also cherry picked to Katello 3.7.0-RC2. The fact that it's in the 3.8 docs tells me those weren't generated in a correct way.

@sjha4 could you regenerate the Katello 3.8 API docs for the website?

tbrisker (approved):
While these docs are usually auto-generated from source and looks like the type there was fixed, there's no harm in correcting the manuals. Thanks @pipopopo!

PR #1206, opened 2018-10-15
Add the 1.20 RPM GPG release key

ekohl:


tbrisker (approved):
ack, also verified this is indeed the correct key in a separate communication channel.

PR #1212, opened 2018-10-20
change apt source filename extension

ghost:
Invalid filename extension
Debian 9 error message:
N: Ignoring file 'foreman.repo' in directory '/etc/apt/sources.list.d/' as it has an invalid filename extension


ekohl:
Thanks!

PR #1215, opened 2018-10-23
Add upgrade and deprecation notices for 1.20 manual

tbrisker:


_includes/manuals/1.20/1.2_release_notes.md
```diff
@@ -350,5 +352,11 @@ same time. In order to prevent this in future, we decided to add a constraint on
 means that existing duplicates need to be unified manually. Either delete all duplicates from `puppetclasses`
 table or try using [foreman-maintain](https://github.com/theforeman/foreman_maintain) that should fix the issue.
 For more information, see [the redmine issue](http://projects.theforeman.org/issues/24923/)
+* A change introduced to the host fact API in 1.19.0 has led to incompatibilities and has been reverted.
+Any scripts using this API may need to be updated accordingly.
+See ([the redmine issue](https://projects.theforeman.org/issues/25188)) for more details.
+* The ability to define `puppetconfdir`, `puppetvardir` and `puppetssldir` via `settings.yaml` has
```
ekohl, line 395: The `ssl_*` settings mentioned can still be set via `settings.yaml` (which the installer does). It's that they previously were derived from `puppetssldir` if unset.

tbrisker: updated the text, let me know it it is clearer now. thanks!

_includes/manuals/1.20/1.2_release_notes.md
```diff
@@ -376,7 +375,10 @@ Foreman should now run properly in FIPS-enabled environments.
 
 ### Deprecations
 
+* The ability to disable login, organizations or locations via settings has been deprecated and will be
+removed in version 1.21. Please enable these settings in your `settings.yaml` file.
```
lzap, line 379: Can you elaborate what this effectively mean for users upgrading or installing 1.20? A warning messages in logs? I believe it's not enough and we should be more verbose, probably some persistent banner, they need to test it now, not after there is no way back. Any ideas?

tbrisker: There is a notification of this change pointing to the discussion thread on discourse that will be displayed to all admins for a month after the upgrade.

ekohl (approved):

PR #1217, opened 2018-10-30
Fix link to 1.20 repositories

iNecas:
Using `foreman_version: 1.20` without quotes caused the version to be converted
to `1.2`, as it was treated as number, instead of a string, leading to
instructing folks to install

    yum -y localinstall http://yum.theforeman.org/releases/1.2/el7/x86_64/foreman-release.rpm

### DIFF
--- a/plugins/katello/3.9/installation/index.md
+++ b/plugins/katello/3.9/installation/index.md
@@ -1,9 +1,9 @@
 ---
 layout: plugins/katello/documentation
 title: Katello Installation
-version: 3.9
-foreman_version: 1.20
-latest: 3.8
+version: '3.9'
+foreman_version: '1.20'
+latest: '3.8'
 script: osmenu.js
 ---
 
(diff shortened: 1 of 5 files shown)


ehelms (approved):

plugins/katello/3.9/installation/index.md
```diff
@@ -108,7 +108,7 @@ yum -y localinstall http://dl.fedoraproject.org/pub/epel/epel-release-latest-7.n
 yum -y install foreman-release-scl python-django
 {% endhighlight %}
 </div>
-
+d
```
ekohl: This looks unintended

ekohl (changes requested):

plugins/katello/3.9/installation/index.md
iNecas: Good catch! Sry, updated

ekohl (approved):

PR #1223, opened 2018-11-05
Setting trusted_puppetmaster_hosts was renamed to trusted_hosts

mavit:


ekohl (approved):
I'm wondering if we should add ```(formerly `trusted_puppetmaster_hosts`)``` since this we only started with this in 1.18.0. On the other hand, 1.17.0 is now unsupported. @tbrisker?

tbrisker:
I'm fine with this as that is the name of the setting in the supported versions. Thanks @mavit and @ekohl !

PR #1225, opened 2018-11-08
Update apidoc for 1.19.1 release

tbrisker:


api/1.19/apidoc/v2/architectures.fr.html
```diff
@@ -504,17 +504,22 @@ <h3>Exemples</h3>
     <pre class="prettyprint">POST /api/architectures
 {
   &quot;architecture&quot;: {
-    &quot;name&quot;: &quot;i386&quot;
+    &quot;name&quot;: &quot;&quot;
   }
 }
-201
+422
```
ekohl, line 510: This is an odd one that jumps out.

api/1.19/apidoc/v2/architectures.html
```diff
@@ -504,17 +504,22 @@ <h3>Examples</h3>
     <pre class="prettyprint">POST /api/architectures
 {
   &quot;architecture&quot;: {
-    &quot;name&quot;: &quot;i386&quot;
+    &quot;name&quot;: &quot;&quot;
   }
 }
-201
+422
```
ekohl, line 510: This doesn't look right. It removes the value for name and only shows an invalid example.

ekohl (changes requested):
My browser is not so happy with the diff ;)

tbrisker:
Hmmm... looking at the log looks like i might have a local environment issue running the 1.19 tests properly to generate the examples. Closing this PR, thanks for noticing @ekohl! any chance you could try to generate this on your side?

PR #1229, opened 2018-11-13
Update installation instructions for foreman_ansible

xprazak2:
I think the installer should be used rather than installing the packages. @dLobatog, does this seem ok to you?

### DIFF
--- a/plugins/foreman_ansible/2.x/index.md
+++ b/plugins/foreman_ansible/2.x/index.md
@@ -145,21 +145,13 @@ If this is the first plugin you're installing, please see the [plugin
 repository section]({{site.baseurl}}plugins/#2.2Packageinstallation) to set up the
 repository first.
 
-To install Foreman Ansible, run the following:
+You can use foreman-installer to install Foreman Ansible:
 
-* Red Hat based (RHEL, CentOS...)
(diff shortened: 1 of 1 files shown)


plugins/foreman_ansible/2.x/index.md
```diff
@@ -145,21 +145,13 @@ If this is the first plugin you're installing, please see the [plugin
...
 repository first.
 
-To install Foreman Ansible, run the following:
+You can use foreman-installer to install Foreman Ansible:
 
-* Red Hat based (RHEL, CentOS...)
+      # foreman-installer --enable-foreman-plugin-ansible --enable-foreman-proxy-plugin-ansible
 
-      # yum install tfm-rubygem-foreman_ansible
+Restarting Foreman might be needed after installing the package. To run Asnible playbooks as Job Templates, you will also need Remote Execution:
```
ekohl: Asnible -> Ansible. The yum package should also restart the foreman application. If not, I think that's a bug.

ekohl (changes requested):

xprazak2:
I fixed the typo and removed the hint to restart Foreman.

ekohl (approved):

PR #1233, opened 2018-11-16
Fixes #23642 - Update 1.16 - 1.19 installer manuals

ekohl:


ekohl:
Looks like I forgot the digitalocean exception and that link is now a 404.

dgoetz (commented):
Except from the Digital Ocean link you already mentioned, it looks good. Some values lose their defaults or another small bit of information, but I think nothing important. 

The most prominent example is `Add a file to /etc/sudoers.d (true) or uses augeas (false)` which will be shortened to `Add a file to /etc/sudoers.d (true)` where the or part was also not perfect as it would only tell an experienced user that it will use augeas to directly add a sudo rule to /etc/sudoers.

Thanks!

ekohl:
That's because it was changed in https://github.com/theforeman/puppet-foreman_proxy/blob/fde8d9b9547b1dc50b6010616b4aaf5bde51fbf4/manifests/init.pp#L68 so that's the actual help text. The fact it was changed in https://github.com/theforeman/puppet-foreman_proxy/commit/72dca5a10f409e80c32c1d416ae9bf0d61e44972 (almost 2 years old) suggests we haven't updated these help texts in a LONG time.

As you can see I've scripted most of this part already and will be updating the release procedure so we're reminded every time.

ekohl:
Updated. I've improved the logic in the script to autodetect plugin manuals and added an exception for digitalocean.

scripts/installer/installer-enable-to-table
```diff
@@ -0,0 +1,101 @@
...
+
+PREFIXES = {
+    '--[no-]enable-foreman-plugin-': 'foreman_',
+    '--[no-]enable-foreman-proxy-plugin-': 'smart_proxy_',
+}
+
+
+PLUGINS = {
+    'digitalocean': {
+        'url': 'https://github.com/theforeman/foreman-digitalocean',
```
ekohl, line 16: https://community.theforeman.org/t/move-foreman-digitalocean-plugin/11845

tbrisker (approved):
Thanks @ekohl and @dgoetz !

PR #1239, opened 2018-11-23
Changes incorporated from https://community.theforeman.org/t/esxi-dep…

amard33p:
…loyment-via-uefi-hook-scripts-issue/11794/11

1. In 1.19 hook data json structure is corrected. The bug is still open though!
2. after_build script needed to cd before editing boot.cfg.
3. kickstart parameters added to boot.cfg.


lzap (approved):
I think this is a reasonable change.

lzap:
Thanks!

PR #1241, opened 2018-11-26
Add CfgMgmt CFP to website

GregSutcliffe:
1 week left on the CFP, figured it should get a push :)

### DIFF
--- a/_includes/alerts.html
+++ b/_includes/alerts.html
@@ -1,5 +1,8 @@
-<!-- <div class='tease'>      -->
-<!--   <p>                    -->
-<!--     something important! -->
-<!--   </p>                   -->
-<!-- </div>                   -->
+<div class='tease'>
+  <p>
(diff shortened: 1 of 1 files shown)


_includes/alerts.html
```diff
@@ -1,5 +1,8 @@
-<!-- <div class='tease'>      -->
-<!--   <p>                    -->
-<!--     something important! -->
-<!--   </p>                   -->
-<!-- </div>                   -->
+<div class='tease'>
+  <p>
+  <a>Conference season is here! Join us at FOSDEM &amp; Config Management Camp, 2nd - 5th February 2019</a>
```
tbrisker: Shouldn't this be 6th?

GregSutcliffe: I guess that's the hackday rather than part of the conference proper, but I can amend if you think it's right.

GregSutcliffe: @tbrisker updated

tbrisker (approved):
Thanks!

PR #1243, opened 2018-11-26
Fix Katello User Guide PNGs

jturel:
Not sure what happened, but @ekohl pointed out that some images were broken and so this PR fixes them


ekohl (approved):
This is an improvement over broken files, but some still show the horizontal navigation so they deserve an update.

jturel:
Fair point! - Opened https://projects.theforeman.org/issues/25557 to track that for us

PR #1248, opened 2018-11-29
Treat foreman_version as a string

ekohl:
Otherwise it results in:

    yum install -y https://yum.theforeman.org/client/1.2/el5/x86_64/foreman-client-release.rpm


akofink (approved):
ACK!

mmoll:
merged, bedankt @ekohl!

PR #1249, opened 2018-11-29
Use a bundler cache in Travis

ekohl:


mmoll:
`There was an error in the .travis.yml file from which we could not recover.`

ekohl:
Yes, the cache is not simply a boolean but a string. Updated

mmoll:
merged, bedankt @ekohl!

PR #1256, opened 2018-12-10
Cleanup in discovery manuals

lzap:
Search engines often offers pretty outdated manuals in the results, we can stop that! This is a proposal, once we agree on solution I will expand this to other plugins and core manual. Discussion:

https://community.theforeman.org/t/proposal-delete-old-manuals-from-our-site/12140


lzap:
Ready for the review, I have the PDFs of deleted versions, want to put them somewhere.

GregSutcliffe:
The PDFs can be placed on downloads.theforeman.org - I think you have access? Let me know if not. Something like "/manuals/<pluginname|core>" should be fine, right?

Once uploaded, I'd like to see some handling of redirects for users. Right now it looks like the old versions are just deleted with no handling of where to look for the old manuals. Can we can some links to the files on downloads.theforeman.org? The user *shouldn't* have to go hunting for them.

lzap:
How about putting the PDFs directly into the git repo into `static/legacy-manuals`?

I was thinking creating a custom 404 page with note about legacy content and link.

lzap:
I can also create RewriteRule but the downside of that is we'd need to update it on every new stable release. I prefer 404 custom page.

lzap:
I think I can do both, here it is:

https://github.com/theforeman/foreman-infra/pull/887

Rebased with the 404 custom page and legacy static directory. The plan is to test this on the discovery plugin and once verified live, I will proceed with the same strategy for core.

GregSutcliffe:
Putting them into `static` would be fine too, and means they're "backed up" in the Git repo automatically. I'm OK with that if you prefer it.

lzap:
Yep, that's what I've figured out actually, see the web PR. It's now in `static/legacy` folder which seems like a great fit for that.

GregSutcliffe:
PDFs look OK, indeed, although it does include the current footer from the site (referencing FOSDEM 2019 :P). I think that'll do for a start.

The only thing I can see as an issue now is that there is no mention of the archived manuals on the manual index (e.g. https://theforeman.org/plugins/foreman_discovery) - IIRC deleting the old manuals will make them disappear. Could we leave the `plugins/foreman_discovery/$version/index.md` files, but with a custom layout that just redirects people to the PDF? Or something similar, anyway - I don't want users coming into the channel asking "Where is the manual for Discovery 6.0?" because the only way to discover it is to go to the URL in your history.

lzap:
Sure what you prefer? I can do stubs or leave a comment on the index page. I lean towards the latter since we can leverage the shiny new 404.html page which actually includes the link to the legacy part already so it's a good stub.

GregSutcliffe:
Yeah, I think a simple comment like "Older manuals can be found <a href=...>here</a>" would do, so long as the link goes to a list of the PDFs. The goal is to avoid a user having to type a URL in, so if I want Discovery 0.2, there should be a way to get to it without editing URLs :)

lzap:
Done, was not sure about article tho, in this case I put no article :-) 


    Older manuals are available <a href="/static/legacy/manuals">in PDF format</a>.

lzap:
@GregSutcliffe do you plan to finish this review? I would love to push this forward, once discovery is done and we agree how to proceed, I want to do the same for core, katello and other plugins.

lzap:
Oh I see now there is this foreman-infra patch which has these redirects (now only 404 document), not sure if this is blocked by that patch. However want to draw @ekohl 's attention to this PR as well. Our ./api folder is now 1.3 GiB in size and we need to do something, I want to follow the same pattern (creating tarballs of old versions).

lzap:
I am currently little block by this as I want to release discovery v15 docs, this will make conflicts. I see no problems merging this, anyone? @ekohl @ehelms @mmoll @tbrisker - want to bring more attention to this, this should (I hope) set new standard for deprecating old manual content.

mmoll:
yes, let's bring this in :ship:

mmoll:
merged, díky @lzap!

PR #1259, opened 2018-12-12
Fix not working sudoers example in plugins/foreman_salt

hbokh:
This will fix commands `salt` and `salt-key` to be run on a Foreman-proxy.

The current example(s) are incorrect and you'll find "HTTP 406 - Not Acceptable" messages in `foreman-proxy/proxy.log`, next to `sudo: foreman-proxy : command not allowed ; TTY=unknown ; PWD=/ ; USER=foreman-proxy ; COMMAND=/usr/bin/salt-key --finger-all --output=json` in `auth.log`.

### DIFF
--- a/plugins/foreman_salt/2.0/index.md
+++ b/plugins/foreman_salt/2.0/index.md
@@ -102,7 +102,7 @@ Install the Salt Smart Proxy package for your operating system (see above). The 
 Add this to /etc/sudoers:
 
     Cmnd_Alias SALT = /usr/bin/salt, /usr/bin/salt-key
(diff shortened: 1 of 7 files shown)


GregSutcliffe (approved):
Looks fine to me - although I'm surprised we haven't seen more reports of this. 

Thanks for your patience @hbokh - sorry for the long delay in reviewing. I'll get this merged now.

hbokh:
Thanks Greg! I think the sudo implementation on FreeBSD is a bit more picky...
The provided (previous) example might have worked on (some) Linux, but I **had** to add the extra ` (ALL) ` on BSD to get rid of the errors.

PR #1264, opened 2018-12-17
explicitly install/upgrade centos-release-scl-rh during upgrade

evgeni:
also clean only yum metadata cache, this should be sufficient

### DIFF
--- a/_includes/manuals/1.20/3.6_upgrade.md
+++ b/_includes/manuals/1.20/3.6_upgrade.md
@@ -38,17 +38,21 @@ appropriate foreman-release package for your operating system:
 
     yum upgrade https://yum.theforeman.org/releases/{{page.version}}/el7/x86_64/foreman-release.rpm
 
-Clean up the yum cache:
+Clean up the yum metadata cache:
 
-    yum clean all
+    yum clean metadata
 
-First install or upgrade the SCL release package to add the rh-ruby25 repository:
+First install or upgrade the SCL release packages to add the rh-ruby25, tfm and tfm-ror52 repositories:
 
-    yum install foreman-release-scl
+    yum install centos-release-scl-rh foreman-release-scl
 
 or:
 
-    yum upgrade foreman-release-scl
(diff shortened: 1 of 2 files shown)


tbrisker (approved):
Thanks @evgeni

_includes/manuals/1.20/3.6_upgrade.md
```diff
@@ -38,17 +38,21 @@ appropriate foreman-release package for your operating system:
...
+Clean up the yum metadata cache:
 
-    yum clean all
+    yum clean metadata
 
-First install or upgrade the SCL release package to add the rh-ruby25 repository:
+First install or upgrade the SCL release packages to add the rh-ruby25, tfm and tfm-ror52 repositories:
 
-    yum install foreman-release-scl
+    yum install centos-release-scl-rh foreman-release-scl
```
ekohl, line 47: `foreman-release-scl` depends on `centos-release-scl-rh` so please remove it from the instructions here.

_includes/manuals/1.20/3.6_upgrade.md
```diff
@@ -38,17 +38,21 @@ appropriate foreman-release package for your operating system:
...
 
 or:
 
-    yum upgrade foreman-release-scl
+    yum upgrade centos-release-scl-rh foreman-release-scl
```
ekohl, line 51: I'm not sure there's so much need to explicitly upgrade `centos-release-scl-rh`. If we need a certain version then `foreman-release-scl` should depend on it.

_includes/manuals/1.20/3.6_upgrade.md
evgeni: The idea here was to bring the user attention to the fact that we actually need more than *our* SCL. How would you rephrase that in text to drop the package name here?

_includes/manuals/1.20/3.6_upgrade.md
evgeni: Same as above.

PR #1277, opened 2019-01-10
adds storybook

amirfefer:
Adding the [ui storybook](http://theforeman.github.io/foreman/) into foreman documentation

### DIFF
--- a/_layouts/storybook.html
+++ b/_layouts/storybook.html
@@ -1,0 +1,11 @@
+---
+pagename: storybook
+---
+
+{% include header.html %}
+<div id="wrap">
+  <div id="content">
+    <iframe src="https://theforeman.github.io/foreman" frameborder="0" width="100%" height="1000px"></iframe>
+  </div>
+</div>
+{% include footer.html %}
--- a/documentation.md
+++ b/documentation.md
@@ -107,5 +107,11 @@ version: '1.20'
       Developer Handbook
     </a>
   </div>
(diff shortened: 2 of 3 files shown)


_layouts/storybook.html
```diff
@@ -0,0 +1,11 @@
+---
+pagename: storybook
+---
+
+{% include header.html %}
+<div id="wrap">
+  <div id="content">
+    <iframe src="https://theforeman.github.io/foreman" frameborder="0" width="100%" height="1000px"></iframe>
```
ekohl, line 8: Is an iframe really the best way? How about a short text that it's generated and we have a rendered version at that URL?

amirfefer: I'm not sure if I understand your suggestion, the storybook is deployed into foreman github page (https://theforeman.github.io/foreman) via storybook script, so afaik no dynamic rendering is possible here, and having a simple link to that page is IMO doesn't look better.
What are your concerns about using iframes? we use it in `API V2` page as well.

ekohl: Since you can't easily fully view an iframe I was thinking about a header text. What about adding something like this above it?

The documentation on UI development is maintained in the [https://github.com/theforeman/foreman](Foreman repository) as a [Javascript Storybook](https://storybook.js.org/). It can be viewed locally by using `npm run storybook` or [viewed online](https://theforeman.github.io/foreman).

IMHO the API V2 page deserves a similar header

PR #1283, opened 2019-01-16
Add warning that mysql should be version 5.7

MarkNReynolds:
Add warning for mysql version as discussed in https://community.theforeman.org/t/update-1-20-from-1-19-mysql-5-7-dependancy-warning/12570/3.


tbrisker (approved):
Thanks @MarkNReynolds !

PR #1285, opened 2019-01-17
Added upgrade vacuum step

lzap:
SSIA - I haven't tested those commands tho.

### DIFF
--- a/_includes/manuals/nightly/3.6_upgrade.md
+++ b/_includes/manuals/nightly/3.6_upgrade.md
@@ -169,7 +169,23 @@ And on Debian/Ubuntu installations with Passenger, run:
 
     service apache2 restart
 
-#### Optional Step 5 - Cleanup of RPMs and repositories
+#### Optional Step 5 - Reclaim database space
+
+After database migrations, some space can sometimes be reclaimed. It's a good idea to perform a *full* database vacuum for PostgreSQL rather than relying on the autovacuum feature to claim maximum space possible. For PostgreSQL do the following:
+
+    foreman-maintain service stop --exclude postgresql
+    su - postgres -c 'vacuumdb --full --dbname=foreman'
+    foreman-maintain service start
+
+MariaDB/MySQL users should do the similar:
+
+    foreman-maintain service stop --exclude mysql
+    mysqlcheck --optimize --all-databases
+    foreman-maintain service start
+
+On platforms where foreman-maintain is not available, stop all Foreman-related services except PostgreSQL or MySQL.
+
+#### Optional Step 6 - Cleanup of RPMs and repositories
 
 On EL6/7, consider removing the unmaintained softwarecollections.org
 repositories, which are replaced by the CentOS SCLorg repos:


_includes/manuals/nightly/3.6_upgrade.md
```diff
@@ -183,7 +183,23 @@ And on Debian/Ubuntu installations with Passenger, run:
...
+
+    foreman-maintain service stop
+    service postgresql start
+    su - postgres -c 'vacuumdb -f -d foreman'
+    foreman-maintain service start
+
+MySQL users should do the similar:
+
+    foreman-maintain service stop
+    service postgresql start
```
tbrisker: This should be mysql or mariadb i think

_includes/manuals/nightly/3.6_upgrade.md
```diff
@@ -183,7 +183,23 @@ And on Debian/Ubuntu installations with Passenger, run:
 
     service apache2 restart
 
-#### Optional Step 5 - Cleanup of RPMs and repositories
+#### Optional Step 5 - Reclaim database space
+
+After database migrations, some space can be often reclaimed. It's good idea to perform database vacuum for postgres on a regular basis as well. For PostgreSQL do the following:
```
tbrisker: s/be often/sometimes be/

_includes/manuals/nightly/3.6_upgrade.md
lzap: Ok going for `mysql` in this case.

lzap:
Done.

_includes/manuals/nightly/3.6_upgrade.md
```diff
@@ -183,7 +183,23 @@ And on Debian/Ubuntu installations with Passenger, run:
 
     service apache2 restart
 
-#### Optional Step 5 - Cleanup of RPMs and repositories
+#### Optional Step 5 - Reclaim database space
+
+After database migrations, some space can sometimes be reclaimed. It's good idea to perform database vacuum for postgres on a regular basis as well rather than relying on autovacuum feature. For PostgreSQL do the following:
```
sean797: Postgres supports "auto-vacuuming" https://www.postgresql.org/docs/9.5/routine-vacuuming.html. I think its generally configured out the box and turned on by default?

lzap: Isn't "rather than relying on autovacuum feature" enough of a statement? What exactly do you want me to elaborate on? Doing the manual vacuum is still an option after migration which can drop some columns or indices.

sean797: Sorry I completely missed that the first time I read it. Time to get some coffee.

chris1984 (changes requested):
Some Grammer nitpicks other than that it looks good, I would rewrite this line like so 
```
After database migrations, some space can sometimes be reclaimed. It's a good idea to perform a database vacuum for Postgres on a regular basis as well rather than relying on the autovacuum feature.
```

_includes/manuals/nightly/3.6_upgrade.md
```diff
@@ -183,7 +183,23 @@ And on Debian/Ubuntu installations with Passenger, run:
...
+
+    foreman-maintain service stop
```
tbrisker: Do we now ship foreman-maintain by default with foreman? thinking about it, perhaps this procedure should actually go into f-m? @mbacovsky

mbacovsky: @ekohl could you please put some light on how we ship f-m with Foreman currently?
@tbrisker, goos idea. I filed new RFE to improve upgrades in f-m https://projects.theforeman.org/issues/25928

ekohl: We have it in the repositories but currently nothing in vanilla Foreman assumes it's being used. It's also possible to do an online vacuum so I don't see why it's needed to stop all services.

_includes/manuals/nightly/3.6_upgrade.md
```diff
@@ -183,7 +183,23 @@ And on Debian/Ubuntu installations with Passenger, run:
...
+    service postgresql start
+    su - postgres -c 'vacuumdb -f -d foreman'
+    foreman-maintain service start
+
+MariaDB/MySQL users should do the similar:
+
+    foreman-maintain service stop
+    service mysql start
+    mysqlcheck -o --all-databases
```
ekohl: In documentation I'd always advise long options so `--optimize`

_includes/manuals/nightly/3.6_upgrade.md
lzap: This is VACUUM FULL (see the `-f` option) which requires exclusive lock and it's better when database is not under heavy load. I can probably point out that in the text.

_includes/manuals/nightly/3.6_upgrade.md
lzap: Good point will do the same for pg.

lzap:
Done, grammar, full options, added "full" word.

chris1984 (approved):
LGTM @lzap

_includes/manuals/nightly/3.6_upgrade.md
ekohl: AFAIK there's also `--exclude` so you don't need to start postgresql afterwards. That said, I'd split it. First just state how to do a vacuum, then that foreman-maintain can be used to stop the services to improve the load issue.

lzap: Now I don't understand what you want me to do.

evgeni: `foreman-maintain service stop --exclude <database>` :)

_includes/manuals/nightly/3.6_upgrade.md
```diff
@@ -183,7 +183,23 @@ And on Debian/Ubuntu installations with Passenger, run:
 
     service apache2 restart
 
-#### Optional Step 5 - Cleanup of RPMs and repositories
+#### Optional Step 5 - Reclaim database space
+
+After database migrations, some space can sometimes be reclaimed. It's a good idea to perform a *full* database vacuum for Postgres rather than relying on the autovacuum feature to claim maximum space possible. For PostgreSQL do the following:
```
evgeni: It's called PostgreSQL, not Postgres :)

evgeni: "for postgresql…" should probably be an own paragraph.

lzap: Well it is accepted alias, it was a main name candidate for some time actually but the idea got dropped eventually: https://wiki.postgresql.org/wiki/ProjectName

But I will use the official name, not looking good to mix them in a single doc.

lzap:
Rebased, man we have many "Postgresql" and "Postgres" mainly in Katello docs :-)

evgeni (changes requested):
pending `foreman-maintain` optimization :)

lzap:
Done thanks.

_includes/manuals/nightly/3.6_upgrade.md
```diff
@@ -169,7 +169,21 @@ And on Debian/Ubuntu installations with Passenger, run:
...
+
+    foreman-maintain service stop --exclude postgresql
+    su - postgres -c 'vacuumdb --full --dbname=foreman'
+    foreman-maintain service start
+
+MariaDB/MySQL users should do the similar:
+
+    foreman-maintain service stop --exclude mysql
+    mysqlcheck --optimize --all-databases
+    foreman-maintain service start
```
mmoll, line 184: I think `foreman-maintain` is not available on DEB based installs, so it would be nice to rephrase that, although I'm not 100% sure, what would be corect then.

lzap: Done, it is also not available on gem installs.

lzap:
Rebased.

lzap:
Ping? Anyone?

_includes/manuals/nightly/3.6_upgrade.md
```diff
@@ -169,7 +169,23 @@ And on Debian/Ubuntu installations with Passenger, run:
...
+
+On platforms where foreman-maintain is not available, stop all Foreman-related services except postgres or mysql.
```
evgeni:
```suggestion
On platforms where foreman-maintain is not available, stop all Foreman-related services except PostgreSQL or MySQL.
```

evgeni (approved):
ACK pending my usual naming nitpicks ;)

_includes/manuals/nightly/3.6_upgrade.md
lzap: *red face* OK :-)

lzap:
Rebased.

mmoll:
merged, díky @lzap!

PR #1288, opened 2019-01-17
Katello 3.11 Docs

jturel:
TODO:

- [x] Add release notes
- [x] Add API docs


johnpmitsch:
From the installation page, is https://fedorapeople.org/groups/katello/releases/yum/3.11/katello/el7/x86_64/katello-repos-latest.rpm expected to work currently?

ekohl:
> From the installation page, is https://fedorapeople.org/groups/katello/releases/yum/3.11/katello/el7/x86_64/katello-repos-latest.rpm expected to work currently?

No, only after the release has gone through the pipeline and verified to work it's pushed to the public location.

johnpmitsch:
The URL doesn't look correct for the client installation
`yum install -y https://yum.theforeman.org/client//el7/x86_64/foreman-client-release.rpm`, will that also be fixed when things are out of the RC phase?

johnpmitsch:
Other than that open question, things look good! :+1:

jturel:
@johnpmitsch nice catch. The page was missing foreman_version. If you leave the container running and just pull down the changes it'll regenerate the HTML (saves time from bringing up the container again)

johnpmitsch:
The client links look much better now, but I am not able to access the sles11 and 12 links? i.e. https://yum.theforeman.org/client/1.21/sles11/x86_64/foreman-client-release.rpm, it looks like that rpm is not in that folder. I'm not sure if that is something that happens later, but figured I would ask since it is there for other distros

jturel:
I see what you mean. Looks like the same problem is present for 1.20 as well - for that reason I'd like to treat that as a separate problem especially since I don't know how to fix it at the moment. Will make a note

johnpmitsch (approved):
Things look good aside from the SLES links, which sound like they are being addressed separately

ekohl:
The problem is that our CI pipeline doesn't publish sles: https://ci.theforeman.org/blue/organizations/jenkins/foreman-client-1.21-rpm-pipeline/detail/foreman-client-1.21-rpm-pipeline/3/pipeline (same for 1.20 I think because it was copied).

_includes/plugins/katello/sidebar_3.11.html
```diff
@@ -0,0 +1,180 @@
...
+    <li class="item">
+      <a href="/plugins/katello/3.11/upgrade/index.html">2.1 Katello</a>
+    </li>
+    <li class="item">
+      <a href="/plugins/katello/3.11/upgrade/smart_proxy.html">2.2 Smart proxy</a>
+    </li>
+    <li class="item">
+      <a href="/plugins/katello/3.11/upgrade/clients.html">2.3 Clients</a>
+    </li>
+    {% if page.version == 'nightly' or page.version == 3.2 %}
```
evgeni: I think we can drop the puppet upgrade that was needed in 3.2 now? :)

evgeni: (doesn't have to be in this PR, we'd be updating other versioned pages too then anyways)

_includes/plugins/katello/sidebar_3.11.html
```diff
@@ -0,0 +1,180 @@
+<ul class="header nav nav-list bs-docs-sidenav">
+  <li class="item">
+    <a>1. Installation</a>
+  </li>
+
+  <ul class="sub-header-static nav nav-list bs-docs-sidenav">
+    <li class="item">
+      <a href="/plugins/katello/3.11/installation/index.html">1.1 Katello</a>
```
ekohl: Every use of `3.11`: can it be `{{ page.version }}` as well?

jturel: fixed for 3.11 and nightly

_includes/plugins/katello/sidebar_3.11.html
jturel: removed this check for 3.11 and nightly & removed upgrades/puppet.md for > 3.2

ehelms (approved):

jturel:
@ehelms care to merge? :)

PR #1291, opened 2019-01-21
allow easier selection of os-specific sections

evgeni:
* use css classes for selection, allowing multiple sections to be marked
  as "to show" or "to hide"
* the above requires the "show/hide" loop to be split in two:
  first hide everything, then show what the author selected
* use this feature to hide tracer on EL5/6 and yum instructions on SUSE

### DIFF
--- a/plugins/katello/nightly/installation/clients.md
+++ b/plugins/katello/nightly/installation/clients.md
@@ -77,6 +77,7 @@ Now you are ready to install the client package:
 
 The `katello-host-tools` package reports errata & package profile information, but does not allow you to run remote actions on the clients.
 
+<div class="el5 el6 el7 f27 f28">
 {% highlight bash %}
 yum install katello-host-tools
 {% endhighlight %}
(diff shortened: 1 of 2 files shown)


evgeni:
review note: before merging, I'd like to apply the same change to katello 3.8/3.9/3.10 docs, not only nightly, but was too lazy before the review :)

plugins/katello/nightly/installation/clients.md
```diff
@@ -87,17 +88,22 @@ We generally recommend using Foreman Remote Execution or Ansible for remote acti
...
 Optionally you can also install `katello-host-tools-tracer` and the client will report processes that need restarting after an update back to the Katello server.
 
 {% highlight bash %}
 yum install katello-host-tools-tracer
 {% endhighlight %}
+</div>
+</div>
 
-For Suse Clients, only katello-host-tools is supported:
+<div class="sles11 sles12" style="display:none;" markdown="1">
```
ekohl, line 100: What's the `markdown=1` part?

evgeni: I don't know, I copied it over from the other sections.

I *think* it indicates that the content is markdown and not plain HTML

ekohl (approved):

ekohl:
Of course I read your note the first time I reviewed but not the second time ...

PR #1294, opened 2019-01-28
add foreman-ansible-modules docs

evgeni:


ehelms:
I can see this approach working just fine. My question is should we put it on this page `https://theforeman.org/documentation.html` similar to how we have an API link?

evgeni:
@ehelms for now I have added is to the "plugins" page, even though it's not strictly a plugin (foreman_maintain is also listed there). "documentation" felt a bit too toplevel, but I am fine with either placement as long as we can get the docs online.

ehelms (approved):

PR #1304, opened 2019-02-07
Don't rely on display:none

jturel:
We were missing a `display:none` on the block showing the katello-host-tools-tracer div which did not guarantee that it would be hidden by osmenu.js unless an OS which should hide that data was selected by default.

I've changed osmenu.js so it will trigger not just on OS dropdown change, but when the page has loaded so display:none is no longer needed at all

### DIFF
--- a/plugins/katello/3.9/installation/clients.md
+++ b/plugins/katello/3.9/installation/clients.md
@@ -27,14 +27,14 @@ Install the appropriate Katello client release packages.
   </select>
 </p>
 
-<div id="el5" style="display:none;" markdown="1">
+<div id="el5" markdown="1">
 {% highlight bash %}
 yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-5.noarch.rpm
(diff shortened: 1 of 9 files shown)


jturel:
This problem was raised by https://projects.theforeman.org/issues/25886

plugins/katello/3.9/installation/clients.md
```diff
@@ -88,6 +88,7 @@ We generally recommend using Foreman Remote Execution or Ansible for remote acti
 {% highlight bash %}
 yum install katello-agent
 {% endhighlight %}
+</div>
```
jturel, line 91: Pretty sure this was a misplaced closing tag so I fixed in all the installation.md in this PR

evgeni: It wasn't misplaced as such, my intention was to have the tracer section *inside* the general k-agent one. But it doesn't really matter technically, as long as the hide/show stuff works.

evgeni (approved):

jturel:
thx evgeni :)

PR #1309, opened 2019-02-14
Move sponsors to former sponsors

ekohl:
I no longer see their resources in Foreman/Jenkins. @mmoll @evgeni?


mmoll:
Correct, maybe the sentences should also in past tense, then?

evgeni:
I have not much knowledge about sponsored stuff, but I agree I don't see any Jenkins nodes that I would associate with these names.

ekohl:
> Correct, maybe the sentences should also in past tense, then?

I thought about it, but then forgot to follow through. Will update.

ehelms (approved):

PR #1311, opened 2019-02-18
Remove puppet upgrade page

johnpmitsch:
on recent Katello versions, this page no longer exists and isn't
relevant, so we should remove the link since it 404s

### DIFF
--- a/_includes/plugins/katello/sidebar_3.10.html
+++ b/_includes/plugins/katello/sidebar_3.10.html
@@ -31,9 +31,6 @@
     <li class="item">
       <a href="/plugins/katello/3.10/upgrade/clients.html">2.3 Clients</a>
     </li>
-    <li class="item">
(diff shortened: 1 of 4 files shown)


johnpmitsch:
Judging from [this community discussion](https://community.theforeman.org/t/katello-3-10-or-katello-3-9-documentation-puppet-upgrade-instruction/12825/9?u=john_mitsch), we maybe should include more info on how puppet 4 to 5 upgrade is automatically performed. I think that is important to communicate to the user since they are asking about it.

But since these are 404ing currently, it may be best to do that in a follow up PR :)

Let me know your thoughts!

jturel (approved):
Looks good! However when I removed the upgrade docs for puppet I was too greedy. I can open a PR which restores currently 404'ing upgrade docs for releases where it's relevant .. 3.6 and below

johnpmitsch:
@jturel sounds good, just double check the relevant releases, I'm not sure if its relevant in 3.6, but I just removed for the most recent releases.

PR #1322, opened 2019-02-27
Add nightly headline feature - type support in Global Parameters

kgaikwad:
Added details about ["Type support in Global Parameters"](https://github.com/theforeman/foreman/pull/5241) under headline feature section and about searching changes under upgrade warning.

### DIFF
--- a/_includes/manuals/nightly/1.2_release_notes.md
+++ b/_includes/manuals/nightly/1.2_release_notes.md
@@ -4,6 +4,11 @@ This section will be updated prior to the next release.
 
 ### Headline features
 
+#### Array, Boolean and other types support in Parameters
+
+Global parameters now support the same types as smart variables and smart class parameters.
+As a part of this change searching by parameter value is no longer allowed. ([#4127](https://projects.theforeman.org/issues/4127))
+
 ### Features
 
 ### Bug Fixes
@@ -11,5 +16,7 @@ This section will be updated prior to the next release.
 ### Deprecations
 
 ### Upgrade warnings
+
+* With types support in global parameters, searching by parameter value is no longer allowed.
 
 ### Contributors


_includes/manuals/nightly/1.2_release_notes.md
```diff
@@ -4,6 +4,11 @@ This section will be updated prior to the next release.
 
 ### Headline features
 
+#### Array, Boolean and other types support in Parameters
+
+Not only string but array, boolean and other types support has been introduced in parameters like puppet parameters.
```
ares: I would change `like puppet parameters` to `, it now supports the same types as smart variables and smart class parameters`.

evgeni (approved):

_includes/manuals/nightly/1.2_release_notes.md
```diff
@@ -4,6 +4,11 @@ This section will be updated prior to the next release.
 
 ### Headline features
 
+#### Array, Boolean and other types support in Parameters
+
+Global parameters now support the same types as smart variables and smart class parameters.
+As a part of this change search by parameters has been modified so that instead of parameter value, searching based on parameter name is allowed. ([#4127](https://projects.theforeman.org/issues/4127))
```
ekohl: I think searching by name was already allowed so it's probably clearer to state that searching by parameter value is no longer allowed.

ekohl (commented):
Thank you for adding this to the headline features already. Makes the release work easier :)

tbrisker:
Should we also update the manual itself with this change? Doesn't have to be the same PR, but would be good to document it as well.

ares:
sounds good to me too now

ekohl (approved):
It is one old bug. Thanks for fixing it

evgeni:
@kgaikwad I guess you're doing the doc update that @tbrisker suggested in a separate PR? or shall we wait with merging this one?

kgaikwad:
@evgeni, 
I would prefer a separate PR for doc update.

evgeni:
@kgaikwad cool, then merged this one and awaiting next :)

PR #1324, opened 2019-02-27
Update manual with taxonomy change

tbrisker:


tbrisker:
@alexjfisher care to take a look since you mentioned we could improve the documentation around this? I'd be happy if you have any further suggestions.

evgeni:
:+1: but waiting on @alexjfisher :)

alexjfisher (approved):

alexjfisher:
The manual changes look good.

I think a blog post, or youtube video or something showing how to get the most out of the taxonomies would be useful.

At the moment, there are lots of gotchas that I think, despite the documentation, will catch people out. Most recently, I cloned one of the default provisioning templates.  Even though the original had taxonomies set, the clone didn't.  It was still possible to associate the clone with an operating system, but a host being built in the default loc/org couldn't download it during provisioning.

Maybe in 1.22 it won't be possible to create resources without taxonomies and they will automatically default if you only have 1 organisation and 1 location.

evgeni (approved):

PR #1325, opened 2019-02-27
Update 1.21 contributor list

tbrisker:


scripts/committers.rb
```diff
@@ -3,8 +3,8 @@
 require 'git'
 
 @tagged_repos = %w(foreman smart-proxy foreman-installer foreman-selinux)
-@tag_from     = "1.19.0"
-@tag_to       = "1.20.0-RC2"
+@tag_from     = "1.20.0"
```
ekohl, line 6: Does this mean the above list is the 1.20 committers?

tbrisker: no, it checks the contributors between the 1.20.0 and the 1.21.0 tags for these repos

PR #1326, opened 2019-02-28
Correct plugin installation instructions

tbrisker:


ekohl:
I did hijack this PR a bit to add https links in the page.

PR #1334, opened 2019-03-12
Updates to upgrade troubleshooting section

jschulthess:
Certificate listing in Foreman web GUI fails if sudo rights of local foreman-proxy user are either incorrect or not read. This documents adds troubleshooting documentation for both.

### DIFF
--- a/_includes/manuals/1.21/3.6_upgrade.md
+++ b/_includes/manuals/1.21/3.6_upgrade.md
@@ -184,3 +184,19 @@ And consider removing unused SCL packages:
 
 See
 [Troubleshooting](http://projects.theforeman.org/projects/foreman/wiki/Troubleshooting)
+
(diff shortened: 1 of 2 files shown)


tbrisker:
Thank you for the PR! would it be perhaps better to place this under section 4.3.7 which discusses the PuppetCA configuration or a new section under section 7 - Troubleshooting?

jschulthess:
True. Both of these places would also make sense. I added it under the upgrade as that was the point in time I encountered the problem.

jschulthess:
Maybe it would be better to have a section under 7 as for example the upgrade section has a reference to the troubleshooting section already.

tbrisker:
@jschulthess either one works for me - to add a subsection to section 7 you'll need to rename a few files and also modify the index at manual/nightlies/index.md.

tbrisker:
ping @jschulthess - any chance you can update this?

lzap:
Thank you for your contribution, I have moved the content to: https://projects.theforeman.org/projects/foreman/wiki/ERF50-5345

PR #1338, opened 2019-03-27
Clarify the language used to describe template-based provisioning without SSH

### LINKED ISSUE #1340: Language used to describe VMware "Image provisioning without SSH" needs clarification
I'm trying to understand how "Image provisioning without SSH" works with VMware and vSphere, and found the description confusing. For example:

* The instructions say "Enable user_data checking the proper checkbox in the compute resource definition", but the checkbox applies to the *image* used within the compute resource, not the compute resource itself.
* Instructions say "Associate a user_data template to the host, using cloud-init syntax", which sort of implies that I use cloud-init to do the association.

I'll submit a PR with my suggested fixes, following the suggestion from https://community.theforeman.org/t/provision-hosts-without-ssh-using-vmware-user-data-template/13406/2?u=stefan_lasiewski

stefanlasiewski:
Fixes #1340


ares (approved):
Awesome, thanks @stefanlasiewski, merging!

PR #1348, opened 2019-04-11
Make sure Katello 3.11 is the latest

jturel:
This gets rid of "These instructions are for installing Katello 3.11, but the latest stable is 3.11. "

### DIFF
--- a/plugins/katello/3.11/installation/index.md
+++ b/plugins/katello/3.11/installation/index.md
@@ -1,7 +1,7 @@
 ---
 layout: plugins/katello/documentation
 title: Katello Installation
-version: 3.11
+version: '3.11'
 foreman_version: 1.21
 latest: '3.11'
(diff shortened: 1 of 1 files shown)


jturel:
@tbrisker thanks for pointing this out

plugins/katello/3.11/installation/index.md
```diff
@@ -3,7 +3,7 @@ layout: plugins/katello/documentation
 title: Katello Installation
 version: 3.11
```
ekohl: How about:

```suggestion
version: '3.11'
```

The benefit would be that once you reach 3.20 you won't need to change it then. Otherwise it'll be interpreted as 3.2.

jturel: You got me there. Updated!

ekohl (approved):

PR #1357, opened 2019-04-23
Add CVE-2019-3845

jturel:


jturel:
@tbrisker not sure if we should merge this in considering we don't have the fix actually released yet.

PR #1360, opened 2019-04-24
Update puppet version requirements

tbrisker:


_includes/manuals/1.22/1.2_release_notes.md
```diff
@@ -384,5 +384,6 @@ The Smart Proxy now provides a v2 features API allowing proxy plugins to declare
 
 * With types support in global parameters, searching by parameter value is no longer allowed.
 * Logger named `blob` where template contents is rendered caused excessive log file growth and has been disabled by default. Enable it via `settings.yaml` or foreman-installer if needed.
+* Puppet 4 support has been dropped from the installer. Please use Puppet 5 or newer for installing or upgrading Foreman. Clients running Puppet 4 should still be able to report to Foreman.
```
ekohl: I'd like to deprecate Puppet 4 client support. How about:
```suggestion
* Puppet 4 support has been dropped from the installer. Please use Puppet 5 or newer for installing or upgrading Foreman. Clients running Puppet 4 should still be able to report to Foreman. Please note Puppet 4 is EOL since December 2018 and Puppet 4 client support in Foreman is now deprecated.
```

tbrisker: works for me

ekohl: Should we update the table as well from Supported to Deprecated?

tbrisker:
updated the table as well

ekohl (approved):
I'll add Puppet 6 in a separate PR

PR #1361, opened 2019-04-24
Add docs for foreman_ansible 3.x

xprazak2:
Draft of docs for the new foreman_ansible version.


plugins/foreman_ansible/3.x/index.md
```diff
@@ -0,0 +1,307 @@
+---
+layout: plugin
+title: Foreman Ansible 3.x Manual
+version: 3.x
+---
+
+# 1. Basics
+
+Foreman Ansible allows you to import hosts via Ansible, along with facts about these hosts and reports of their playbook runs. This plugin 3.x series should be compatible with any version of Foreman above 1.22, including 1.22. Foreman Ansible relies on [Foreman Remote Execution](plugins/foreman_remote_execution) in order to be able to run playbooks remotely.
```
ekohl:
```suggestion
Foreman Ansible allows you to import hosts via Ansible, along with facts about these hosts and reports of their playbook runs. This plugin 3.x series should be compatible with Foreman 1.22 and newer. Foreman Ansible relies on [Foreman Remote Execution](plugins/foreman_remote_execution) in order to be able to run playbooks remotely.
```

plugins/foreman_ansible/3.x/index.md
```diff
@@ -0,0 +1,307 @@
...
+  } else {
+    $(".ansible_version_none").show();
+  }
+}
+</script>
+
+In order to make Ansible send us data from the hosts, we set up a callback on your host that runs Ansible.
+To provide specific installation instructions, please select your Ansible version:
+
+<select onChange="update_ansible_version(this);">
```
ekohl: IMHO we should hard require Ansible 2.5+ so there's only a single set of instructions. Reading these instructions it may even be 2.2+ since those instructions are the same.

plugins/foreman_ansible/3.x/index.md
```diff
@@ -0,0 +1,307 @@
...
+    FOREMAN_SSL_KEY: The private key when using SSL client certificates (default  "/etc/foreman/client_key.pem")
+  </li>
+  <li>
+    FOREMAN_SSL_VERIFY: Use *False* to disable certificate checks. You can also set it
+    to use the local CA bundle "True", or a path to a CA file for a custom CA.
+  </li>
+</ul>
+
+<p>
+Since the 'foreman' and 'foreman-proxy' are 'no-login' users, you should set these in the <code>EnviromentFile</code> used by systemd, <code>/etc/sysconfig/foreman</code> and <code>/etc/sysconfig/foreman-proxy</code>. An alternative for other users is to put them in their <code>.bashrc</code> as <code>export FOREMAN_URL=yoururl</code>. In most distributions environment variables set in <code>/etc/environment</code> will be shared among all users.
```
ekohl: We no longer support running directly from Foreman so `<code>/etc/sysconfig/foreman</code>` is no longer needed

plugins/foreman_ansible/3.x/index.md
```diff
@@ -0,0 +1,307 @@
...
+
+```yaml
+template: src=foo.cfg.j2 dest={% raw %}{{ remote_install_path }}{% endraw %}/foo.cfg
+```
+
+uses the `remote_install_path` variable. You can use Foreman to override these variables. The variables will only be overridden if you run your roles from Foreman, as Foreman sends an inventory with a dictionary contianing variable names and values when it runs Ansible roles.
+
+There are several ways to override an Ansible variable:
+
+1. Import the variable following [this section of the manual](plugins/foreman_ansible/2.x/index.html#4.1.2ImportingVariables). Once the variable is imported, you may change the default value:
```
ekohl: Any reason this refers to the 2.x manual?

plugins/foreman_ansible/3.x/index.md
```diff
@@ -0,0 +1,307 @@
...
+# 5. Help
+
+Please follow our [standard procedures and contacts]({{site.baseurl}}support.html).
+
+## 5.1 Troubleshooting
+
+Errors regarding the import of facts and reports will show up on /var/log/foreman/production.log of your Foreman host.
+
+If you find a bug, please file it in
+[Redmine](http://projects.theforeman.org/projects/ansible/issues/new).
```
ekohl:
```suggestion
[Redmine](https://projects.theforeman.org/projects/ansible/issues/new).
```

ekohl (changes requested):
Looks like the PNGs are much larger. Have you used a tool like pngcrush to minimize them?

xprazak2:
I made changes as suggested. I also crushed the pngs.

tbrisker:
What is the status here @xprazak2 @ekohl ?

ekohl (approved):

PR #1362, opened 2019-04-24
More 1.22 manual updates

tbrisker:


scripts/release_notes.rb
```diff
@@ -73,7 +73,7 @@ def gather_issues(offset = 0)
   puts "#### #{category}"
 
   issues.each do |issue|
-    puts "* #{issue['subject'].gsub('`','\\\`')} ([##{issue['id']}](#{URL}/issues/#{issue['id']}))"
+    puts "* #{issue['subject'].gsub('`','\\\`').gsub('<','&lt;').gsub('>','&gt;')} ([##{issue['id']}](#{URL}/issues/#{issue['id']}))"
```
ekohl, line 76: Perhaps we need a HTML and Markdown escape function at some point :)

PR #1363, opened 2019-04-25
Update Katello GPG Keys docs

jturel:
GPG Keys in Katello were broadened in scope to include SSL Certificates - so a general term 'Content Credentials' was used to describe them. The docs changes reflect that & I added a redirect from /plugins/katello/nightly/user_guide/gpg_keys to the new content credentials docs. I also refreshed all of the screenshots since they were way old!

See the README for how to test this locally w/ docker


chris1984 (approved):
Looks good

tbrisker (approved):
Thanks @jturel !

PR #1364, opened 2019-04-29
Add 1.22 API Doc

tbrisker:
I did notice a few failures during the tests when generating the examples so there may be a few that are wrong.


lzap:
I am good however api/ is now 1.3 GiB. I suggest to drop old versions. I can do this probably in separate PR.

lzap:
The link on http://localhost:4000/documentation.html still points to 1.21.

lzap:
I take it back, it works when you select the version from the dropdown.

PR #1367, opened 2019-05-03
Katello 3.11 is the current latest

jturel:


jturel:
@chris1984 once 3.12 is GA we'll need a PR like this to update the 'latest' version

ehelms (approved):

PR #1373, opened 2019-05-16
Show latest doc for openscap

xprazak2:


lzap (approved):
Which reminds me the same for my own PR :-)

PR #1380, opened 2019-06-03
Bump Katello 3.12 to RC2

chris1984:


chris1984:
@jturel can I get a review, please? I spun up a 3.12 box as well and did some basic tests and looked over a few of the pages and everything works as well so I think we are good for the announcement as well.

tbrisker (approved):
Thanks @chris1984 !

PR #1388, opened 2019-06-07
Updated compatibility table for 1.22

lzap:
Also fixed version for 1.21 version which was not correct.


tbrisker (approved):
Thanks @lzap !

PR #1393, opened 2019-06-19
Add version selector to the quickstart page

Ron-Lavi:
Selection_141

cc @lzap


_config.yml
```diff
@@ -76,25 +76,25 @@ plugins:
...
-  - 1.6
-  - 1.5
-  - 1.4
-  - 1.3
-  - 1.2
-  - 1.1
\ No newline at end of file
+  - "1.22"
+  - "1.21"
+  - "1.20"
```
Ron-Lavi, line 81: As a number the outcome is `1.2` instead of `1.20`

_layouts/manual.html
```diff
@@ -1,6 +1,7 @@
 ---
 pagename: Manual
 latestversion: "1.22"
+url_suffix: "index"
```
lzap: Mind naming this "version_selector_target" instead? This feels like some global variable.

Ron-Lavi: updated :)

lzap:
Nice one more thing, on the quickstart guide I see a warning _This quickstart guide is for Foreman 1.22, but 1.20 is the current stable version._ and the reason is that both are using a layout variable `layout.latestversion` which are set in two files separately. Can you drop this variable and start using `foreman_version` from the main config? This will simplify things a bit.

lzap:
You know what, here is the patch: https://github.com/theforeman/theforeman.org/pull/1396

Ron-Lavi:
> on the quickstart guide I see a warning This quickstart guide is for Foreman 1.22, but 1.20 is the current stable version.

Thanks ! I also noticed that the `url` is incorrect there, directing to the manuals page instead of the quickstart, fix here: https://github.com/theforeman/theforeman.org/pull/1394

PR #1399, opened 2019-06-25
fix dead links to documentation.html

evgeni:


ekohl (approved):
I don't like there's no `documentation.html` anymore without even a redirect. That's bad for SEO. Will submit a PR.

PR #1400, opened 2019-06-25
Add a proper redirect on documentation.html

ekohl:
This page used to exist. Just making it a 404 is bad, especially since it's the second hit on Google for 'foreman documentation'. This redirects it to the quickstart guide.

### DIFF
--- a/documentation.html
+++ b/documentation.html
@@ -1,0 +1,8 @@
+---
+layout: default
+redirect: /manuals/latest/quickstart_guide.html
+---
+
+# Documentation
+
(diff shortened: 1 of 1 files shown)


zjhuntin:
Is there a way that we could just have the page redirect to the proper documentation page instead of having a page to tell them to go there?

ekohl:
Actually it does a redirect via javascript. If you go to https://theforeman.org/api.html you'll see you end up at the real documentation. It's not the prettiest but it's better than a 404.

zjhuntin (approved):
Ah okay, great, ACK from me.

PR #1403, opened 2019-06-27
Remove broken EL6 references from Katello CLI

chris1984:
Yum repos do not exist at the urls so removing support going back to 3.11


chris1984:
Yum repos do not exist at the urls so removing support going back to 3.11

evgeni (approved):

PR #1404, opened 2019-06-28
Post: dropping MySQL

lzap:
Do **not** comment on grammar or typos, please **checkout and push changes** into my branch. If there are many comments I will ask you to do so :-)

Thanks.

### DIFF
--- a/_includes/manuals/nightly/3.7_migration.md
+++ b/_includes/manuals/nightly/3.7_migration.md
@@ -1,0 +1,204 @@
+
+### Migrating to PostgreSQL
+
+Starting from Foreman 1.25, MySQL is no longer a supported database
+management system. This chapter describes the process of migrating from MySQL
+to PostgreSQL.
+
+Full backup is absolutely necessary before starting the process. We recommend
+both database dump and snapshot of whole operating system and data if possible
+when running in a VM or on a LVM.
+
+#### Prerequisites
+
+It's a good idea to delete unwanted records from the database. The smaller the
+database is, the faster the migration process will be. Typical offenders are
+audit records, facts and reports. Foreman ships with rake tasks and cron job to
+execute these cleanup tasks, but they can be executed manually. Steps are shown
+down below after all instances are stopped.
+
+Foreman does not support plugin uninstallation, any leftovers in the database
+might cause issues during migration.
+
+#### Preparing the environment
+
+The first step is to install required dependencies:
+
+    yum install foreman-postgresql
+
+or
+
+    apt-get install foreman-postgresql
+
+Next, stop all your foreman instances and make sure, they stay stopped during
+the migration. On Red Hat compatible systems, you need to stop the `httpd` and
+(if applicable) `foreman-tasks` service. On EL7, there is a helper utility
+called `foreman-maintain` that can be used to ease the process of shutting down
+services. Make sure that neither Puppet nor a monitoring system automatically
+starts these services again. On Debian compatible systems, stop the `apache2`
+service.
+
+You can issue the MySQL command `SHOW FULL PROCESSLIST;` to show you all active
+connections to your MySQL database. Make sure that there are none.
+
+The foreman user that has access to your PostgreSQL database and is used for
+the conversion temporarily needs super user privileges.
+
+```bash
+## postgres@postgres:~  $ su - postgres
+## postgres@postgres:~  $ psql
+psql (9.6.2)
+Type "help" for help.
+
+postgres=# ALTER USER foreman WITH SUPERUSER;
+ALTER ROLE
+postgres=#
+```
+
+#### Pre-migrate cleanup (optional)
+
+Before proceeding migration, it's a good idea to permanently delete all audit
+records older than few days, unless they are required for compliance reasons:
+
+    foreman-rake audits:expire days=10 RAILS_ENV="production"
+
+To permanently delete all reports older than 2 days:
+
+    foreman-rake reports:expire days=2 batch_size=50000 sleep_time=0.000001 RAILS_ENV="production"
+
+The command above deletes records in batches of 50k items, it can take a lot of
+time to process all records. In both cases, progress can be seen in the
+`production.log`. Alternatively a much faster process can be applied instead but
+it will delete all data. The following command removes all audit records:
+
+    echo "truncate table audits;" | mysql -u foreman -p -D foreman
+
+The following command removes all reports:
+
+    echo "truncate table sources;" | mysql -u foreman -p -D foreman
+    echo "truncate table messages;" | mysql -u foreman -p -D foreman
+    echo "truncate table resources;" | mysql -u foreman -p -D foreman
+    echo "truncate table logs;" | mysql -u foreman -p -D foreman
+
+The following command removes all facts:
+
+    echo "truncate table fact_values;" | mysql -u foreman -p -D foreman
+    echo "truncate table fact_names;" | mysql -u foreman -p -D foreman
+
+#### Migrating
+
+To migrate the data from one database to the other, we will use a rake task.
+We will be using `foreman-rake db:convert:prod2dev` to migrate the data.
+
+Foreman internally uses a development framework called Rails. In Rails, you can
+have a specific configuration for e.g. a development and a production
+environment. Our prod2dev rake task assumes, that the source database is
+configured as the production database and the new target database is configured
+as the development database. All data in the development database will be
+overwritten.
+
+You can configure the new database as Foreman's new development database by
+adding a new connection in `/etc/foreman/database.yml`.
+
+It may look like this:
+
+```yaml
+# The new target database
+development:
+  adapter: postgresql
+  host: postgres.example.com
+  port: 5432
+  database: foreman
+  username: foreman
+  password: "foreman"
+  pool: 25
+
+# The old source database
+production:
+  adapter: mysql2
+  host: mysql.example.com
+  port: 3306
+  database: foreman
+  username: foreman
+  password: "foreman"
+  pool: 25
+```
+
+In the next step, call a rake task which will create new database:
+
+```sh
+forman-rake db:create RAILS_ENV=development
+```
+
+The new database needs to be set up with all the tables required for Foreman to
+operate, so in order to prepare the new database to receive the actual data, we
+need to run all of Foreman's database migrations.
+
+```sh
+foreman-rake db:migrate RAILS_ENV=development
+```
+
+Next, we can finally migrate all the data. This will take a long time depending
+on the size of your database. It is highly recommended to run this in a screen
+or tmux session.
+
+```sh
+foreman-rake db:convert:prod2dev
+
+```
+
+The task performs several actions:
+
+* Migration of the data via prod2dev task
+* Reset of PostgreSQL sequences
+* Dynflow data migration
+
+#### Wrapping up
+
+Then, make sure to resore the original permissions of your foreman database
+user. This can be done on PostgreSQL by issuing this SQL command.
+
+```sql
+ALTER ROLE "foreman" NOSUPERUSER;
+```
+
+To switch your Foreman instance to the new database, you need to change the
+production database configuration in `/etc/foreman/database.yml` to point to
+the development installation we configured earlier:
+
+```yaml
+production:
+  adapter: postgresql
+  host: postgres.example.com
+  port: 5432
+  database: foreman
+  username: foreman
+  password: "foreman"
+  pool: 25
+```
+
+Then you can finally start Foreman's services again. On EL7 use
+`foreman-maintain service` helper or do:
+
+```sh
+systemctl start httpd.service
+systemctl start dynflowd.service
+```
+
+On Debian systems:
+
+```sh
+systemctl start apache2.service
+systemctl start dynflowd.service
+```
+
+Foreman installer can be used to apply the database configuration change and
+start the services. In that case, perform:
+
+    foreman-installer \
+      --foreman-db-type=postgresql \
+      --foreman-db-manage=true \
+      --foreman-db-host=postgres.example.com \
+      --foreman-db-database=foreman \
+      --foreman-db-username=foreman \
+      --foreman-db-password=foreman
--- a/_posts/2017-10-31-migrating-foreman-from-mysql-to-postgresql.md
+++ b/_posts/2017-10-31-migrating-foreman-from-mysql-to-postgresql.md
@@ -7,6 +7,8 @@ tags:
 - foreman
 ---
 
+*The [official Foreman manual](/manuals/1.23/index.html#3.7MigratingtoPostgreSQL) now contains the up to date and recommended migration procedure, please follow the instructions there.*
+
 While Foreman supports both MySQL and Postgresql as a database management system, Katello requires Postgresql. If you want to add the Katello plugin on an existing Foreman setup with MySQL, you need to migrate your database. But there are other reasons why you might want to switch your database management system. This blog post describes what steps are required to migrate your data.
 
 
--- a/_posts/2019-08-01-dropping-support-for-mysql.md
+++ b/_posts/2019-08-01-dropping-support-for-mysql.md
@@ -1,0 +1,90 @@
+---
+layout: post
+title: Dropping support for MySQL
+date: 2019-09-04 16:30:00
+author: Lukas Zapletal
+tags:
+- foreman
+---
+
+Foreman is essentially a Ruby on Rails application. A big one. And Ruby on
+Rails applications usually support multiple relational databases thanks to
+ActiveRecord stack. That's also the case for Foreman. But it's not that easy.
+
+<!--more-->
+
+Foreman currently requires one of the following databases:
+
+* PostgreSQL 9.2 or newer
+* MySQL 5.5 or newer
+
+This is not an extensive list you might think. Supporting two databases must
+not be that hard. Truth is, it's a challenge.
+
+Those two items on the list are both relational database systems with SQL
+standard, the problem is that very often behavior is different. Most of our
+development team use PostgreSQL and there are dozens of things that are
+different in MySQL. Therefore we need to identify those problems, find a way to
+fix them and then carry extra bit of code in our codebase for life. This takes
+a lot of our time we could invest in features, bugfixing or performance
+improvements.
+
+We have dicussed preferring PostgreSQL many times, but [the last time was
+different](https://community.theforeman.org/t/minimum-rdbm-requirements/14127)
+and we came to conclusion that dropping MySQL the right thing to do. [The
+plan](https://community.theforeman.org/t/rfc-dropping-support-for-mysql/14185)
+was then set and we are at the beginning of the journey where the first and
+really difficult step is on you, our users:
+
+**Foreman 1.24 will be the last version with MySQL support.**
+
+The upcoming versions 1.23 and 1.24 will also send a notification to all
+administrators in case it's running on MySQL. Don't worry, we will let you
+close the warning as soon as you see it, but we want to communicate this
+properly and in advance. We highly recommend to do the database migration to
+PostgreSQL right after upgrade to version 1.23 because if there is a blocker
+bug that prevents from migrating to MySQL, there is still a backup plan of
+holding off migration and staying on MySQL for one more release until we come
+up with a version which will be PostgreSQL only. This is important:
+
+**Do not wait with the migration for Foreman 1.24.**
+
+Along with the 1.23 release, we will provide official [migration
+documentation](/manuals/1.23/index.html#3.7MigratingtoPostgreSQL), and
+mention it in our Release Notes as well. If you do such testing, please come
+back to us with time estimation so we can tell our users in advance what to
+expect.
+
+It was not easy decision, we tried hard in the past to support both databases.
+If you search for [PRs containing
+MySQL](https://github.com/theforeman/foreman/pulls?page=3&q=is%3Apr+mysql)
+you'll find countless workarounds, fixes and endless discussions about how to
+solve problems in a way that doesn't break for PostgreSQL users. There is one
+common theme everywhere: most developers do not regularly use MySQL both for
+development and for production testing. This is compounded by the fact that the
+two commercial products based on Foreman, Red Hat Satellite and Atix
+Orcharhino, only support PostgreSQL, and many of the Foreman developers are
+employed by them. As none of their customers use MySQL, there is less testing
+and usage of it leading to longer times until bugs are discovered and fixed.
+Also, one of the biggest and most popular plugins, Katello, does not support
+MySQL, so many community users are not using it either.
+
+We also think that dropping MySQL support is big enough reason to bump the
+major release version to 2.0. Chances are that PostgreSQL only version will not
+be versioned 1.25 but 2.0. The final decision will be made before the release
+process starts, but we wanted to let you know in advance.
+
+We don't stop here, we already have ideas for more improvements we could do
+once PostgreSQL is the only supported RDBM. For example:
+
+* hash indexes to save disk space and memory
+* macaddr, macaddr6, ip, ip6 types
+* JSON/JSONB type
+* data partitioning
+* full text search
+* stored procedures or functions
+* other performance optimalizations
+
+There are many ways we can make Foreman better and stronger. Maybe you have
+your own ideas! We want feedback, visit us at https://community.theforeman.org
+and drop a comment.
--- a/manuals/nightly/index.md
+++ b/manuals/nightly/index.md
@@ -64,6 +64,8 @@ previous_version: "1.22"
 {%include manuals/{{ page.version }}/3.5.6_cli.md %}
 ## 3.6 Upgrade
 {%include manuals/{{ page.version }}/3.6_upgrade.md %}
+## 3.7 Migrating to PostgreSQL
+{%include manuals/{{ page.version }}/3.7_migration.md %}
 
 # 4. General Foreman
 {%include manuals/{{ page.version }}/4_general_foreman.md %}
--- a/_includes/manuals/1.23/3.7_migration.md
+++ b/_includes/manuals/1.23/3.7_migration.md
@@ -1,0 +1,204 @@
+
+### Migrating to PostgreSQL
+
+Starting from Foreman 1.25, MySQL is no longer a supported database
+management system. This chapter describes the process of migrating from MySQL
+to PostgreSQL.
+
+Full backup is absolutely necessary before starting the process. We recommend
+both database dump and snapshot of whole operating system and data if possible
+when running in a VM or on a LVM.
+
+#### Prerequisites
+
+It's a good idea to delete unwanted records from the database. The smaller the
+database is, the faster the migration process will be. Typical offenders are
+audit records, facts and reports. Foreman ships with rake tasks and cron job to
+execute these cleanup tasks, but they can be executed manually. Steps are shown
+down below after all instances are stopped.
+
+Foreman does not support plugin uninstallation, any leftovers in the database
+might cause issues during migration.
+
+#### Preparing the environment
+
+The first step is to install required dependencies:
+
+    yum install foreman-postgresql
+
+or
+
+    apt-get install foreman-postgresql
+
+Next, stop all your foreman instances and make sure, they stay stopped during
+the migration. On Red Hat compatible systems, you need to stop the `httpd` and
+(if applicable) `foreman-tasks` service. On EL7, there is a helper utility
+called `foreman-maintain` that can be used to ease the process of shutting down
+services. Make sure that neither Puppet nor a monitoring system automatically
+starts these services again. On Debian compatible systems, stop the `apache2`
+service.
+
+You can issue the MySQL command `SHOW FULL PROCESSLIST;` to show you all active
+connections to your MySQL database. Make sure that there are none.
+
+The foreman user that has access to your PostgreSQL database and is used for
+the conversion temporarily needs super user privileges.
+
+```bash
+## postgres@postgres:~  $ su - postgres
+## postgres@postgres:~  $ psql
+psql (9.6.2)
+Type "help" for help.
+
+postgres=# ALTER USER foreman WITH SUPERUSER;
+ALTER ROLE
+postgres=#
+```
+
+#### Pre-migrate cleanup (optional)
+
+Before proceeding migration, it's a good idea to permanently delete all audit
+records older than few days, unless they are required for compliance reasons:
+
+    foreman-rake audits:expire days=10 RAILS_ENV="production"
+
+To permanently delete all reports older than 2 days:
+
+    foreman-rake reports:expire days=2 batch_size=50000 sleep_time=0.000001 RAILS_ENV="production"
+
+The command above deletes records in batches of 50k items, it can take a lot of
+time to process all records. In both cases, progress can be seen in the
+`production.log`. Alternatively a much faster process can be applied instead but
+it will delete all data. The following command removes all audit records:
+
+    echo "truncate table audits;" | mysql -u foreman -p -D foreman
+
+The following command removes all reports:
+
+    echo "truncate table sources;" | mysql -u foreman -p -D foreman
+    echo "truncate table messages;" | mysql -u foreman -p -D foreman
+    echo "truncate table resources;" | mysql -u foreman -p -D foreman
+    echo "truncate table logs;" | mysql -u foreman -p -D foreman
+
+The following command removes all facts:
+
+    echo "truncate table fact_values;" | mysql -u foreman -p -D foreman
+    echo "truncate table fact_names;" | mysql -u foreman -p -D foreman
+
+#### Migrating
+
+To migrate the data from one database to the other, we will use a rake task.
+We will be using `foreman-rake db:convert:prod2dev` to migrate the data.
+
+Foreman internally uses a development framework called Rails. In Rails, you can
+have a specific configuration for e.g. a development and a production
+environment. Our prod2dev rake task assumes, that the source database is
+configured as the production database and the new target database is configured
+as the development database. All data in the development database will be
+overwritten.
+
+You can configure the new database as Foreman's new development database by
+adding a new connection in `/etc/foreman/database.yml`.
+
+It may look like this:
+
+```yaml
+# The new target database
+development:
+  adapter: postgresql
+  host: postgres.example.com
+  port: 5432
+  database: foreman
+  username: foreman
+  password: "foreman"
+  pool: 25
+
+# The old source database
+production:
+  adapter: mysql2
+  host: mysql.example.com
+  port: 3306
+  database: foreman
+  username: foreman
+  password: "foreman"
+  pool: 25
+```
+
+In the next step, call a rake task which will create new database:
+
+```sh
+forman-rake db:create RAILS_ENV=development
+```
+
+The new database needs to be set up with all the tables required for Foreman to
+operate, so in order to prepare the new database to receive the actual data, we
+need to run all of Foreman's database migrations.
+
+```sh
+foreman-rake db:migrate RAILS_ENV=development
+```
+
+Next, we can finally migrate all the data. This will take a long time depending
+on the size of your database. It is highly recommended to run this in a screen
+or tmux session.
+
+```sh
+foreman-rake db:convert:prod2dev
+
+```
+
+The task performs several actions:
+
+* Migration of the data via prod2dev task
+* Reset of PostgreSQL sequences
+* Dynflow data migration
+
+#### Wrapping up
+
+Then, make sure to resore the original permissions of your foreman database
+user. This can be done on PostgreSQL by issuing this SQL command.
+
+```sql
+ALTER ROLE "foreman" NOSUPERUSER;
+```
+
+To switch your Foreman instance to the new database, you need to change the
+production database configuration in `/etc/foreman/database.yml` to point to
+the development installation we configured earlier:
+
+```yaml
+production:
+  adapter: postgresql
+  host: postgres.example.com
+  port: 5432
+  database: foreman
+  username: foreman
+  password: "foreman"
+  pool: 25
+```
+
+Then you can finally start Foreman's services again. On EL7 use
+`foreman-maintain service` helper or do:
+
+```sh
+systemctl start httpd.service
+systemctl start dynflowd.service
+```
+
+On Debian systems:
+
+```sh
+systemctl start apache2.service
+systemctl start dynflowd.service
+```
+
+Foreman installer can be used to apply the database configuration change and
+start the services. In that case, perform:
+
+    foreman-installer \
+      --foreman-db-type=postgresql \
+      --foreman-db-manage=true \
+      --foreman-db-host=postgres.example.com \
+      --foreman-db-database=foreman \
+      --foreman-db-username=foreman \
+      --foreman-db-password=foreman
--- a/manuals/1.23/index.md
+++ b/manuals/1.23/index.md
@@ -64,6 +64,8 @@ previous_version: "1.22"
 {%include manuals/{{ page.version }}/3.5.6_cli.md %}
 ## 3.6 Upgrade
 {%include manuals/{{ page.version }}/3.6_upgrade.md %}
+## 3.7 Migrating to PostgreSQL
+{%include manuals/{{ page.version }}/3.7_migration.md %}
 
 # 4. General Foreman
 {%include manuals/{{ page.version }}/4_general_foreman.md %}


ares:
reads well, :+1:, is there some timing expectation or can we merge as is right away? is the banner for 1.23 ready?

_posts/2019-06-28-dropping-support-for-mysql.md
```diff
@@ -0,0 +1,47 @@
...
+
+We don't stop here, the next target is dropping SQLite3 forcing all developers to use PostgreSQL. It's the only reasonable way we can do in order to leverage all powerful features from the database system of our choice. However, again it's not easy. In the build process, we have a part that generates what's called APIPie bindings, a stack which provides API documentation and bindings metadata. And this needs to boot up Rails, and Rails needs a database to run on and we all do this in a build chroot which does not provide environment for starting up daemon like PostgreSQL. We need to figure out how to workaround this problem, until then we still support SQLite3 for development but developers be ready to say goodbye to SQLite3 any new version.
```
tbrisker: I don't think this paragraph is relevant for our users. This is a more developer oriented issue.

_posts/2019-06-28-dropping-support-for-mysql.md
```diff
@@ -0,0 +1,47 @@
...
+
+It was not easy decision, we tried hard in the past to support both databases. If you search for [PRs containing MySQL](https://github.com/theforeman/foreman/pulls?page=3&q=is%3Apr+mysql) you'll find counless workarounds, fixes and endless discussions about how to solve problems in a way that doesn't break for PostgreSQL users. There is one common theme everywhere: most developers do not use MySQL both for development and for production testing. This stems from the fact that Red Hat, a company which pays most of the users to do day-to-day development of Foreman, only have a product that is based on PostgreSQL. So we often try to reproduce things on PostgreSQL or blindly provide patches and let affected users to hotfix their instances. Also the biggest and most popular plugin called Katello does not support MySQL, so many community users are not using it at all. Although we haven't asked about it explicitly in the Foreman Community Survey yet, we believe that the majority of our users use PostgreSQL already.
```
tbrisker: Red Hat isn't the reason developers don't use MySQL, some RH developers actually do, and in any case this isn't the reason for moving away from PG. I would drop that sentence.

_posts/2019-06-28-dropping-support-for-mysql.md
```diff
@@ -0,0 +1,47 @@
...
+<!--more-->
+
+Foreman 1.22 currently requires one of the following databases:
+
+* PostgreSQL 9.2 or newer
+* MySQL 5.5 or newer
+
+This is not an extensive list you might think. Supporting two databases must not be that hard. Truth is, it's a challenge.
+
+Those two items on the list are both relational database systems with SQL standard, the problem is that very often behavior is different. Historically, some features were implemented differently, however it's pretty clear that PostgreSQL is now much closer to the RDBM "ideal" for both developers and operations, should I say. There are dozens of things that are just not right in MySQL and we need to identify those problems, find a way to fix them and then carry extra bit of code in our codebase for life. This takes a lot of our time we could invest in features, bugfixing or performance improvements.
```
tbrisker: Each DB has it's own merits. We aren't moving to PG because it's "better" but rather because supporting just one has its benefits, and we already have plugins and other services that require it. PG also has weird problems we need to identify and fix, but having less databases to support will reduce the amount of such workarounds and will stop forcing us to always write db-agnostic code.

tbrisker (changes requested):
Fixed some grammar issues, other comments inline. Also, where is the banner for 1.23 added?

_posts/2019-06-28-dropping-support-for-mysql.md
ekohl: https://about.gitlab.com/2019/06/27/removing-mysql-support/ is a similar blog about the exact same topic. One thing stood out to me:

> As a side note – we don't use MySQL ourselves, and not doing so meant we weren't encountering issues BEFORE our users.

Both commercial vendors (Red Hat and ATIX) both only deploy on PostgreSQL. That means they care about proper support. MySQL always has been best effort by the community. The Foreman project wants to deliver high quality software and that's more than just 'it compiles, ship it'.

That's at least my take on this.

_posts/2019-06-28-dropping-support-for-mysql.md
lzap: Sure removing.

_posts/2019-06-28-dropping-support-for-mysql.md
lzap: However is it a reason not to communicate this? My thinking it - I don't want to leave users just with simple "it will be gone", I want to be as much as transparent about our plans, goals and long term vision.

ekohl: Perhaps you can link to a Discourse RFC that talks about the future?

_posts/2019-06-28-dropping-support-for-mysql.md
lzap: I am rewriting the paragraph and dropping all my biased words about MySQL :-)

lzap:
Added new commit with the change, please make sure to squash on merge.

_posts/2019-06-28-dropping-support-for-mysql.md
lzap: Dropped it.

lzap:
Ping, I actually rebased once again and created new chapter 3.7 in the documentation which is based on Timos article but I added past experience with MySQL migrations we are seeing on Discourse. This is now ready.

manuals/nightly/index.md
```diff
@@ -64,6 +64,8 @@ previous_version: "1.22"
 {%include manuals/{{ page.version }}/3.5.6_cli.md %}
 ## 3.6 Upgrade
 {%include manuals/{{ page.version }}/3.6_upgrade.md %}
+## 3.7 Migrating to PostgreSQL
+{%include manuals/{{ page.version }}/3.7_migration.md %}
```
tbrisker, line 68: forgot `git add`?

_posts/2019-06-28-dropping-support-for-mysql.md
```diff
@@ -0,0 +1,45 @@
...
+
+Along with the 1.23 release, we will provide official [migration documentation](/manuals/nightly/index.html#3.7MigratingtoPostgreSQL), and mention it in our Release Notes as well. If you do such testing, please come back to us with time estimation so we can tell our users in advance what to expect.
+
+It was not easy decision, we tried hard in the past to support both databases. If you search for [PRs containing MySQL](https://github.com/theforeman/foreman/pulls?page=3&q=is%3Apr+mysql) you'll find counless workarounds, fixes and endless discussions about how to solve problems in a way that doesn't break for PostgreSQL users. There is one common theme everywhere: most developers do not use MySQL both for development and for production testing. This stems from the fact that Red Hat, a company which pays most of the users to do day-to-day development of Foreman, only have a product that is based on PostgreSQL. So we often try to reproduce things on PostgreSQL or blindly provide patches and let affected users to hotfix their instances. Also the biggest and most popular plugin called Katello does not support MySQL, so many community users are not using it at all.
```
tbrisker:
```suggestion
It was not easy decision, we tried hard in the past to support both databases. If you search for [PRs containing MySQL](https://github.com/theforeman/foreman/pulls?page=3&q=is%3Apr+mysql) you'll find countless workarounds, fixes and endless discussions about how to solve problems in a way that doesn't break for PostgreSQL users. There is one common theme everywhere: most developers do not regularly use MySQL both for development and for production testing. This is compounded by the fact that the two commercial products based on Foreman, Red Hat Satellite and Atix Orcharhino, only support PostgreSQL, and many of the Foreman developers are employed by them. As none of their customers use MySQL, there is less testing and usage of it leading to longer times until bugs are discovered and fixed. Also, one of the biggest and most popular plugins, Katello, does not support MySQL, so many community users are not using it either.
```

_posts/2019-06-28-dropping-support-for-mysql.md
```diff
@@ -0,0 +1,45 @@
...
+
+We have dicussed preferring PostgreSQL many times, but [the last time was different](https://community.theforeman.org/t/minimum-rdbm-requirements/14127) and we came to conclusion that dropping MySQL the right thing to do. [The plan](https://community.theforeman.org/t/rfc-dropping-support-for-mysql/14185) was then set and we are at the beginning of the journey where the first and really difficult step is on you, our users:
+
+**Foreman 1.23 will be the last version with MySQL support.**
+
+The upcoming version 1.23 will also feature a warning UI strip in case it's running on MySQL. Don't worry, we will let you close the warning as soon as you see it, but we want to communicate this properly and in advance. We highly recommend to do the database migration to PostgreSQL during upgrade to version 1.23 because if there is a blocker bug that prevents from migrating to MySQL, there is still a backup plan of holding off migration and staying on MySQL for one more release until we come up with version 1.24 which will be PostgreSQL only. This is important:
```
tbrisker: The UI warning is still missing

tbrisker (changes requested):

manuals/nightly/index.md
lzap: Oh yes, pretty much :-)

_posts/2019-06-28-dropping-support-for-mysql.md
lzap: Going to push PR shortly.

_posts/2019-06-28-dropping-support-for-mysql.md
lzap: Actually I dropped the whole paragraph, I just forgot to push properly.

lzap:
Squashed, rebased, added the missing file, renamed the filename to more appropriate date "1st of August".

lzap:
Here is the change that squashing hide for you to review, sorry about that:

```diff
--- a/_posts/2019-06-28-dropping-support-for-mysql.md
+++ b/_posts/2019-06-28-dropping-support-for-mysql.md
@@ -1,7 +1,7 @@
 ---
 layout: post
 title: Dropping support for MySQL
-date: 2019-06-28  9:34:21
+date: 2019-07-22 9:34:21
 author: Lukas Zapletal
 tags:
 - foreman
@@ -18,9 +18,9 @@ Foreman 1.22 currently requires one of the following databases:
 
 This is not an extensive list you might think. Supporting two databases must not be that hard. Truth is, it's a challenge.
 
-Those two items on the list are both relational database systems with SQL standard, the problem is that very often behavior is
 different. Historically, some features were implemented differently, however it's pretty clear that PostgreSQL is now much closer to the RDBM "ideal" for both developers and operations, should I say. There are dozens of things that are just not right in MySQL and we need to identify those problems, find a way to fix them and then carry extra bit of code in our codebase for life. This takes a lot of our time we could invest in features, bugfixing or performance improvements.
+Those two items on the list are both relational database systems with SQL standard, the problem is that very often behavior is different. Most of our development team use PostgreSQL and there are dozens of things that are different in MySQL. Therefore we need to identify those problems, find a way to fix them and then carry extra bit of code in our codebase for life. This takes a lot of our time we could invest in features, bugfixing or performance improvements.
 
-We have dicussed preferring PostgreSQL many times, but [the last time was different](https://community.theforeman.org/t/minimum-rdbm-requirements/14127) and we came to conclusion it's the right thing to do. [The plan](https://community.theforeman.org/t/rfc-dropping-support-for-mysql/14185) was then set and we are at the beginning of the journey where the first and really difficult step is on you, our users:
+We have dicussed preferring PostgreSQL many times, but [the last time was different](https://community.theforeman.org/t/minimum-rdbm-requirements/14127) and we came to conclusion that dropping MySQL the right thing to do. [The plan](https://community.theforeman.org/t/rfc-dropping-support-for-mysql/14185) was then set and we are at the beginning of the journey where the first and really difficult step is on you, our users:
 
 **Foreman 1.23 will be the last version with MySQL support.**
 
@@ -28,13 +28,11 @@ The upcoming version 1.23 will also feature a warning UI strip in case it's runn
 
 **Do not wait with the migration for Foreman 1.24.**
 
-Along with the 1.23 release, we will provide official migration documentation, and mention it in our Release Notes as well. If you want to give the migration a test shot today, there's an existing [migration blog post](https://theforeman.org/2017/10/migrating-foreman-from-mysql-to-postgresql.html) written by Timo. We don't expect the official document be very much different. If you do such testing, please come back to us with time estimation so we can tell our users in advance what to expect.
+Along with the 1.23 release, we will provide official [migration documentation](/manuals/nightly/index.html#3.7MigratingtoPostgreSQL), and mention it in our Release Notes as well. If you do such testing, please come back to us with time estimation so we can tell our users in advance what to expect.
 
-It was not easy decision, we tried hard in the past to support both databases. If you search for [PRs containing MySQL](https://github.com/theforeman/foreman/pulls?page=3&q=is%3Apr+mysql) you'll find counless workarounds, fixes and endless discussions about how to solve problems in a way that doesn't break for PostgreSQL users. There is one common theme everywhere: most developers do not use MySQL both for development and for production testing. This stems from the fact that Red Hat, a company which pays most of the users to do day-to-day development of Foreman, only have a product that is based on PostgreSQL. So we often try to reproduce things on PostgreSQL or blindly provide patches and let affected users to hotfix their instances. Also the biggest and most popular plugin called Katello does not support MySQL, so many community users are not using it at all. Although we haven't asked about it explicitly in the Foreman Community Survey yet, we believe that the majority of our users use PostgreSQL already.
+It was not easy decision, we tried hard in the past to support both databases. If you search for [PRs containing MySQL](https://github.com/theforeman/foreman/pulls?page=3&q=is%3Apr+mysql) you'll find counless workarounds, fixes and endless discussions about how to solve problems in a way that doesn't break for PostgreSQL users. There is one common theme everywhere: most developers do not use MySQL both for development and for production testing. This stems from the fact that Red Hat, a company which pays most of the users to do day-to-day development of Foreman, only have a product that is based on PostgreSQL. So we often try to reproduce things on PostgreSQL or blindly provide patches and let affected users to hotfix their instances. Also the biggest and most popular plugin called Katello does not support MySQL, so many community users are not using it at all.
 
-We don't stop here, the next target is dropping SQLite3 forcing all developers to use PostgreSQL. It's the only reasonable way we can do in order to leverage all powerful features from the database system of our choice. However, again it's not easy. In the build process, we have a part that generates what's called APIPie bindings, a stack which provides API documentation and bindings metadata. And this needs to boot up Rails, and Rails needs a database to run on and we all do this in a build chroot which does not provide environment for starting up daemon like PostgreSQL. We need to figure out how to workaround this problem, until then we still support SQLite3 for development but developers be ready to say goodbye to SQLite3 any new version.
-
-Let's evaluate after some time, because we already have ideas for many improvements we could do once PostgreSQL is the only supported RDBM. For example:
+We don't stop here, we already have ideas for more improvements we could do once PostgreSQL is the only supported RDBM. For example:
```

lzap:
Forgot to add your last minute change in the "It was not an easy decition" paragraph, now it should be ready for re-review. Sorry I was unable to revive changes from reflog, I screwed git hard today locally :-(

lzap:
Updated the text: 1.24 is the last version with MySQL. Added information about possible major version bump 2.0 (1.25) - it's not decided yet but we can communicate this early.

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,185 @@
+
+### Migrating to PostgreSQL
+
+Starting from Foreman 1.25/2.0, MySQL is no longer a supported database
+management system. This chapter describes the process of migrating from MySQL
+to PostgreSQL.
+
+Full backup is absolute necessary before starting the process. We recommend
```
tbrisker:
```suggestion
Full backup is absolutely necessary before starting the process. We recommend
```

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,185 @@
...
+both database dump and snapshot of whole operating system and data if possible
+when running in a VM or on a LVM.
+
+#### Prerequisites
+
+It's a good idea to delete unwanted records from the database. The smaller the
+database is, the faster will be the migration process. Typical offenders are
```
tbrisker:
```suggestion
database is, the faster the migration process will be. Typical offenders are
```

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,185 @@
...
+good. On Red Hat compatible systems, you need to stop the `httpd` and (if
+applicable) `foreman-tasks` service. There is a helper utility called
+`foreman-maintain` when installed to ease the process of shutting down
+services. Make sure that neither Puppet nor a monitoring system automatically
+starts these services again.
+
+You can issue the MySQL command `SHOW FULL PROCESSLIST;` to show you all active
+connections to your MySQL database. Make sure, there are none.
+
+TODO: (WHY DO WE NEED THIS): If you are using the foreman-tasks plugin, you
```
tbrisker: Is this still needed? we should get this into the migration script if needed instead of manually patching from @timogoebel 's gist. cc: @adamruzicka

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,185 @@
...
+#### Prerequisites
+
+It's a good idea to delete unwanted records from the database. The smaller the
+database is, the faster will be the migration process. Typical offenders are
+audit records, facts and reports. Foreman ships with rake tasks and cron job to
+execute these cleanup tasks, but they can be executed manually. Steps are shown
+down below after all instances are stopped.
+
+Foreman database must be cleaned up from any leftovers from uninstalled
+plugins. When a Foreman plugin is activated, database migration is perfored and
```
tbrisker:
```suggestion
plugins. When a Foreman plugin is activated, database migration is performed and
```

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,185 @@
...
+new tables and indices are created. However when a plugin is uninstalled and
+deactivated, no tables are ever dropped and the data is left in the database
+forever. It's good idea to review at least OS installed package log to find out
+if any foreman package was ever uninstalled. Unfortunately, there is no list of
+tables for each plugin.
```
tbrisker: I'm not sure how helpful this is - say they manage to find out what plugins were uninstalled, what then? we don't have instructions for cleaning up any leftovers and would need to help any users hitting issues because of that. It isn't just tables and indices, it could also be foreign keys, sti types and more. Perhaps it would be better to only warn that if any plugins have been uninstalled there may be issues and point to discourse for help or suggest reinstalling the plugins at least just for the migration step?

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,185 @@
...
+
+#### Preparing the environment
+
+This blog post assumes that you already have the source and target databases
```
tbrisker: this is no longer a blog post

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,185 @@
...
+set up. Make sure to setup the Postgresql database with UTF-8 enabled.
+
+The first step is to install required dependencies:
+
+    yum install -y foreman-postgresql
```
tbrisker: what about deb?

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,185 @@
...
+
+Next, stop all your foreman instances and make sure, they stay stopped for
+good. On Red Hat compatible systems, you need to stop the `httpd` and (if
```
tbrisker: stay stopped during the migration

tbrisker: on debian?

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,185 @@
...
+applicable) `foreman-tasks` service. There is a helper utility called
+`foreman-maintain` when installed to ease the process of shutting down
```
tbrisker: only on EL right now.

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,185 @@
...
+services. Make sure that neither Puppet nor a monitoring system automatically
+starts these services again.
+
+You can issue the MySQL command `SHOW FULL PROCESSLIST;` to show you all active
+connections to your MySQL database. Make sure, there are none.
```
tbrisker:
```suggestion
connections to your MySQL database. Make sure that there are none.
```

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,185 @@
...
+
+postgres=# ALTER USER foreman WITH SUPERUSER;
+ALTER ROLE
+postgres=#
+```
+
+#### Migrating
+
+Before proceeding migration, it's a good idea to permanently delete all audit
+records older than few days:
```
tbrisker:
```suggestion
records older than few days, unless they are required for compliance reasons:
```

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,185 @@
...
+Type "help" for help.
+
+postgres=# ALTER USER foreman WITH SUPERUSER;
+ALTER ROLE
+postgres=#
+```
+
+#### Migrating
+
+Before proceeding migration, it's a good idea to permanently delete all audit
```
tbrisker, line 60: Add a heading - pre-migration cleanup? Also, perhaps it is simpler and faster to truncate all the reports and audits tables?

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,185 @@
...
+To permanently delete all reports older than 2 days:
+
+    foreman-rake reports:expire days=2 batch_size=50000 sleep_time=0.000001 RAILS_ENV="production"
+
+The command above deletes records in batches of 50k items, it can take a lot of
+time to process all records. In both cases, progress can be seen in the
+`production.log`.
+
+To migrate the data from one database to the other, we will use a rake task.
+You might already be familiar to this task from upgrading Foreman to a newer
```
tbrisker: This isn't a task normally used in upgrades, i think we can drop this sentence.

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,185 @@
...
+version.
+
+Command `foreman-rake -T` lists all available tasks. We will be using
```
tbrisker: I think we can drop the -T sentence and combine the second one with the previous paragraph

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,185 @@
...
+  password: "foreman"
+  pool: 25
+```
+
+The new database needs to be set up with all the tables required for Foreman to
+operate, so in order to prepare the new database to receive the actual data, we
+need to run all of Foreman's database migrations.
+
+```sh
+forman-rake db:migrate RAILS_ENV=development
```
tbrisker: Don't we also need to run `db:create` first?

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,185 @@
...
+```sh
+forman-rake db:convert:prod2dev
+
+```
+
+#### Wrapping up
+
+Postgresql uses sequences in order to store the next free primary key id for a
+table. These sequences do not get updated by the rake tasks and have to be
+adjusted manually. On MySQL, this is not needed as the concept of sequences is
```
tbrisker: Should we add this script to the migration script in case the target is pg?

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,185 @@
...
+not known there.
+
+The Postgresql Wiki has a [script to repair the
+sequences](https://wiki.postgresql.org/wiki/Fixing_Sequences). Copy the script
+and save it to a file, e.g. `sequences.sql`. Then run it like this:
+
+```sh
+su - postgres
+vim sequences.sql
+# Paste the script and save the file
```
tbrisker: Didn't we already save the file before?

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,185 @@
...
+To switch your Foreman instance to the new database, you need to change the
+production database configuration in `/etc/foreman/database.yml` to point to
+the development installation we configured earlier. You can either do this
+manually or re-run foreman-installer with the new parameters for Postgresql.
+
+Then you can finally start Foreman's services again.
+
+```sh
+systemctl start httpd.service
+systemctl start foreman-tasks.service
```
tbrisker: doesn't httpd also start tasks now?

_posts/2017-10-31-migrating-foreman-from-mysql-to-postgresql.md
```diff
@@ -7,6 +7,8 @@ tags:
 - foreman
 ---
 
+This content was moved to the [official Foreman manual](/manuals/nightly/index.html#3.7MigratingtoPostgreSQL).
```
tbrisker:
```suggestion
*The [official Foreman manual](/manuals/1.23/index.html#3.7MigratingtoPostgreSQL) now contains the up to date and recommended migration procedure, please follow the instructions there.*
```

_posts/2019-08-01-dropping-support-for-mysql.md
```diff
@@ -0,0 +1,90 @@
...
+**Foreman 1.24 will be the last version with MySQL support.**
+
+The upcoming versions 1.23 and 1.24 will also send a notification to all
+administrators in case it's running on MySQL. Don't worry, we will let you
+close the warning as soon as you see it, but we want to communicate this
+properly and in advance. We highly recommend to do the database migration to
+PostgreSQL during upgrade to version 1.23 because if there is a blocker bug
+that prevents from migrating to MySQL, there is still a backup plan of holding
+off migration and staying on MySQL for one more release until we come up with
+versions 1.24 or 1.25 which will be PostgreSQL only. This is important:
```
tbrisker:
```suggestion
a version which will be PostgreSQL only. This is important:
```

_posts/2019-08-01-dropping-support-for-mysql.md
```diff
@@ -0,0 +1,90 @@
...
+
+**Do not wait with the migration for Foreman 1.24.**
+
+Along with the 1.23 release, we will provide official [migration
+documentation](/manuals/nightly/index.html#3.7MigratingtoPostgreSQL), and
```
tbrisker:
```suggestion
documentation](/manuals/1.23/index.html#3.7MigratingtoPostgreSQL), and
```

tbrisker (changes requested):
See inline comments. @timogoebel - since the migration procedure is based on your blog post, could you please also give it a look?

_includes/manuals/nightly/3.7_migration.md
timogoebel: dynflow uses tables that aren't managed by rails. That's why the migration script can't find a corresponding rails model and bails out.

_includes/manuals/nightly/3.7_migration.md
timogoebel: I think this would be great to have. Or another rake task that runs the commands on the database if it's mysql. Updating the sequences should be rather safe to do.

_includes/manuals/nightly/3.7_migration.md
adamruzicka: What @timogoebel said, Dynflow uses Sequel to access the db and does its own migrations.

_includes/manuals/nightly/3.7_migration.md
adamruzicka: I'm not aware of it. Why would it do that?

_includes/manuals/nightly/3.7_migration.md
tbrisker: Than why not just add the patch to the convert script? looking at the patch on the gist, it should be a noop if tasks isn't installed?

_includes/manuals/nightly/3.7_migration.md
ekohl: No, it doesn't the installer ensures it's running. The current name of `foreman-tasks` is `dynflowd` btw. `foreman-tasks` is a legacy alias.

_includes/manuals/nightly/3.7_migration.md
timogoebel: Yeah, we can add it to core and ship it with a release.

_includes/manuals/nightly/3.7_migration.md
tbrisker: Ok, ignore me, i was thinking of the dev environment which also starts dynflow.

tbrisker: would still be good to change the service name though

_includes/manuals/nightly/3.7_migration.md
tbrisker: let's do it than and get it into 1.23 please, along with the post-migration sequence step.

tbrisker:
follow up thought - iiuc we explicitly skip the foreman tasks table, will the migration work for users with tasks? or will they lose all tasks information? cc @adamruzicka

adamruzicka:
From the top of my head, if we skip all `foreman_tasks_*` and `dynflow_*` tables they will lose:
- historic task data (can be exported before the migration)
- in-progress tasks (needs dynflow tables)
- recurring logic definitions
- tasks scheduled into the future (needs dynflow tables)
  - katello sync plans will have to be re-enabled to work again
  - recurring REX jobs will need to be wiped and set up again

tbrisker:
How should we migrate the tasks data then? I think historic data might be fine and there shouldn't be any running tasks ideally during the upgrade, but what about the others?
>     * katello sync plans will have to be re-enabled to work again
This is the only one we don't need to worry about at all since katello already requires postgres

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,185 @@
...
+new tables and indices are created. However when a plugin is uninstalled and
+deactivated, no tables are ever dropped and the data is left in the database
+forever. It's good idea to review at least OS installed package log to find out
+if any foreman package was ever uninstalled. Unfortunately, there is no list of
+tables for each plugin.
+
+#### Preparing the environment
+
+This blog post assumes that you already have the source and target databases
+set up. Make sure to setup the Postgresql database with UTF-8 enabled.
```
ohadlevy: how does one do that? using the installer? imho you need to allow the user to copy-paste, or even write a small script to automate the migration.

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,185 @@
...
+```sh
+wget -O /tmp/convert-tasks-patch.diff https://gist.githubusercontent.com/timogoebel/cd29e5206678ca0e0d78222af70583ec/raw/bbe998624098c1c3e656fe2582d33b0d9f9285aa/0001-convert-skip-foreman-tasks-tables.patch
+yum install patch -y
+cd /usr/share/foreman
+
+patch -p1 < /tmp/convert-tasks-patch.diff
+```
+
+The foreman user that has access to your Postgresql database and is used for
+the conversion temporarily needs super user privileges.
```
ohadlevy, line 45: how would one create the account in the first place? traditionally this has been done by the installer.

_includes/manuals/nightly/3.7_migration.md
ohadlevy: tbh I don't think we would like to ask users to delete their audits, only if they dont use it, I would just say it can speed things up, but recommend users to delete it.

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,185 @@
...
+
+Foreman internally uses a development framework called Rails. In Rails, you can
+have a specific configuration for e.g. a development and a production
+environment. Our prod2dev rake task assumes, that the source database is
+configured as the production database and the new target database is configured
+as the development database. All data in the development database will be
+overwritten.
+
+You can configure the new database as Foreman's new development database by
+adding a new connection in `/etc/foreman/database.yml`.
```
ohadlevy, line 101: tbh I would patch the script to read the definitions from elsewhere, we can provide a simple yaml file with source/destination instead of messing up with rails settings.

_includes/manuals/nightly/3.7_migration.md
ohadlevy: exactly, really depending on how you configured pg, its users etc

ohadlevy (changes requested):
thanks for doing this, TBH IMHO we should just write a script that handles the migration (or as part of foreman maintain), it would be a lot easier to ask users to try it out vs giving them a doc.

adamruzicka:
> How should we migrate the tasks data then?

This should do the trick https://github.com/theforeman/foreman/pull/6962

_includes/manuals/nightly/3.7_migration.md
lzap: I think I saw a patch somewhere but can't find it. Can you guys point me on it if such a thing exists?

_includes/manuals/nightly/3.7_migration.md
lzap: Ok will do.

_posts/2019-08-01-dropping-support-for-mysql.md
lzap: Why not to point to the most recent version when a random user finds the post after some time? Updated instructions might be quite helpful I am pretty sure we will run into issues.

lzap:
> TBH IMHO we should just write a script that handles the migration

Maybe it's just me but if I am migrating a production instance, I want to execute every one last of commands manually to see what I am doing (and taking notes). If others think this is a good idea I can write such thing however.

ohadlevy:
> > TBH IMHO we should just write a script that handles the migration
> 
> Maybe it's just me but if I am migrating a production instance, I want to execute every one last of commands manually to see what I am doing (and taking notes). If others think this is a good idea I can write such thing however.

my thought process that we need to make it easy and reliable, those people who care (like you) would want to see the procedure, how it works, and handle it step by step, others, would just want to get it running as easily as possible.

I think there is a lot of things that can go wrong in this process, having a script that is tested and improved over time, would make it more reliable than docs... my worry is that if its a hard procedure, people will simply not migrate...

btw you can still do a script that goes step by step and ask for confirmation of each step (or skip)...

lzap:
> btw you can still do a script that goes step by step and ask for confirmation of each step (or skip)...

Can I do that in foreman maintain @mbacovsky ?

_includes/manuals/nightly/3.7_migration.md
lzap: Found it: https://github.com/theforeman/foreman/pull/6962

_includes/manuals/nightly/3.7_migration.md
lzap: Not following.

_includes/manuals/nightly/3.7_migration.md
lzap: We do this via cron anyway both for reports and (now) for audits. I will update the heading.

lzap: Marked the title as "optional" tho.

_includes/manuals/nightly/3.7_migration.md
lzap: Into which migration script, `prod2dev` is not supposed to do MySQL migrations.

_includes/manuals/nightly/3.7_migration.md
lzap: Replaced with https://github.com/theforeman/foreman/pull/6974

_includes/manuals/nightly/3.7_migration.md
lzap: Changed to dynflowd then.

_includes/manuals/nightly/3.7_migration.md
lzap: Yeah I also noticed that later on, I have removed this completely, everything is UTF-8 by default these days. Maybe @timogoebel remember why he added that?

_includes/manuals/nightly/3.7_migration.md
lzap: You don't need super privileges, what you really need is grant to create tables which we handle in the installer. For the migration using super user is probably fine.

_includes/manuals/nightly/3.7_migration.md
lzap: Well, I am not keen doing that, I can test that once or twice on my environment and I don't want this to be rather for the worse.

lzap:
Forget about that @mbacovsky we can't use foreman-maintain as it is not on Debian.

lzap:
Added bunch of commits, squash during merge. There are still few open questions tho.

_posts/2019-08-01-dropping-support-for-mysql.md
tbrisker: because we will not have that section in the manual once we drop mysql

_includes/manuals/nightly/3.7_migration.md
tbrisker: you gave the service name for rh systems, on debian it's apache2 iirc

_posts/2019-08-01-dropping-support-for-mysql.md
lzap: Why would we ever remove section named "Migration from MySQL to PostgreSQL" in the future? People are far behind with upgrading, we should keep it.

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,171 @@
+
+### Migrating to PostgreSQL
+
+Starting from Foreman 1.25/2.0, MySQL is no longer a supported database
```
tbrisker:
```suggestion
Starting from Foreman 1.25, MySQL is no longer a supported database
```
(at least until we agree on it)

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,171 @@
...
+    yum install foreman-postgresql
+
+or
+
+    apt-get install foreman-postgresql
+
+Next, stop all your foreman instances and make sure, they stay stopped during
+the migration. On Red Hat compatible systems, you need to stop the `httpd` and
+(if applicable) `foreman-tasks` service. On EL7, there is a helper utility
+called `foreman-maintain` when installed to ease the process of shutting down
```
tbrisker:
```suggestion
called `foreman-maintain` that can be used to ease the process of shutting down
```

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,171 @@
...
+psql (9.6.2)
+Type "help" for help.
+
+postgres=# ALTER USER foreman WITH SUPERUSER;
+ALTER ROLE
+postgres=#
+```
+
+#### Pre-migrate cleanup (optional)
+
```
tbrisker, line 59: Perhaps first suggest truncating these table if they are not needed? that would be much faster then deleting possibly millions of records. Also, are there other tables we might suggest cleaning up, such as facts?

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,171 @@
...
+    yum install foreman-postgresql
+
+or
+
+    apt-get install foreman-postgresql
+
+Next, stop all your foreman instances and make sure, they stay stopped during
+the migration. On Red Hat compatible systems, you need to stop the `httpd` and
+(if applicable) `foreman-tasks` service. On EL7, there is a helper utility
```
tbrisker, line 35: Still missing the service name for debian (apache2 iirc)

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,171 @@
...
+
+```sh
+foreman-rake db:migrate RAILS_ENV=development
+```
+
+Before proceeding with the actual data migration, it's required to do separate
+migration of dynflow tables.
+
+```sh
+foreman-rake dynflow:migrate
```
tbrisker: Isn't this step done as part of the migration script now? Thinking of which, should we also run the regular migration task as part of it? (my main though is this is a very long process, the fewer manual steps the user has to do the better)

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,171 @@
...
+forman-rake db:convert:prod2dev
+
+```
+
+Postgresql uses sequences in order to store the next free primary key id for a
+table. These sequences do not get updated by the rake tasks and have to be
+adjusted manually.
+
+```sh
+forman-rake sequence:reset RAILS_ENV=development
```
tbrisker: Let's call this as part of the convert script as well and not bother the user with another manual step

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,171 @@
...
+user. This can be done on Postgresql by issuing this SQL command.
+
+```sql
+ALTER ROLE "foreman" NOSUPERUSER;
+```
+
+To switch your Foreman instance to the new database, you need to change the
+production database configuration in `/etc/foreman/database.yml` to point to
+the development installation we configured earlier. You can either do this
+manually or re-run foreman-installer with the new parameters for Postgresql.
```
tbrisker: It's not clear what needs to be changed here if done manually or what are the new installer parameters for postgresql (and what parameters need to be reset to remove the mysql settings)

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,171 @@
...
+
+Then you can finally start Foreman's services again.
+
+```sh
+systemctl start httpd.service
```
tbrisker, line 184: or apache2

_posts/2019-08-01-dropping-support-for-mysql.md
tbrisker: because we don't want to carry around stuff in the manual that isn't used anymore, if people in the future do the upgrade they will still find the instructions in the 1.23/1.24 manual

tbrisker (changes requested):

_includes/manuals/nightly/3.7_migration.md
lzap: I'd rather advertise the clean way of doing things. I am going to do add a note that truncate is an option.

_includes/manuals/nightly/3.7_migration.md
lzap: Missed this one.

_includes/manuals/nightly/3.7_migration.md
lzap: I don't think so. More than that, it's not yet merged: https://github.com/theforeman/foreman/pull/6962

_includes/manuals/nightly/3.7_migration.md
lzap: I disagree, there is no such thing as the convert script. We do have `db:convert:prod2dev` which is a generic production to development convert script - your target database don't need to be PostgreSQL at all. I want to stick with least surprise, thus an explicit step is better. It's also better when calling for help - it's more obvious at which point something failed or what to do.

_includes/manuals/nightly/3.7_migration.md
lzap: Added bunch of instructions.

lzap:
Pushed bunch of changes as new commits.

_includes/manuals/nightly/3.7_migration.md
tbrisker: That will be merged once @adamruzicka gets back to the comments. 
My thought is that these tasks should be idempotent, so we can call them as part of the prod2dev flow to make sure the target db is in the right state. if we do that, we can make these instruction simpler for the users - so less chance of missing a step causing the next step to fail.

_includes/manuals/nightly/3.7_migration.md
tbrisker: Whenever prod2dev is called and the target db is postgres we would need to call this task, which is a noop anyways when the db isn't posgres. 
Users shouldn't need to care about database internals - they want to migrate from one db to the other. If this fails we'll be able to tell from the error what fails, and if they care about what the internals of the task does they can read it.

tbrisker (changes requested):
Almost there - I think the remaining issue is we disagree regarding having the convert script call other tasks to simplify the process. I would like users to have a simple path to migrate, the more complicated it is the more chance some of them will miss steps or just stick to 1.24 forever. 
The only other minor comment is that there is some inconsistent capitalization of PosgreSQL - some places use `Postgresql` instead.

lzap:
Pushing the PostgreSQL capitalization. I'd be fine with simplified workflow as long as we keep the current `prod2dev` task unchanged (because it's still quite useful task for development purposes) and introduce a different task "wrapper" that will perform everything what's needed.

tbrisker:
That works for me. lets make a `mysql_to_pg` task that wraps around these and does as much as possible on its own.

lzap:
Let me do this in https://github.com/theforeman/foreman/pull/6974 then. It's related.

lzap:
Integrated the new wrapper task.

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,204 @@
...
+```sh
+forman-rake db:create RAILS_ENV=development
+```
+
+The new database needs to be set up with all the tables required for Foreman to
+operate, so in order to prepare the new database to receive the actual data, we
+need to run all of Foreman's database migrations.
+
+```sh
+foreman-rake db:migrate RAILS_ENV=development
```
tbrisker, line 138: can we run this also inside the `mysql2pgsql` task?

lzap: This can be dangerous - the db migration must be performed in the development Rails env. Now, the mysql2pgsql wrapper task today contains only conversion task which is Rails environment independent. However I am going to add another task which resets postgresql sequence and that already requires RAILS_ENV to be set correctly (production). If we merge the `db:migrate` here, there is a conflict in Rails environments.

It could be solved indeed but that can be confusing when adding more and more tasks into the wrapper one.

lzap:
Any other blockers @tbrisker other than the PRs to be merged?

tbrisker (approved):
This looks good now, but I prefer to finish merging the related PRs first to make sure we don't have a mismatch between the instructions here and what actually needs to be done.

_includes/manuals/nightly/3.7_migration.md
```diff
@@ -0,0 +1,204 @@
...
+```
+
+Next, we can finally migrate all the data. This will take a long time depending
+on the size of your database. It is highly recommended to run this in a screen
+or tmux session.
+
+```sh
+foreman-rake db:convert:mysql2pgsql
```
tbrisker:
```suggestion
foreman-rake db:convert:prod2dev
```

_posts/2019-08-01-dropping-support-for-mysql.md
```diff
@@ -0,0 +1,90 @@
...
+was then set and we are at the beginning of the journey where the first and
+really difficult step is on you, our users:
+
+**Foreman 1.24 will be the last version with MySQL support.**
+
+The upcoming versions 1.23 and 1.24 will also send a notification to all
+administrators in case it's running on MySQL. Don't worry, we will let you
+close the warning as soon as you see it, but we want to communicate this
+properly and in advance. We highly recommend to do the database migration to
+PostgreSQL during upgrade to version 1.23 because if there is a blocker bug
```
tbrisker: Since we want them to use the updated script:
```suggestion
PostgreSQL right after upgrade to version 1.23 because if there is a blocker bug
```

tbrisker (changes requested):
Core is in, two small changes to this and also please copy the changes from nightly to the 1.23 manual as well.

_posts/2019-08-01-dropping-support-for-mysql.md
```diff
@@ -0,0 +1,90 @@
+---
+layout: post
+title: Dropping support for MySQL
+date: 2019-08-01 9:34:21
```
tbrisker: Also this should be updated

tbrisker (changes requested):

lzap:
All done, ping me if something's wrong! Cheers.

tbrisker (approved):
Thanks @lzap !

PR #1409, opened 2019-07-02
Clean up unattended_url upgrade warning

ekohl:
IMHO this is clearer for the user. cc @ezr-ondrej


ezr-ondrej (approved):
Yeah! it sounds much better, thanks @ekohl  :+1:

PR #1412, opened 2019-07-09
Use bundle exec in ktest

chris1984:


chris1984:
@tbrisker can I get a review and merge please :)

ekohl (approved):
This has nothing to do with SCLs. You need to do the same on Fedora since we no longer automatically activate bundler (which Foreman used to do).

PR #1413, opened 2019-07-10
remove global Jed use from documentation

amirfefer:
_related to https://github.com/theforeman/foreman/pull/6888_

global Jed was deprecated due to moving it to webpack a year ago


tbrisker (approved):
Thanks @amirfefer !

PR #1416, opened 2019-07-18
Update Katello to 3.12.1

chris1984:
@jturel here is the updated docs pr


jturel:
@chris1984 as far as I know we don't ever update the menus for .z versions :z

PR #1417, opened 2019-07-22
Added fdi.script option doc

lzap:
This was hidden feature but it is good idea to document it now I guess. It helped to solve few issues already.


tbrisker (approved):
Thanks @lzap !

PR #1419, opened 2019-07-23
Add warning on ansible-runner repository dependency

ezr-ondrej:
Warning about side effect of theforeman/puppet-foreman_proxy#515.


tbrisker:
is this going to be in 1.23 or 1.24? we just branched the manual for 1.23, so if that is where this lands this should change to the 1.23 manual

ares:
it's 1.23

tbrisker:
in that case please move it to the 1.23 upgrade notes @ezr-ondrej

ezr-ondrej:
Moved  to the 1.23 notes.

tbrisker (approved):
Thanks @ezr-ondrej !

PR #1427, opened 2019-08-01
Remove old events calendar

tbrisker:
Events are now managed on discourse instead of on the website, but the
old calendar was never removed, along with it's dependencies.


evgeni (approved):

ekohl:
I wonder if we should import them in discourse for completeness. It may be interesting to see which events we've been to in our history.

PR #1432, opened 2019-08-05
Allow using podman instead of docker

ekohl:


zjhuntin (approved):
Tested locally and works great!

evgeni (approved):

PR #1435, opened 2019-08-07
Update Puppet documentation

ekohl:
This is incomplete since at least the Puppet CA part needs to document the HTTP API-based implementation that's used since Puppet 6.

### DIFF
--- a/_includes/manuals/nightly/3.1.3_puppet_versions.md
+++ b/_includes/manuals/nightly/3.1.3_puppet_versions.md
@@ -1,84 +1,45 @@
-Foreman integrates with Puppet and Facter in a few places, but generally using a recent, stable version will be fine.  The exact versions of Puppet, Puppet Server and Facter that Foreman is compatible with are listed below.
+Foreman integrates with Puppet and Facter in a few places, but generally using a recent, stable version will be fine.  The exact versions of Puppet, Puppetserver and Facter that Foreman is compatible with are listed below.
 
 <table class="table table-bordered table-condensed">
   <tr>
     <th>Puppet version</th>
-    <th>Foreman installer</th>
-    <th>Smart proxy</th>
-    <th><a href="/manuals/{{page.version}}/index.html#3.5.4PuppetReports">Report/fact processors</a></th>
-    <th><a href="/manuals/{{page.version}}/index.html#3.5.5FactsandtheENC">External node classifier</a></th>
+    <th>Foreman installer (AIO)</th>
+    <th>Foreman installer (non-AIO)</th>
+    <th>Smart Proxy</th>
   </tr>
   <tr>
-    <td>0.x-3.x</td>
+    <td>0.x-4.3</td>
     <td>Not supported</td>
-    <td>Untested</td>
-    <td>Untested</td>
-    <td>Untested <span class='footnote'>*</span></td>
+    <td>Not supported</td>
+    <td>Not supported</td>
   </tr>
   <tr>
-    <td>4.x (AIO)</td>
+    <td>4.4-4.10</td>
+    <td>Not supported</td>
     <td>Not supported</td>
     <td>Deprecated</td>
-    <td>Deprecated</td>
-    <td>Deprecated</td>
-  </tr>
-  <tr>
-    <td>4.x (non-AIO)</td>
-    <td>Untested</td>
-    <td>Not Supported</td>
-    <td>Untested</td>
-    <td>Untested</td>
-  </tr>
-  <tr>
-    <td>5.x (AIO)</td>
-    <td>Supported</td>
-    <td>Supported</td>
-    <td>Supported</td>
-    <td>Supported</td>
-  </tr>
-  <tr>
-    <td>5.x (non-AIO)</td>
-    <td>Untested</td>
-    <td>Not Supported</td>
-    <td>Untested</td>
-    <td>Untested</td>
-  </tr>
-</table>
-
-<span class='footnote'>*</span> Puppet versions < 2.6.5 require `Parametrized_Classes_in_ENC` in Foreman to be disabled.
-
-#### Puppet Server compatibility
-
-Puppet Server is the application for serving Puppet agents used by default in Puppet 4/5 AIO installations.
-
-<table class="table table-bordered table-condensed">
-  <tr>
-    <th>Puppet Server version</th>
-    <th>Foreman installer</th>
-    <th>Notes</th>
-  </tr>
-  <tr>
-    <td>1.x-2.7</td>
-    <td>Untested</td>
-    <td>Version 2.0-2.1 requires server_use_legacy_auth_conf, server_puppetserver_version to be set on the Puppet Server.</td>
-  </tr>
-  <tr>
-    <td>2.8</td>
-    <td>Supported</td>
-    <td></td>
   </tr>
   <tr>
     <td>5.x</td>
     <td>Supported</td>
-    <td></td>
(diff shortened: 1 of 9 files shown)


_includes/manuals/nightly/3.3.1_rpm_packages.md
```diff
@@ -31,19 +31,16 @@ To enable the optional repository on a RHEL 7 system using subscription-manager,
...
 #### Available repositories
 
-Four main repos are provided at <http://yum.theforeman.org>:
+Four main repos are provided at <https://yum.theforeman.org>:
 
-* `/releases/latest` or `/releases/VERSION` (e.g. `/releases/{{page.version}}`) carries the stable releases and updates of Foreman and its dependencies.
-* `/rc` carries release candidates only in the few weeks prior to a release.  Ensure after a release is made that you use the main releases repo instead.
-* `/nightly` carries the latest development builds and as such, may be unstable or occasionally broken.
-* `/plugins` or `/plugins/VERSION` (e.g. `/plugins/{{page.version}}`) carries the stable plugin releases.
+* `/client` or `/client/VERSION` (e.g. `/client/{{page.version}}`) carries packages relevant to clients. This is optional and Foreman does not require anything installed on clients.
```
ekohl, line 40: Not really Puppet related, but needed an update as well.

_includes/manuals/nightly/3.3.1_rpm_packages.md
```diff
@@ -85,7 +82,7 @@ Install foreman and other foreman-* packages to add functionality:
 
 1. Configure by editing `/etc/foreman/settings.yaml` and `/etc/foreman/database.yml`
 1. After changing the database, migrate it: `sudo -u foreman /usr/share/foreman/extras/dbmigrate`
-1. Start the foreman service or set up passenger: `service foreman start`
+1. Start the foreman service or set up passenger: `systemctl start foreman`
```
ekohl, line 85: This needs the foreman-service package to be installed. Will need a larger overhaul.

_includes/manuals/nightly/3.1.3_puppet_versions.md
```diff
@@ -1,84 +1,45 @@
...
+    <td>Not supported</td>
+    <td>Not supported</td>
     <td>Not supported</td>
-    <td>Untested</td>
-    <td>Untested</td>
-    <td>Untested <span class='footnote'>*</span></td>
   </tr>
   <tr>
-    <td>4.x (AIO)</td>
+    <td>4.4-4.10</td>
```
ekohl, line 17: Technically the Smart Proxy should work with 4.4+ since the API hasn't changed. Do we want to advertise that or just say it's unsupported?

tbrisker: I think deprecated is fine, so we can drop it if needed

_includes/manuals/nightly/3.1.3_puppet_versions.md
```diff
@@ -1,84 +1,45 @@
...
+    <td>Untested</td>
     <td>Supported</td>
-    <td></td>
   </tr>
 </table>
 
 #### AIO installer compatibility
 
-The Foreman installer supports both AIO and non-AIO configurations when using Puppet 4/5, switching behavior automatically based on the version of Puppet installed (usually during the first run when answers are stored).
+The Foreman installer has code for both AIO and non-AIO configurations, switching behavior automatically based on the version of Puppet installed (usually during the first run when answers are stored). Only AIO installations are tested.
```
tbrisker, line 38: Perhaps we should consider dropping non-AIO compatibility? considering it isn't tested and puppetserver isn't included in non-AIO. Would that simplify the installer as well?

tbrisker (commented):
Looks good in general, happy for all the clean up of old stuff. I do not know if all the path changes are correct but I trust that you do. My main question is regarding support for non-AIO installs - should we still support that? it sounds like it's complicating things quite a lot (different settings and paths everywhere) and we don't actually test that it works. Since we require foreman to be installed on its own server, why not require AIO always even if the user uses non-AIO elsewhere?

_includes/manuals/nightly/3.1.3_puppet_versions.md
ekohl: It wouldn't simplify it a lot. We could drop some 15 lines from `params.pp` (https://github.com/theforeman/puppet-puppet/blob/10f437eb86f2dc45a470cc7ae1bdea7e33a2338a/manifests/params.pp#L133-L149) but for selfish reasons I don't want to drop them because it's how I deploy my Puppet agents.

ekohl:
I've updated 4.3.6_smartproxy_puppet.md and 4.3.7_smartproxy_puppetca.md to fully document the current configs. If this is merged, the same should be applied to 1.23.

_includes/manuals/nightly/3.1.3_puppet_versions.md
tbrisker: What about removing this option from the manual then and leaving it as an advanced option?

ekohl: I'd prefer to explicitly mention this rather than wait for questions about this. A lot of admins will assume the system packages and the Debian packages will accept both. That means if they didn't set up the AIO repositories, it'll work. The RPMs only accept the AIO package since there are no up to date non-AIO on EL and we don't package for Fedora.

_includes/manuals/nightly/3.2.3_installation_scenarios.md
```diff
@@ -59,7 +59,7 @@ Transfer the following files to the same paths on the new host:
 * /var/lib/puppet/ssl/certs/new-puppetmaster.example.com.pem
 * /var/lib/puppet/ssl/private_keys/new-puppetmaster.example.com.pem
 
-Under a Puppet 4 AIO installation, substitute above paths with `/etc/puppetlabs/puppet/ssl/`.
+Under a Puppet AIO installation, substitute above paths with `/etc/puppetlabs/puppet/ssl/`.
```
tbrisker: Shouldn't the default instructions be for AIO since that's the tested and recommended path and the comment for non-AIO?

_includes/manuals/nightly/3.1.3_puppet_versions.md
tbrisker: I would prefer to be as opinionated as possible on the installation instructions we provide our users as well as the supported setups, but I won't block on this.

_includes/manuals/nightly/3.3.1_rpm_packages.md
tbrisker: shouldn't this be `systemctl start httpd` for starting passanger?

tbrisker (changes requested):
Just a couple of small comments, otherwise looks like a great clean up!

_includes/manuals/nightly/3.2.3_installation_scenarios.md
ekohl: Good point. In other places I've swapped them and should do the same here.

It also looks like I should update this to make note of Puppet 5 vs Puppet 6 for which the instructions here are outdated.

ekohl:
I've done a partial update of the deployment scenarios but haven't tested it yet

tbrisker:
This is already a good improvement. i'm going to merge this now before branching the manual for 1.24 and we can improve as needed afterwards.

tbrisker:
Thanks @ekohl !

PR #1442, opened 2019-08-15
Fix incorrectly documented configuration option

strangeglyph:
The correct name is `authorize_login_delegation_auth_source_user_autocreate`, compare https://github.com/theforeman/foreman/blob/f2ee562ece5d5084f39160524ac57b23b7bde251/app/models/setting/auth.rb#L23


tbrisker (approved):
Thanks @strangeglyph and congrats on the first PR merged here!

PR #1450, opened 2019-08-23
Added maximum subtree hash limit setting

lzap:
SSIA

https://github.com/theforeman/foreman/pull/6850

also

https://community.theforeman.org/t/foreman-1-24-will-drop-too-many-facts/15133


tbrisker (approved):
Thanks @lzap !

PR #1457, opened 2019-09-03
Sunsetting foreman_cockpit because functionality being integrated in remote execution

dgoetz:
According to https://community.theforeman.org/t/sunsetting-foreman-cockpit/14787

### DIFF
--- a/plugins/foreman_cockpit/1.0/index.md
+++ b/plugins/foreman_cockpit/1.0/index.md
@@ -3,8 +3,12 @@ layout: plugin
 title: Foreman Cockpit 1.0 manual
 version: 1.0
 # uncomment for older versions than stable or nightly
-#robots: noindex
+robots: noindex
 ---
+
+<div class="bs-callout-info bs-callout">
+This plugin is not compatible with Foreman since Foreman 1.17 and its functionality is now integrated in <a href="plugins/foreman_remote_execution/">Foreman Remote Execution</a>.
+</div>
 
 # 1. {{ page.title }}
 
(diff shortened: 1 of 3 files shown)


ekohl:
I wonder about this. We never really had a policy around removing things. Perhaps we should add banners that it's outdated?

dgoetz:
It is actually broken because there was no release to make it compatible again and it can also break other plugins (because of alias_method_chain). This is why I would completely remove it. 

Functionality will return within REX, so we could also replace with a banner and link to REX documentation when it is added.

ekohl:
I was thinking that there still might be links to it. Displaying a big banner that it's broken since Foreman 1.x will inform them they shouldn't even try, even if we did ship it for some releases in a broken state.

tbrisker:
i tend to agree with ekohl, some people might still be using it on older versions. Would also be good to mention in the upgrade notes for 1.24.

tbrisker (approved):
thanks @dgoetz !

dgoetz:
You were faster than I while waiting for the tests to pass. I am not totally happy with it, because documentation for the new feature in REX is missing. If I remember then, I will add a more details to the banner when documentation is added.

tbrisker:
Actually, thinking about it again, can you remove the link to the plugin from the plugins list and also add the same disclaimer on the top of the 1.0 manual?

dgoetz:
I removed the link and added the disclaimer also to the versioned site. I recognized the noindex option for robots which I added to both sites, so they are hopefully not showing up when searching and instead the future documentation for Cockpit in REX will be found.

tbrisker (approved):
Thanks @dgoetz !

PR #1461, opened 2019-09-10
Add warning for changing the API response code

shiramax:


_includes/manuals/nightly/1.2_release_notes.md
```diff
@@ -21,5 +21,6 @@ This section will be updated prior to the next release.
   * duplex_
   * link_
   * wol_
+* Updating a nested resource in the API with a non existing id will no longer fail with  the response status code 404 not found , But it will fail with 422 unprocessable_entity and will present a correct error message.
```
tbrisker:
```suggestion
* Updating a nested resource in the API with a non existing id will no longer fail with the response status code `404 Not Found`, But it will instead fail with `422 Unprocessable Entity` and will present a correct error message. [#26754](https://projects.theforeman.org/issues/26754)
```

shiramax: @tbrisker done :)

_includes/manuals/nightly/1.2_release_notes.md
```diff
@@ -21,5 +21,5 @@ This section will be updated prior to the next release.
   * duplex_
   * link_
   * wol_
-
```
tbrisker, line 24: i think this empty line should be after the bullet so it renders properly

tbrisker (approved):
Thanks @shiramax !

PR #1463, opened 2019-09-12
1.23 api docs with examples

upadhyeammit:
Adding api docs with Examples.


upadhyeammit:
@tbrisker request to check. Thank You !

api/1.23/apidoc/v2/architectures.html
```diff
@@ -577,6 +679,20 @@ <h2>
...
+  &quot;architecture&quot;: {
+    &quot;name&quot;: &quot;newx86_64&quot;
+  }
+}
+403
+{
+  &quot;error&quot;: {
+    &quot;message&quot;: &quot;Access denied&quot;,
+    &quot;details&quot;: &quot;Missing one of the required permissions: edit_architectures&quot;
+  }
```
tbrisker, line 694: this looks wrong...

tbrisker (approved):
There's one example from those i sampled that looks off, but this is still much better than nothing so i'll merge this now and we can iterate on improving later.

PR #1471, opened 2019-09-25
Add note to resolve chrome version mismatch

ezr-ondrej:


contribute.md
```diff
@@ -83,6 +83,9 @@ You can also exclude these features by using `bundle install --without libvirt p
 1. Or a single test: `bundle exec bin/rake test TEST=test/functional/your_test.rb`
 1. Once done, stop any background processes with `bundle exec spring stop` ([more info](/handbook.html#UsingtheSpringpreloaderindevelopment))
 
+Our test environment requires chrome to run integration tests.
+If you run into error `Chrome version must be ...`, check your chrome version and run `npm install chromedriver@<chrome-version>`.
+
```
lzap, line 89: Our test environment requires Google Chrome to run integration tests and by default a particular chromedriver library is installed. This may not match the Chrome version installed on your system, in that case install the correct driver version: npm install chromedriver@<chrome-version>`. The error message you can encounter is: *Chrome version must be between XX and YY*.

lzap (changes requested):

contribute.md
ezr-ondrej: Yeah, much better.

lzap (approved):

lzap:
Thank you!

PR #1493, opened 2019-11-05
Add Dynflow with Sidekiq to release notes

adamruzicka:


_includes/manuals/1.24/1.2_release_notes.md
```diff
@@ -10,6 +10,26 @@ Foreman is now capable of performing Single Sign-On with the help of various Ope
 #### Debian 10 (Buster) support
 Packages for Foreman and its various plug-ins are now being built for Debian 10 (Buster) as well.
 
+#### Dynflow deployment with Sidekiq
+Foreman is now able to use Dynflow deployed with Sidekiq, as described in [Upcoming changes to Dynflow](https://community.theforeman.org/t/upcoming-changes-to-dynflow/14926). This new deployment model should address issues we have seen when restarting the Dynflow executor. This is currently included as a preview and can be enabled by hand, if desired.
+
+To enable it on EL7 and derivatives:
```
ezr-ondrej, line 18: Maybe worth noting that on debian the package is missing, but will be available in the future? Or some other note why it's not on debian available yet.

ezr-ondrej (commented):
I guess it's allright as we want only the headup on this feature. But it would make sense to move the guide/documentation to the manual for the next release.

Only thing I would note is why (or at least point that it is so) the debian is not availabe.

_includes/manuals/1.24/1.2_release_notes.md
```diff
@@ -10,6 +10,28 @@ Foreman is now capable of performing Single Sign-On with the help of various Ope
...
+yum install -y foreman-dynflow-sidekiq
+
+# Disable old Dynflow deployment, the old and new can run at once, but it isn't advised
+systemctl disable --now dynflowd
+
+# Enable the new dynflow services
+systemctl enable --now dynflow-sidekiq@{orchestrator,worker}
+```
+
+On Debian based platforms this feature is not packaged yet, but it is planned for the near future.
```
tbrisker: I would move this to the main block above, so debian users know right away not to try it.

tbrisker (approved):
Thanks @adamruzicka !

PR #1494, opened 2019-11-05
Use scopes for Katello front matter

jturel:
I'd like to switch to this method of defining section-wide values rather than each data file having its own copy. It make branching and updating this much simpler.

https://jekyllrb.com/docs/configuration/front-matter-defaults/


_data/plugins/katello.yml
```diff
@@ -0,0 +1,31 @@
+versions:
+  - name: "3.13"
+    display: "3.13"
+  - name: "3.12"
+    display: "3.12"
+  - name: "3.11"
+    display: "3.11"
+  - name: "3.10"
+    display: "3.10"
+  - name: 3.9
```
ekohl, line 10: Should we also quote this for consistency?

ekohl (approved):
I like this. Haven't gone through all the changes to verify but I'd be fine with just deploying this given you're not touching all releases.

_data/plugins/katello.yml
jturel: This would require updating the 'version' value in all the older docs markdown files so the comparison can be made correctly. Or giving them the same treatment as 3.13 and pulling the version out. For now I'd rather leave it and figure out the plan for those older docs (such as archiving them).

ekohl: Ok, fair enough.

PR #1495, opened 2019-11-07
1.24 API docs

upadhyeammit:
API docs for 1.24


upadhyeammit:
@tbrisker request to check ! thank you !

api/1.24/apidoc/v2/architectures/create.en_GB.html
```diff
@@ -0,0 +1,223 @@
...
+
+
+  <h2>Examples</h2>
+    <pre class="prettyprint">POST /api/architectures
+{
+  &quot;architecture&quot;: {
+    &quot;name&quot;: &quot;&quot;
+  }
+}
+422
```
tbrisker, line 61: this doesn't look right...

tbrisker:
Some docs are better than none, i'll merge this and we'll try to figure out why some examples are wrong and fix them.

PR #1504, opened 2019-11-21
Tuning documentation for foreman installer

sthirugn:


_includes/manuals/1.24/3.2.4_tuning_options.md
```diff
@@ -0,0 +1,54 @@
+The Foreman installer supports automatic tuning of your environment using predefined tuning profiles.  These tuning profiles are the result of a culmination of extensive learning from foreman environments deployed at scale in large customer environments.
```
ehelms: Capitalize Foreman in the last sentence. Change customer to user perhaps?

_includes/manuals/1.24/3.2.4_tuning_options.md
```diff
@@ -0,0 +1,54 @@
...
+
+When the foreman-installer is run, it is deployed with a `default` predefined tuning profile. The definition of this profile can be found in `/usr/share/foreman-installer/config/foreman.hiera/tuning/common.yaml`.
```
ehelms: Might reword to `with a predefined tuning profile named default` to make it clearer.

ehelms: Consider changing `this` to `the default profile` to make it clearer

_includes/manuals/1.24/3.2.4_tuning_options.md
```diff
@@ -0,0 +1,54 @@
...
+```bash
+foreman-installer --tuning default
+```
+
+**Note** Using the `--tuning` option above does not update `/ect/foreman-installer/custom-hiera.yml`, instead it directly updates the required configuration as specified in the corresponding tuning profile.
+
+Use the following command to identify the current tuning level:
+
+```bash
+cat /etc/foreman-installer/scenarios.d/foreman.yaml | grep 'tuning'
```
ekohl: I'd prefer to recommend `foreman-installer --help | grep tuning`.

_includes/manuals/1.24/3.2.4_tuning_options.md
```diff
@@ -0,0 +1,54 @@
...
+foreman-installer --tuning extra-extra-large
+```
+
+To reset to the default profile:
+
+```bash
+foreman-installer --tuning default
+```
+
+**Note** Using the `--tuning` option above does not update `/ect/foreman-installer/custom-hiera.yml`, instead it directly updates the required configuration as specified in the corresponding tuning profile.
```
ekohl: I think we can leave this out. This was part of the initial design but was never communicated. Adding it here will only confuse people. (There's also a typo in `/etc`)

ekohl (commented):
General note: when we document things in Foreman, they should be included in `nightly`. When we branch the next version's documentation we use that as a base. If you only document it in `1.24`, it'll be lost in `2.0`.

Since the `--tuning` parameter is only available in Katello scenarios I think the main Foreman documentation would be misleading. Shall we move it to Katello's documentation?

_includes/manuals/1.24/3.2.4_tuning_options.md
ehelms: Does that show the current value?

ekohl: Note the `default:` part.

```console
# foreman-installer --help | grep tuning
    --tuning INSTALLATION_SIZE    Tune for an installation size. Choices: default, medium, large, extra-large, extra-extra-large (default: "default")
```
After changing it to large:
```console
# foreman-installer --help | grep tuning
    --tuning INSTALLATION_SIZE    Tune for an installation size. Choices: default, medium, large, extra-large, extra-extra-large (default: "large")
```

sthirugn:
> General note: when we document things in Foreman, they should be included in `nightly`. When we branch the next version's documentation we use that as a base. If you only document it in `1.24`, it'll be lost in `2.0`.
> 
> Since the `--tuning` parameter is only available in Katello scenarios I think the main Foreman documentation would be misleading. Shall we move it to Katello's documentation?

I am not aware of the scenario differences in the installer. Let me know which repo to move this to, I can happily move it to.

_includes/manuals/1.24/3.2.4_tuning_options.md
sthirugn: oh wait, this totally beats the `default` purpose.  I thought `default` is something which is really `default`, do we really  use it to show the current configured value across the installer?

_includes/manuals/1.24/3.2.4_tuning_options.md
sthirugn: Many users are used to seeing override in custom-hiera.yml for tuning profile. Hence I added it. This was also brought out by [QE](https://bugzilla.redhat.com/show_bug.cgi?id=1772390#c0). In my new commit, I added a new that the users can still use custom-hiera.yml to override any configuration. Let me know what you think.

ekohl:
> I am not aware of the scenario differences in the installer. Let me know which repo to move this to, I can happily move it to.

Katello is documented at https://github.com/theforeman/theforeman.org/tree/gh-pages/plugins/katello.

_includes/manuals/1.24/3.2.4_tuning_options.md
ekohl: The word "default" is a bit overloaded. In this case `default:` means the default value for the command line parameter if not specified. So `default: "default"` means the default tuning is configured.

_includes/manuals/1.24/3.2.4_tuning_options.md
ekohl: Good point. It may also be good to recommend users who have previously tuned manually should review their values. They can overwrite built in tuning but is not recommended.

_includes/manuals/1.24/3.2.4_tuning_options.md
sthirugn: Updated now

sthirugn:
I addressed all the review comments except the one for moving this doc to the katello repository.

sthirugn:
> > I am not aware of the scenario differences in the installer. Let me know which repo to move this to, I can happily move it to.
> 
> Katello is documented at https://github.com/theforeman/theforeman.org/tree/gh-pages/plugins/katello.

Ewoud, foreman users would need to do tuning as well.  Should we just keep the doc here in foreman but add a note that some configuration may be purely katello related?

ekohl:
> Ewoud, foreman users would need to do tuning as well. Should we just keep the doc here in foreman but add a note that some configuration may be purely katello related?

The whole option `--tuning` doesn't exist when using the `foreman` scenario so it's not really possible to use it. On a technical level is can easily be implemented as well but due to time constraints this wasn't done.

ehelms:
Sounds like for 1.24, move the docs to Katello only since thats the only scenario this works for. But we should file an issue (https://projects.theforeman.org/projects/puppet-foreman/issues/new) to enable this for Foreman scenario for 2.0 or 2.1.

sthirugn:
2 things updated:
* Moved the tuning options doc to katello 3.14 now
* created a new [`issue`](https://projects.theforeman.org/issues/28352) to track the doc update for Foreman 2.0/2.1

plugins/katello/3.14/installation/index.md
```diff
@@ -170,6 +170,74 @@ arguments. Multiple entries can be entered via
...
+
+**Note**
+
+- Using the `--tuning` option does not update `/etc/foreman-installer/custom-hiera.yml`, instead it directly updates the required configuration as specified in the corresponding tuning profile. You can still use `custom-hiera.yml` to override any configuration if really needed.
+- If you had already used `custom-hiera.yml` and starting to use the tuned profiles, you may want to review the definition of tuned profiles (`/usr/share/foreman-installer/config/foreman.hiera/tuning/sizes/`) and remove the duplicated configuration entries from your `custom-hiera.yml`.
+- Use `foreman-installer --help | grep tuning` to identify the current tuning level.
+
+Sample output for `default` tuning:
+
+```bash
```
ekohl, line 206: On Github you can use `console` as well and use `#` (or `$`, but this should run as root) to separate the command from output:
```console
# foreman-installer --help | grep tuning
     --tuning INSTALLATION_SIZE    Tune for an installation size. Choices: default, medium, large, extra-large, extra-extra-large (default: "default")
```

Don't know if our website renders it in the same way.

plugins/katello/3.14/installation/index.md
```diff
@@ -170,6 +170,74 @@ arguments. Multiple entries can be entered via
...
+  * extra-large
+  * extra-extra-large
+
+Definitions of various tuning profiles can be found in this directory `/usr/share/foreman-installer/config/foreman.hiera/tuning/sizes/`.  Based on your environment needs, use one of the tuning profiles in the installer as shown below.
+
+```bash
+foreman-installer --tuning medium
+```
+
+```bash
```
ekohl: I'm wondering if it makes sense to specify all options since it's essentially the same.

plugins/katello/3.14/installation/index.md
```diff
@@ -170,6 +170,74 @@ arguments. Multiple entries can be entered via
...
+Sample output for `medium` tuning:
+
+```bash
+foreman-installer --help | grep tuning
+    --tuning INSTALLATION_SIZE    Tune for an installation size. Choices: default, medium, large, extra-large, extra-extra-large (default: "medium")
+```
+
+It is difficult to find the exact tuning profile for a specific environment in the first attempt because it depends on various factors like the number of managed hosts, the features used in scale (E.g., Remote Execution), the bulk action on hosts, the total amount of content, amount of host traffic to foreman, etc.  Our recommendation is that you start with the tuning profile guidance as shown in the below table based on the number of managed hosts and scale up your environment as needed.
+
+**Note** The information in the table below is just a guidance.  It is strongly recommended that you monitor the foreman environment regularly and tune up as required.
```
ekohl: https://github.com/theforeman/foreman-installer/commit/8f8f2beaf46ed1658a28b3877a571054a5c5cec6 makes these a hard lower limits and need `--disable-system-checks` to skip. Maybe worth mentioning.

ekohl (commented):
Katello has the same thing with a nightly and 3.14. I'm fine with first applying it in one place and only copying when you're done to avoid duplicating the work but something that should be fixed before merging.

plugins/katello/3.14/installation/index.md
sthirugn: I wanted to make it super clear, but it feels redundant - I can combine them - providing just one example.

plugins/katello/3.14/installation/index.md
sthirugn: yes I have seen it in other projects.  But I don't see any reference of it being used in this project. So I used the `bash` format.

plugins/katello/3.14/installation/index.md
sthirugn: good point, will add

sthirugn:
> Katello has the same thing with a nightly and 3.14. I'm fine with first applying it in one place and only copying when you're done to avoid duplicating the work but something that should be fixed before merging.

Good to know. Updated nightly now.

ehelms (approved):
@ekohl final review is yours

plugins/katello/3.14/installation/index.md
```diff
@@ -170,6 +170,65 @@ arguments. Multiple entries can be entered via
         reverse: true
         contact: hostmaster.example.com.
 
+## Tuning options
+The Foreman installer supports automatic tuning of your environment using predefined tuning profiles.  These tuning profiles are the result of a culmination of extensive learning from Foreman environments deployed at scale in large user environments.
+
+When the foreman-installer is run, it is deployed with a predefined tuning profile named `default`. The definition of the `default` profile can be found in `/usr/share/foreman-installer/config/foreman.hiera/tuning/common.yaml`.
```
ekohl: This may be a bit misleading. As it's written now it implies when `medium` is selected, `common` is no longer applied which isn't true. Common is used for all tuning. In a mathematical way: `tuning = common + profile`. Technically Hiera looks at `foreman.hiera/tuning/sizes/default.yaml` and then `foreman.hiera/tuning/common.yaml`. When a file doesn't exist, Hiera just continues.

ehelms: This is true for now unless we make an empty default yaml file. So for me it reads OK.

ekohl: I disagree. `common` is also used with `medium`. The current text implies you **only** need to look at it when you use default. Given later on we recommend users to look at their tuning, they may keep redundant tuning in `custom-hiera.yaml`.

I'd split it up and leave out how to find the exact tuning till later.

```suggestion
When the foreman-installer is run, it is deployed with a predefined tuning profile named `default`. Other than the default tuned profile, foreman-installer supports 4 different tuning profiles:
```

plugins/katello/3.14/installation/index.md
```diff
@@ -170,6 +170,65 @@ arguments. Multiple entries can be entered via
...
+
+Other than the default tuned profile, foreman-installer supports 4 different tuning profiles:
+
+  * medium
+  * large
+  * extra-large
+  * extra-extra-large
+
+Definitions of various tuning profiles can be found in this directory `/usr/share/foreman-installer/config/foreman.hiera/tuning/sizes/`.  Based on your environment needs, use one of the tuning profiles (`medium`, `large`, `extra-large`, `extra-extra-large`) in the installer.  For example, `medium` profile can be applied like:
```
ekohl:
```suggestion
Based on your environment needs, use one of the tuning profiles (`medium`, `large`, `extra-large`, `extra-extra-large`) in the installer.  For example, `medium` profile can be applied like:
```

plugins/katello/3.14/installation/index.md
```diff
@@ -170,6 +170,65 @@ arguments. Multiple entries can be entered via
...
+To reset to the default profile:
+
+```bash
+foreman-installer --tuning default
+```
+
+**Note**
+
+- Using the `--tuning` option does not update `/etc/foreman-installer/custom-hiera.yml`, instead it directly updates the required configuration as specified in the corresponding tuning profile. You can still use `custom-hiera.yml` to override any configuration if really needed.
+- If you had already used `custom-hiera.yml` and starting to use the tuned profiles, you may want to review the definition of tuned profiles (`/usr/share/foreman-installer/config/foreman.hiera/tuning/sizes/`) and remove the duplicated configuration entries from your `custom-hiera.yml`.
```
ekohl:
```suggestion
- If you had already used `custom-hiera.yml` and starting to use the tuned profiles, you may want to review the definition of tuned profiles (`/usr/share/foreman-installer/config/foreman.hiera/tuning/common.yaml` and `/usr/share/foreman-installer/config/foreman.hiera/tuning/sizes/`) and remove the duplicated configuration entries from your `custom-hiera.yml`.
```

plugins/katello/3.14/installation/index.md
```diff
@@ -170,6 +170,65 @@ arguments. Multiple entries can be entered via
...
+foreman-installer --tuning medium
+```
+
+To reset to the default profile:
+
+```bash
+foreman-installer --tuning default
+```
+
+**Note**
```
ekohl: I'd move all these notes to the bottom of the section, below the table. My reasoning is that docs should go from most relevant to least relevant. Using `custom-hiera.yaml` is unsupported and you are expected to be an advanced user already. That means it should be at the bottom.

plugins/katello/3.14/installation/index.md
ehelms: I kinda disagree with that idea as this goes back to my whole problem with the naming in the first place. The default tuning profile is common. Common is the default. Common is *always* applied so it is the defaults. When looking at say medium, I should be pointed at the medium profile as thats what makes medium _different_ than the default.

sthirugn: I have a design question on the real necessity of `foreman.hiera/tuning/common.yaml` file.
I think we should do one of the two options:
Option 1
**********
1. Strip out the changing parameters of  `foreman.hiera/tuning/common.yaml` and move them to a new `foreman.hiera/tuning/sizes/default.yaml` file.
2. And mention in docs that tuning will be always applied in this order:
   - common.yaml
   - `selected tuning profile`. For e.g., medium.yaml  (Note any configuration here overrides default .yaml)
   - custom-hiera.yaml always precedes all other tuning profiles above.

Option 2
**********
- Rename `foreman.hiera/tuning/common.yaml` to `default.yaml` and move it to `foreman.hiera/tuning/sizes`. Note: No common.yaml anymore.
- And mention in docs that tuning will be always applied in this order:
   - default.yaml
   - `selected tuning profile`. For e.g., medium.yaml  (Note any configuration here overrides default .yaml)
   - custom-hiera.yaml always precedes all other tuning profiles above.

plugins/katello/3.14/installation/index.md
sthirugn: Good point about not emphasizing too much on `custom-hiera.yaml` for entry-level. Now I created a new sub-section.  Moved the note on custom-hiera.yaml to the end of the first section (not to the very end as per your comment) because the second sub-section is mostly for choosing the right tuned profile.

plugins/katello/3.14/installation/index.md
```diff
@@ -170,6 +170,68 @@ arguments. Multiple entries can be entered via
...
+
+- Definitions of various tuning profiles can be found in this directory `/usr/share/foreman-installer/config/foreman.hiera/tuning/sizes/`.  Note that `common.yaml` is always applied and the selected tuning profile (e.g., `medium`) is applied on top and takes precedence.
+- Using the `--tuning` option does not update `/etc/foreman-installer/custom-hiera.yml`, instead it directly updates the required configuration as specified in the corresponding tuning profile. You can still use `custom-hiera.yml` to override any configuration if really needed.
+- If you had already used `custom-hiera.yml` and starting to use the tuned profiles, you may want to review the definition of tuned profiles (`/usr/share/foreman-installer/config/foreman.hiera/tuning/common.yaml` and `/usr/share/foreman-installer/config/foreman.hiera/tuning/sizes/`) and remove the duplicated configuration entries from your `custom-hiera.yml`.
+- You can also optionally use `foreman-installer --noop` to run the installer in a test mode and identify what configurations will be changed before actually running the installer.
```
ekohl: To me it's clear that this can also be combined with `--tuning <profile>` and I also know that `--noop` implies not saving the answers (including tuning). Is that also clear for the average user?

plugins/katello/3.14/installation/index.md
```diff
@@ -170,6 +170,68 @@ arguments. Multiple entries can be entered via
...
+
+### Which tuning profile should you choose?
```
ekohl, line 218: :+1: I like an explicit section here

plugins/katello/3.14/installation/index.md
```diff
@@ -170,6 +170,68 @@ arguments. Multiple entries can be entered via
...
+
+It is difficult to find the exact tuning profile for a specific environment in the first attempt because it depends on various factors like the number of managed hosts, the features used in scale (E.g., Remote Execution), the bulk action on hosts, the total amount of content, amount of host traffic to foreman, etc.  Our recommendation is that you start with the tuning profile guidance as shown in the below table based on the number of managed hosts and scale up your environment as needed.
```
ekohl: `the bulk action` looks odd. Did you mean `bulk actions`?
```suggestion
It is difficult to find the exact tuning profile for a specific environment in the first attempt because it depends on various factors like the number of managed hosts, the features used in scale (E.g., Remote Execution), bulk actions on hosts, the total amount of content, amount of host traffic to foreman, etc.  Our recommendation is that you start with the tuning profile guidance as shown in the below table based on the number of managed hosts and scale up your environment as needed.
```

ekohl (approved):
Looks good to me. Minor questions inline, but those don't need to block merging.

sthirugn:
@ekohl @ehelms thank you for the review comments. I learnt a lot from this PR. I don't have write access to this repo so if any of you can merge this please.

plugins/katello/3.14/installation/index.md
sthirugn: ah good point, I will update this.

sthirugn:
wait before merging, didnt see the last minor comments from @ekohl

plugins/katello/3.14/installation/index.md
sthirugn: done

plugins/katello/3.14/installation/index.md
sthirugn: done

sthirugn:
Updated, ready to merge.

ekohl (approved):

ekohl:
Thanks!

PR #1506, opened 2019-11-25
1.24 installer options

upadhyeammit:
This adds 1.24 installer options into manual.


upadhyeammit:
@tbrisker request to check. 
Thank You !

tbrisker (approved):
Thanks @upadhyeammit ! @ekohl can you also take a look to make sure these are all expected? they seem reasonable to me

ekohl (approved):
Yes, that's correct.

PR #1517, opened 2019-12-26
Bump minimum Ruby version for core

mmoll:


mmoll:
Is it really that one occurrence? :astonished:

ekohl (approved):
Yep, most people install from packages and there we enforce it.

PR #1520, opened 2020-01-03
Drop Debian Stretch from the nightly manual

ekohl:


mmoll:
merged, bedankt @ekohl!

PR #1528, opened 2020-02-05
add pulp3 migration docs

jlsherrill:


jlsherrill:
Note that this procedure does not currently fully work (content switchover has not been implemented yet)

johnpmitsch (approved):
:+1:

ianballou:
Is this error that I was hitting before with `katello:pulp3_migration` going to be a potential issue if Foreman isn't running? `ERROR -- /connector-database-core: No executor available (Dynflow::Error)`

Not sure if this was just an issue in the dev environment.

jlsherrill:
@ianballou thats only in a dev environment.  That issue would only happen in production if the dynflow service was not running, in which case the application isn't fully running and is an expected error condition.

PR #1535, opened 2020-02-20
Katello 3.15 update

jjeffers:
Updates for Katello 3.15.


_config.yml
```diff
@@ -9,7 +9,7 @@ url: https://theforeman.org
 baseurl: "" #keep at "" for the Atom feed
 
 # latest *stable* version
-foreman_version: '1.24'
+foreman_version: '2.0'
```
jturel: This won't get updated until Foreman 2.0 is GA and someone on the Foreman side will update it

jjeffers:
@jturel anything else we might need for this PR?

plugins/katello/3.15/release_notes/release_notes.md
```diff
@@ -0,0 +1,78 @@
...
+title: Release Notes
+---
+
+# Katello {{ page.version }} Release Notes
+
+For the full release notes, see the [Changelog](https://github.com/Katello/katello/blob/master/CHANGELOG.md).
+
+## Features
+
+### Subscriptions
```
jturel: We need to get back to having actual release notes instead of an abbreviated changelog. It's my fault I set that example last release. Maybe I intended to use what I had there as a baseline for the real notes? Who knows...

I took a shot for 3.15:

**Features**

- Improved experience for Simple Content Access users with clarified messaging and UI tweaks
- Errata can be searched by their issued date
- `hammer gpg` has been renamed to `hammer content-credentials` to match the UI
- Host updates will no longer use Dynflow for increased performance
- << Something about Pulp3? >>
- Improved performance of Subscriptions list API
- Content count information is now stored on Content View Versions directly for increased performance

**Deprecations**

- Puppet and OSTree repository types will no longer be available in Katello 3.16

@jlsherrill @jjeffers you guys might be able to articulate a pulp3 related bullet point(s) in terms a user would care about

plugins/katello/3.15/release_notes/release_notes.md
```diff
@@ -0,0 +1,78 @@
+---
+title: Release Notes
+---
+
+# Katello {{ page.version }} Release Notes
+
+For the full release notes, see the [Changelog](https://github.com/Katello/katello/blob/master/CHANGELOG.md).
```
jturel: This is a 404

plugins/katello/3.15/release_notes/release_notes.md
jturel: We may also want to mention that katello-agent is deprecated here, even though we're officially deprecating in 3.16 - I don't think it can hurt. cross-linking https://github.com/Katello/katello/pull/8588

jturel (approved):
ACK!

We can probably follow on pretty quickly with an update to the API docs. I don't think we should have any API changes by GA

PR #1538, opened 2020-03-04
update fam docs

evgeni:


ekohl (commented):
What's the `objects.inv` file? Is that an intermediate file just for building?

evgeni:
It's the "sphinx inventory", an index used to link to other sphinx pages (like docs.ansible.com and python.org/docs).

evgeni:
@ekohl updated

PR #1542, opened 2020-03-09
Drop --upgrade parameter for Katello

ehelms:


plugins/katello/nightly/upgrade/smart_proxy.md
```diff
@@ -61,10 +61,10 @@ scp ~/myproxy.example.com-certs.tar myproxy.example.com:
 
 ## Step 5 - Run Installer
 
-The installer with the --upgrade flag will run the right database migrations for all component services, as well as adjusting the configuration to reflect what's new in Katello {{ page.version }}
+The installer will run the right database migrations for all component services, as well as adjusting the configuration to reflect what's new in Katello {{ page.version }}
 
 {% highlight bash %}
-foreman-installer --scenario foreman-proxy-content --upgrade\
+foreman-installer --scenario foreman-proxy-content\
```
ekohl: There's no need to specify the scenario in an upgrade and you've dropped it in the main scenario. I'd like to standardize on "only provide the scenario on the initial install". Thoughts?

ehelms: Absolutely, thanks for catching this.

ekohl (approved):
Now only to get rid of the other `--certs-*` "upgrade" parameters :)

PR #1544, opened 2020-03-11
Foreman 2.0 apidocs

upadhyeammit:
This adds apidocs for Foreman 2.0


upadhyeammit:
@tbrisker request to check, thank you !

tbrisker (approved):
Random sampling seems to be ok. I wonder if we should start only publishing apidocs in english in the future...

tbrisker:
looks like we missed something in https://github.com/theforeman/theforeman.org#updating-api-auto-generated-docs-by-apipie, all the stylesheets and js is missing in the folder - https://theforeman.org/api/2.0/apidoc/v2.html

PR #1548, opened 2020-03-15
Fixes #28602 - Remove Smart Variables from 2.0

amirfefer:


_includes/manuals/2.0/4.2.5_matchers.md
```diff
@@ -1,5 +1,5 @@
 
-The same smart matching technology underpins both smart variables and smart class parameters, so is described below.  It provides the following features for each parameter:
+The same smart matching technology underpins smart class parameters, so is described below.  It provides the following features for each parameter:
```
tbrisker: I think the first sentence needs to be changed a bit to make sense now that it only refers to smart class params

tbrisker (changes requested):
Thanks @amirfefer ! looks good, just a small change inline. Also, we need to make the same changes in the nightly manual as well (since that will be the base for 2.1 and future manuals)

tbrisker (approved):
Thanks @amirfefer !

PR #1556, opened 2020-03-31
use auto-built foreman-ansible-modules docs

evgeni:


Rakefile
```diff
@@ -12,7 +12,7 @@ end
 
 desc 'Start server'
 task :server do
-  sh "jekyll serve --watch --incremental"
+  sh "jekyll serve --no-watch --incremental"
```
ekohl: What's the reason for this change?

evgeni: me commiting too much :)

evgeni: (watch would bomb out on my machine for some reason, blame inotify, so I disabled it)

ehelms (approved):

PR #1559, opened 2020-04-02
Update missing javascripts for apidocs

upadhyeammit:
Added missing js and templates.


tbrisker (approved):
Thanks @upadhyeammit !

PR #1564, opened 2020-04-16
Reword Katello memory requirements

ianballou:
Simply recommending 12 GB isn't sufficient since it isn't enough to sync/publish repos with lots of content, like some RHEL-related repos.

Questions:
Should 12 GB still be a baseline?
I edited only nightly and 3.15, should we go back further than that?

### DIFF
--- a/plugins/katello/3.15/installation/index.md
+++ b/plugins/katello/3.15/installation/index.md
@@ -29,7 +29,7 @@ Katello does not currently support installation on existing Foreman deployments.
 Katello may be installed onto a baremetal host or on a virtual guest.  The minimum requirements are:
 
 * Two Logical CPUs
-* 8 GB of memory (12 GB highly recommended)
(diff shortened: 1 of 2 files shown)


plugins/katello/3.15/installation/index.md
```diff
@@ -29,7 +29,7 @@ Katello does not currently support installation on existing Foreman deployments.
 Katello may be installed onto a baremetal host or on a virtual guest.  The minimum requirements are:
 
 * Two Logical CPUs
-* 8 GB of memory (12 GB highly recommended)
+* 8 GB of memory (12 GB or greater may be required depending on the amount of managed content)
```
jlsherrill: i think we should likely bump the minimum to 12 and the recommended to 16?

ianballou: Sure, I was considering that too.

ekohl (approved):

ianballou:
Can someone merge this for me? I don't have write access.

PR #1569, opened 2020-04-22
Updated a deadlink of jpegPhoto attr intro on a MSFT Technet blog

inntran:
Link to this legacy technet blog is returning 404, so I managed to find its new home.


tbrisker (approved):
Nice find, thanks @inntran !

PR #1574, opened 2020-04-27
Drop --foreman-db-type parameter documentation

ekohl:
Foreman 2.0 dropped this parameter and now only supports PostgreSQL.


evgeni:
I find the use of `foreman-initial-admin-password` as an example rather odd, as this is one of the parameters that only affects new installations. Shall we point to a different param instead?

ekohl:
I found it hard to come up with a good example and thought it would be good because this was one that a lot of users will use. Got any other suggestion?

ekohl:
@tbrisker thoughts? My reasoning for using this one is that it's a very commonly used parameter and couldn't find a better example.

ehelms (approved):

tbrisker (approved):
fine by me, if anyone comes up with a better example feel free to update it.

PR #1578, opened 2020-05-01
fix version check to not show warning on 3.15

jlsherrill:


jlsherrill:
you can see the banner at the top here: https://theforeman.org/plugins/katello/3.15/installation/index.html  

 These instructions are for installing Katello 3.15, but the latest stable is 3.15.

ekohl (approved):
Sounds like float vs string.

PR #1579, opened 2020-05-04
Removes installer HTTP proxy options as they no longer exist.

jjeffers:


plugins/katello/3.15/advanced/http_proxies.md
```diff
@@ -15,18 +15,6 @@ HTTP proxies can be created and then assigned to a product though bulk selection
 
 ## Creation
```
ekohl, line 16: Should this header be dropped and the `Creating ...` header be decreased one level?

jjeffers: I considered that, too. If no objections, lets try that.

jjeffers:
Screenshot from 2020-05-05 07-38-15 @ekohl adjusting the header level.

ekohl (approved):
Looking at the screenshot, I'd consider dropping `HTTP Proxy Support` too since there's now a duplicate header. That was already present before you started so IMHO not a blocker.

PR #1581, opened 2020-05-10
Drop old api docs

tbrisker:
Drops API docs for versions <=1.18.


lzap:
+31 −22,521,868 got me Ooopps 500  :-)

ekohl:
I've been wondering if we should have `apidoc.theforeman.org` which isn't stored in git or is its own repo. This repo is huge, partly because of it.

PR #1586, opened 2020-05-14
Katello 3.16 update

ianballou:
Questions:
1) I'm curious what people think about the release notes, did my best first pass at it
2) Did I update the API docs correctly?


_config.yml
```diff
@@ -109,7 +109,7 @@ defaults:
       path: "plugins/katello"
     values:
       layout: "plugins/katello/documentation"
-      latest: "3.15"
+      latest: "3.16"
```
jturel: This (latest) shouldn't be changed until 3.16 goes GA. It actually means 'latest stable release'

_data/plugins/katello.yml
```diff
@@ -1,4 +1,6 @@
 versions:
+  - name: "3.16"
+    display: "3.16"
```
jturel: The display should be 'RC1' here. That will show in the dropdown for the version selector

_data/plugins/katello.yml
```diff
@@ -1,4 +1,6 @@
 versions:
+  - name: "3.16"
+    display: "RC1"
```
jturel: Sorry, that should have been '3.16 RC1' precisely

plugins/katello/3.16/release_notes/release_notes.md
```diff
@@ -0,0 +1,58 @@
...
+# Katello {{ page.version }} Release Notes
+
+For the full release notes, see the [Changelog](https://github.com/Katello/katello/blob/KATELLO-3.16/CHANGELOG.md).
+
+## Features
+
+* Support for Pulpcore 3.3
+* Pulp 2 to Pulp 3 File and Docker improvements
+* Qpid replaced with ActiveMQ Artemis for messaging between Candlepin and Katello
+* "Subsciptions expiring soon" email notification customization options
```
jturel: The whole email is new so this could be 'Subscription expiration notification emails'

plugins/katello/3.16/release_notes/release_notes.md
```diff
@@ -0,0 +1,58 @@
...
+* Installation of katello-host-tools-tracer from the Katello UI
```
jturel, line 16: Bulk Traces remediation would be a good one to highlight also!

jturel:
Looks good. One quirk with the API docs: they are using the layout with the sidebar, but they should not be. Compare to a previous release to see what I mean. Not immediately sure what's up with that because we specify the correct layout in _config.yml:

```
  - scope:                                                                                                                                                                                    
      path: "plugins/katello/*/api"                                                                                                                                                           
    values:                                                                                                                                                                                   
      layout: "plugins/katello/documentation_no_menu"
```

plugins/katello/3.16/release_notes/release_notes.md
ianballou: Which commit(s) added this feature?

ianballou:
Not sure why the sidebar is there, the related code looks just like 3.15.  I'll have to poke around a bit.

jturel:
> Not sure why the sidebar is there, the related code looks just like 3.15. I'll have to poke around a bit.

Found it. The layout is being overridden in plugins/katello/3.16/api/index.md - just remove the 'layout' line

jturel:
Looks great. Let me know when RC1 is released and I'll press merge

jturel (approved):

PR #1592, opened 2020-05-20
Describe Dynflow scaling

adamruzicka:
Taken verbatim from
https://gist.github.com/adamruzicka/1991892ce22b18e030f9a4db95406319

CC @ares

### DIFF
--- a/_includes/manuals/2.0/5.1.12_dynflow_scaling.md
+++ b/_includes/manuals/2.0/5.1.12_dynflow_scaling.md
@@ -1,0 +1,172 @@
+Starting with Foreman 2.0 Foreman uses Dynflow backed by Sidekiq by default on all supported platforms. This change is described in greater detail in [Upcoming changes to Dynflow](https://community.theforeman.org/t/upcoming-changes-to-dynflow/14926). The original Dynflow executor service (called `dynflowd`) was replaced by `redis` and a set of `dynflow-sidekiq@*` services. This new deployment model should address issues we have seen when restarting the Dynflow executor. It also allows us scale more easily, dedicate workers to specific queues and increase the overall throughput. The following examples show how to scale Dynflow up in the new model.
+
+### Clean state
+
+Out of the box, foreman ships with orchestrator and a single worker
+If you have Katello, you will get an additional worker for processing of the host queue
+```
+root@modest-gator:~# ls -l /etc/foreman/dynflow/
+total 1
+-rw-r--r--. 1 root foreman 51 May 14 07:35 orchestrator.yml
+-rw-r--r--. 1 root foreman 59 May 14 07:35 worker.yml
+```
+
+The orchestrator consumes items only from the `dynflow_orchestrator` queue and has only one thread. This is by design. *Do not change this*
+```
+root@modest-gator:~# cat /etc/foreman/dynflow/orchestrator.yml
+:concurrency: 1
+:queues:
+  - dynflow_orchestrator
+```
+
+The regular worker has 5 threads and consumes items from
+default and remote_execution queues. It is not recommended
+to change the configuration for this process either, because the changes
+would be overwritten by the installer, if run again.
+
+```
+root@modest-gator:~# cat /etc/foreman/dynflow/worker.yml
+:concurrency: 5
+:queues:
+  - default
+  - remote_execution
+```
+
+### Scaling up
+Scaling up is pretty straightforward, especially if you want to only scale up what you have
+Here we use symbolic links to "share" the actual configuration among `worker`, `worker-1` and `worker-2`
+```
+root@modest-gator:/etc/foreman/dynflow# ln -s worker.yml worker-1.yml
+root@modest-gator:/etc/foreman/dynflow# ln -s worker.yml worker-2.yml
+```
+
+Check that the symbolic links are poining to the right files
+```
+root@modest-gator:/etc/foreman/dynflow# ls -l /etc/foreman/dynflow/
+total 2
+-rw-r--r--. 1 root foreman 51 May 14 07:35 orchestrator.yml
+lrwxrwxrwx. 1 root root    10 May 14 11:24 worker-1.yml -> worker.yml
+lrwxrwxrwx. 1 root root    10 May 14 11:24 worker-2.yml -> worker.yml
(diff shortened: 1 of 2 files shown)


_includes/manuals/2.0/5.1.12_dynflow_scaling.md
```diff
@@ -0,0 +1,172 @@
+Starting with Foreman 2.0 Foreman uses Dynflow backed by Sidekiq by default on all supported platforms. This change is described in greater detail in [Upcoming changes to Dynflow](https://community.theforeman.org/t/upcoming-changes-to-dynflow/14926). The original Dynflow executor service (called `dynflowd`) was replaced by `redis` and a set of `dynflow-sidekiq@*` services. This new deployment model should address issues we have seen when restarting the Dynflow executor. It also allows us scale more easily, dedicate workers to specific queues and increase the overall throughput. The following examples show how to scale Dynflow up in the new model.
+
+### Clean state
+
+Out of the box, foreman ships with orchestrator and a single worker
+If you have Katello, you will get an additional worker for processing of the host queue
+```
+root@modest-gator:~# ls -l /etc/foreman/dynflow/
```
lzap: Nitpick: I'd probably just leave `#` here.

_includes/manuals/2.0/5.1.12_dynflow_scaling.md
```diff
@@ -0,0 +1,172 @@
...
+```
+
+### Scaling down
+Scaling down is just reverse process, let's say having `worker`, `worker-1` and `worker-2` was
+an overkill and we think we would be fine with just `worker` and `worker-1`.
+```
+root@modest-gator:/etc/foreman/dynflow# systemctl disable --now dynflow-sidekiq@worker-2
+Removed /etc/systemd/system/multi-user.target.wants/dynflow-sidekiq@worker-2.service.
+
+root@modest-gator:/etc/foreman/dynflow# rm /etc/foreman/dynflow/worker-2.yml
```
lzap: I wonder why we don't actually ship 42 symlinks so users would not need to create or remove them and they could simply start and stop services. Overall I very much like this design, good work.

Anyway, I think you can leave a note that the file can be left there and won't hurt (I guess) until more instances are required.

lzap: Btw 42 was just joke, 1-9 would be probably enough. We could even create a helper command: `foreman-maintain dynflow scale-up --by 3` :-)

_includes/manuals/2.0/5.1.12_dynflow_scaling.md
```diff
@@ -0,0 +1,172 @@
...
+```
+
+The regular worker has 5 threads and consumes items from
+default and remote_execution queues. It is not recommended
+to change the configuration for this process either, because the changes
+would be overwritten by the installer, if run again.
+
+```
+root@modest-gator:~# cat /etc/foreman/dynflow/worker.yml
+:concurrency: 5
```
lzap, line 29: Out of curiousity: why 5 and not 35? If we know it's thread safe, then we should probably put more work per process.

_includes/manuals/2.0/5.1.12_dynflow_scaling.md
```diff
@@ -0,0 +1,172 @@
...
+Either open it up in your favourite `$EDITOR` or do it with `sed`.
+
+Here we can see the worker from previous example has 5 threads.
+```
+# 1785 is the process id taken from previous invocation of systemctl status
+root@modest-gator:/etc/foreman/dynflow# ps -fp 1785
+foreman     1785       1  0 11:32 ?        00:00:14 sidekiq 5.2.8  [0 of 5 busy]
+```
+
+Next we change it to 15
```
lzap, line 142: This is bad documentation honestly, we explain the process but not why you would want to do this and what are expected results. I know this is probably something our perf team should confirm, then let's resivist this later on.l

lzap (changes requested):
Other than the hostname I am good.

_includes/manuals/2.0/5.1.12_dynflow_scaling.md
adamruzicka: The original idea was to have the installer to this, but it somehow fell through the cracks. Not having to do this by hand would be nice

_includes/manuals/2.0/5.1.12_dynflow_scaling.md
adamruzicka: IIRC the original all-in-one executor had 5 threads for the default queue, so in here we just copied it over.

Threads in ruby aren't exactly a silver bullet, so more threads doesn't always mean better performance. We need to find the right balance between the number of processes and number of threads per process.

I admit 5 may be too conservative, but I'd like to wait for the performance team to run some tests before we change it.

lzap: Generally speaking, our actions (or whatever these are called in dynflow) are I/O bound so I would expect better performance from higher concurrency. Performance team should confirm. We should at least describe this down below what are the sane numbers here.

_includes/manuals/2.0/5.1.12_dynflow_scaling.md
adamruzicka: This applies to this entire chapter. I'm afraid I don't have enough data to back any claims around why anyone should do this.

lzap (approved):
Thanks it's good enough.

ekohl:
Please submit a PR that adds this to 2.1 and nightly, otherwise it gets lost.

_includes/manuals/2.0/5.1.12_dynflow_scaling.md
```diff
@@ -0,0 +1,172 @@
...
+:concurrency: 5
+:queues:
+  - default
+  - remote_execution
+```
+
+### Scaling up
+Scaling up is pretty straightforward, especially if you want to only scale up what you have
+Here we use symbolic links to "share" the actual configuration among `worker`, `worker-1` and `worker-2`
+```
```
ekohl, line 38: You can use console highlighting for all of these

adamruzicka:
Opened a PR copying this over to 2.1 and nightly manuals https://github.com/theforeman/theforeman.org/pull/1594

ares:
Thanks @adamruzicka!

PR #1595, opened 2020-05-21
Clarify character limit is 2048 and not 512

melcorr:
As reported by Foreman community member johnnys

Seems confusing to have two contradictory statements in the docs. 

Clarified with @lzap that this was accurate.

Removed a section that clarifies this from troubleshooting. 
Added it to main body of text.

### DIFF
--- a/plugins/foreman_discovery/15.0/index.md
+++ b/plugins/foreman_discovery/15.0/index.md
@@ -991,7 +991,8 @@ option, run:
 
 Keep in mind that kernel command line option is limited, Red Hat Enterprise
 Linux 7 (or CentOS 7) has limitation of 2048 characters and there are already a
-lot of characters, so in practice the script option must be shorter than this in total. When it exceeds the threshold, the rest of the kernel command
+lot of characters, so in practice the script option must be shorter than this in total. 
+When it exceeds the threshold, the rest of the kernel command
 line option will be ignored, therefore it will likely fail. Typically the script
 can be used to download a larger script using `curl` command from a remote
 HTTP(s) server.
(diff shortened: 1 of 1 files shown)


plugins/foreman_discovery/15.0/index.md
```diff
@@ -990,10 +990,9 @@ option, run:
     fdi.script=H4sIAK4hZ1oCA0tNzshXUA/JyCxWAKJEhZLU4hJ1LgA92U9qFgAAAA==
 
 Keep in mind that kernel command line option is limited, Red Hat Enterprise
-Linux 7 (or CentOS 7) has limitation of 512 characters and there are already a
-lot of characters, so in practice the script option must be shorter than 300
-characters in total. When it exceeds the threshold, rest of the kernel command
-line option will be ignored, therefore it will likely fail.Typically the script
+Linux 7 (or CentOS 7) has limitation of 2048 characters and there are already a
+lot of characters, so in practice the script option must be shorter than this in total. When it exceeds the threshold, rest of the kernel command
```
ekohl: I realize it was already already there, but:
```suggestion
lot of characters, so in practice the script option must be shorter than this in total. When it exceeds the threshold, the rest of the kernel command
```

The rest of the file uses short lines with hard breaks. Is this something you want to keep?

melcorr: It's weird. I did not touch anything in that regard and the lines render in my editor without this long line. I'll add the break.

melcorr: Done thanks!

ekohl (approved):

tbrisker:
Thanks @melcorr and @ekohl !

lzap:
For the record, it was I think bumped in 7.2. Can't remember.

PR #1599, opened 2020-05-28
Added Bayerlein to who uses foreman list

melcorr:
Deleted fork and started again 
This is to add bayerlein who added themselves to the who-uses-foreman form to be displayed on the site.

### DIFF
--- a/who_uses.md
+++ b/who_uses.md
@@ -4,7 +4,7 @@ title: Who Uses Foreman
 ---
 ## Who Uses Foreman?
 
-The following companies or individuals have confirmed use of Foreman. If you 
+The following companies or individuals have confirmed use of Foreman. If you
 want help to spread the word, [let us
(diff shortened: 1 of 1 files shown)


who_uses.md
```diff
@@ -4,14 +4,15 @@ title: Who Uses Foreman
 ---
 ## Who Uses Foreman?
 
-The following companies or individuals have confirmed use of Foreman. If you 
+The following companies or individuals have confirmed use of Foreman. If you
 want help to spread the word, [let us
 know](https://docs.google.com/a/zapletalovi.com/forms/d/1Ybz-Wz-odbGJM6Kw1pZuIOsZZ_vxXbH1I8B-iKFfuPs)
 and we will add you to the page.
```
ekohl, line 10: Not blocking this PR, but would it make sense to encourage a PR instead of using Google services that few have access to?

ekohl (approved):

who_uses.md
melcorr: @lzap ?

lzap: I wanted to avoid this to keep privacy, we can do probably both.

PR #1602, opened 2020-05-29
fix 'easy to setup them'

bish0polis:
.. because that's not English and it makes the brain itchy.


ekohl (approved):
Looks like this was copied all the way from 0.1 where it's also present. It'd be great if we fixed it all but I'm ok with just fixing the latest version.

tbrisker:
Thanks @bish0polis and welcome! looking forward to more contributions :)

PR #1605, opened 2020-06-05
Add headline feature for API fetch all

ezr-ondrej:


_includes/manuals/nightly/1.2_release_notes.md
```diff
@@ -4,6 +4,10 @@ This section will be updated prior to the next release.
 
 ### Headline features
 
+#### API pagination supports 'all'
+
+Foreman API resource listings support pagination. Foreman 2.2 introduces 'all' value for `per_page` parameter, that can be usefull on simple resources to fetch all records in one request.
```
ekohl: How about something like:
```suggestion
Foreman API resource listings support pagination. Foreman 2.2 introduces 'all' as a valid value for the `per_page` parameter which can can be used fetch all records in one request. The returned data structure is unchanged and contains pagination information for a consistent API experience. Care must be taken since some requests might be very expensive but it's a better alternative than specifying a very high number with the same goal.
```

Perhaps @melcorr has some suggestions about this as well.

ezr-ondrej: Sounds much better :+1:
Any other suggestions welcome though.

tbrisker (approved):

tbrisker:
thanks @ezr-ondrej and @ekohl !

PR #1606, opened 2020-06-05
Add scl for 2.x installer options generation

upadhyeammit:
From 2.0 we need ruby from scl and kafo-export-params also requires scl. Adding this change which is working for 1.y and 2.y both versions.

### DIFF
--- a/scripts/installer/Dockerfile
+++ b/scripts/installer/Dockerfile
@@ -4,4 +4,9 @@ RUN yum -y install epel-release https://yum.puppetlabs.com/puppet5/puppet5-relea
 
 ARG FOREMAN_VERSION
 RUN yum -y install https://yum.theforeman.org/releases/$FOREMAN_VERSION/el7/x86_64/foreman-release.rpm
+RUN yum -y install centos-release-scl-rh
 RUN yum -y install foreman-installer
+
+COPY entrypoint.sh /usr/bin/entrypoint.sh
+RUN chmod +x /usr/bin/entrypoint.sh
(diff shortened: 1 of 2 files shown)


upadhyeammit:
@ekohl request to review. Thank You !

scripts/installer/entrypoint.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+if [ -x "$(command -v scl_source)" ]; then
+  source scl_source enable tfm
+  source scl_source enable rh-ruby25
```
ekohl: tfm should already pull in the correct Ruby SCL so I think this is redundant
```suggestion
```

scripts/installer/Dockerfile
```diff
@@ -4,4 +4,9 @@ RUN yum -y install epel-release https://yum.puppetlabs.com/puppet5/puppet5-relea
 
 ARG FOREMAN_VERSION
 RUN yum -y install https://yum.theforeman.org/releases/$FOREMAN_VERSION/el7/x86_64/foreman-release.rpm
+RUN yum -y install foreman-release-scl
```
ekohl: This package was dropped in the latest version, but for the past few releases it has only depended on `centos-release-scl-rh`
```suggestion
RUN yum -y install centos-release-scl-rh
```

ekohl (changes requested):

upadhyeammit:
@ekohl replaced scl package to centos-release-scl-rh , request to check, thank you !

ekohl (approved):
Thanks!

PR #1609, opened 2020-06-08
drop ARM Debian

evgeni:


evgeni:
now includes #1610

tbrisker (approved):
Thanks @evgeni !

PR #1611, opened 2020-06-08
Katello 4.0 extended Pulp 3 deprecation warnings

ianballou:
MongoDB Storage Engine Migration Hook is being removed from the foreman installer.  I figured we should also make it more obvious that Mongo will be going away in general as well.  Related issue: https://projects.theforeman.org/issues/30006


jeremylenz (approved):
LGTM :+1:

tbrisker (approved):
Thanks @ianballou !

PR #1615, opened 2020-06-09
docs: Add some information on the common Katello path

kallisti5:
* As a new user I ran into this
* As a Foreman veteran I still run into forgetting this.
* Googling shows a *LOT* of people running into this.
* Give users guidence that if they're running Katello (common),
  they're special.


lzap (approved):
LGTM, thanks. If you want to continue contributing, feel free to check out our emerging new docs at https://docs.theforeman.org/web/

tbrisker (approved):
Thank you very much @kallisti5 ! this has been a constant source of confusion for users.

PR #1617, opened 2020-06-15
Document puppetrun removal

domitea:
Added note to deprecation about removing puppetrun

### DIFF
--- a/_includes/manuals/2.1/1.2_release_notes.md
+++ b/_includes/manuals/2.1/1.2_release_notes.md
@@ -389,6 +389,9 @@ By default it is deployed using [systemd socket activation](http://0pointer.de/b
 
 ### Deprecations
 
+* Ability to trigger puppet run for hosts from Foreman is deprecated and will be removed in version 2.2 ([more info](https://community.theforeman.org/t/dropping-run-puppet-button-with-a-replacement-using-remote-execution-plugin/12781/25)). The same functionality can be achieved using the Remote Execution plugin. There are two options:
+    * There is a `Puppet run once - SSH default` job template that takes `puppet_options` as parameter before it's invoked. There are also similar templates for other Remote Execution providers.
+    * Another option is for `mcollective`/`choria` users. The best way is to write your own job template that leverages `load_hosts` macro for choosing hosts and use it as a parameter for `mco` program, for example: `mco rpc puppetd runonce -I <%= load_hosts(search: input('hosts')).join(' ') %>`. This template would be executed against a host that has `mco` and access to the target hosts, instead of being executed at the target hosts directly.
 * Trends and statistics are being moved to separate plugin next release. See [manual](/plugins/foreman_statistics) of the new Statistics plugin.
 * Helper named `foreman_url` now requires argument which represents template kind to render. Previously, no argument was allowed to render `provision` template kind. All default templates were updated, for custom templates rewrite `foreman_url` to `foreman_url('provision')`.
 


_includes/manuals/2.1/1.2_release_notes.md
```diff
@@ -389,6 +389,7 @@ By default it is deployed using [systemd socket activation](http://0pointer.de/b
 
 ### Deprecations
 
+* Puppetrun are being removed from Foreman core.
```
lzap: Puppetrun feature is being removed...

Mabe add a little bit of reasoning and what the alternative is.

domitea:
@lzap the note has been improved. I hope that's everything is correct.

@ares / another puppet users, is the Mcollective note correct?

_includes/manuals/2.1/1.2_release_notes.md
```diff
@@ -389,6 +389,9 @@ By default it is deployed using [systemd socket activation](http://0pointer.de/b
 
 ### Deprecations
 
+* Puppetrun feature is being removed from Foreman core. This is one of many steps to achieve having Puppet related features in own plugin. However, the function of Puppetrun can be subsided by Remote Execution plugin. There is two options:
```
tbrisker: What version is it going to be removed? 2.1 still has it. The plugin will only be for extracting ENC functionality, so it's not directly related to this change (other than as part of the overall effort of making puppet optional).
Please also link to the discourse thread discussing this change for further information.

domitea:
@tbrisker The note has been updated, is it better now?

lzap (approved):

_includes/manuals/2.1/1.2_release_notes.md
```diff
@@ -389,6 +389,9 @@ By default it is deployed using [systemd socket activation](http://0pointer.de/b
 
 ### Deprecations
 
+* Puppetrun feature is being removed from Foreman core since version 2.2 ([more info](https://community.theforeman.org/t/dropping-run-puppet-button-with-a-replacement-using-remote-execution-plugin/12781/25)). However, the function of Puppetrun can be subsided by Remote Execution plugin. There is two options:
```
tbrisker: How about:
```suggestion
* Ability to trigger puppet run for hosts from Foreman is deprecated and will be removed in version 2.2 ([more info](https://community.theforeman.org/t/dropping-run-puppet-button-with-a-replacement-using-remote-execution-plugin/12781/25)). The same functionality can be achieved using the Remote Execution plugin. There are two options:
```

_includes/manuals/2.1/1.2_release_notes.md
```diff
@@ -389,6 +389,9 @@ By default it is deployed using [systemd socket activation](http://0pointer.de/b
...
+    * There is `Puppet run once - SSH default` job template that takes `puppet_options` as parameter before it's invoked
+    * Another option is for `mcollective`/`choria` users. The best way is write own job template that leverages `load_hosts` macro for choosing hosts and then this selection is used as a parameter for `mco` program, for ie. `mco ... --runs <%= load_hosts(input('search')).join(',') %>`
```
tbrisker: Is this already working? @ares  mentioned in his last post that we need to make `load_hosts` macro available for remote execution templates, I'm not sure it has been done already. It would be good if we have an example template in REX already rather than giving users an example here that hasn't been tested to work.
It should also be mentioned that this option requires triggering the job on a host that has these tools installed and has access to, not on the

_includes/manuals/2.1/1.2_release_notes.md
```diff
@@ -389,6 +389,9 @@ By default it is deployed using [systemd socket activation](http://0pointer.de/b
 
 ### Deprecations
 
+* Puppetrun feature is being removed from Foreman core since version 2.2 ([more info](https://community.theforeman.org/t/dropping-run-puppet-button-with-a-replacement-using-remote-execution-plugin/12781/25)). However, the function of Puppetrun can be subsided by Remote Execution plugin. There is two options:
+    * There is `Puppet run once - SSH default` job template that takes `puppet_options` as parameter before it's invoked
```
tbrisker: I believe there are also ansible and salt templates for the same action in those plugins

tbrisker (commented):
see inline

_includes/manuals/2.1/1.2_release_notes.md
ares: I defer to verify this to @domitea, however I think it should already be available, loader macros are included in [base scope](https://github.com/theforeman/foreman/blob/develop/lib/foreman/renderer/scope/base.rb#L7) since it's usefull for all template types.

I think example is fine if we test it can generate the list of hosts. We don't know how exactly people want to use this (mco vs choria). If we see the interest, we can build the Foreman official template once we hear, how users actually do it.

The more real life example though would be

```erb
 mco rpc puppetd runonce -I <%= load_hosts(input('hosts')).join(' ') %>
```

_includes/manuals/2.1/1.2_release_notes.md
domitea: Thanks you!

_includes/manuals/2.1/1.2_release_notes.md
domitea: The `load_hosts` macro is working in job templates because as @ares wrote, it's in base scope. I haven't tested real-life example due to my lack of puppet knowledge.

_includes/manuals/2.1/1.2_release_notes.md
domitea: Yes, you're right. But I think it's out of this PR because we're discussing here stuff related to puppet. What do you think @ares?

tbrisker: how about:
```suggestion
    * There is a `Puppet run once - SSH default` job template that takes `puppet_options` as parameter before it's invoked. There are also similar templates for other Remote Execution providers.
```

_includes/manuals/2.1/1.2_release_notes.md
```diff
@@ -389,6 +389,9 @@ By default it is deployed using [systemd socket activation](http://0pointer.de/b
 
 ### Deprecations
 
+* Ability to trigger puppet run for hosts from Foreman is deprecated and will be removed in version 2.2 ([more info](https://community.theforeman.org/t/dropping-run-puppet-button-with-a-replacement-using-remote-execution-plugin/12781/25)). The same functionality can be achieved using the Remote Execution plugin. There are two options:
+    * There is `Puppet run once - SSH default` job template that takes `puppet_options` as parameter before it's invoked
+    * Another option is for `mcollective`/`choria` users. The best way is write own job template that leverages `load_hosts` macro for choosing hosts and then this selection is used as a parameter for `mco` program, for ie. `mco rpc puppetd runonce -I <%= load_hosts(search: input('hosts')).join(' ') %>`
```
tbrisker:
```suggestion
    * Another option is for `mcollective`/`choria` users. The best way is to write your own job template that leverages `load_hosts` macro for choosing hosts and use it as a parameter for `mco` program, for example: `mco rpc puppetd runonce -I <%= load_hosts(search: input('hosts')).join(' ') %>`. This template would be executed against a host that has `mco` and access to the target hosts, instead of being executed at the target hosts directly.
```

tbrisker (changes requested):
Suggesting a few minor grammar corrections and some additional information.

tbrisker (approved):
Thanks @domitea !

PR #1618, opened 2020-06-16
Add link to provisioning guide from QSG

melcorr:
@tbrisker is this OK as a response to https://community.theforeman.org/t/foreman-2-1-0-rc3-release-process/19226/5

If so, I can add it everywhere.


tbrisker:
Looks good! let's add that just in the current manuals - 2.0, 2.1 (which you did) and nightly, I don't think we need to go back to older version.

tbrisker (approved):
Thanks @melcorr ! :medal_sports:

PR #1624, opened 2020-06-22
Make Katello installation consequence clear

melcorr:
We have a lot of new users who fall into the trap of only realising when it is too late that they would like to use Katello. I tried to make this clearer in the sentence that has been recently introduced.

### DIFF
--- a/_includes/manuals/2.0/2.1_quickstart_installation.md
+++ b/_includes/manuals/2.0/2.1_quickstart_installation.md
@@ -1,6 +1,10 @@
 [The Foreman installer](/manuals/{{page.version}}/index.html#3.2ForemanInstaller) uses Puppet **(4 or later required)** to install Foreman. This guide assumes that you have a newly installed operating system, on which the installer will setup Foreman, a Puppet master, and the [Smart Proxy](/manuals/{{page.version}}/index.html#4.3SmartProxies) by default. It's **not advisable** to follow the steps below on an existing system, since the installer will affect the configuration of several components.
 
-**If you're looking for a special scenario such as Katello content management, please follow the [Katello]({{ site.baseurl }}plugins/katello) installation instructions**
+<div class="alert alert-info">
+
+**If you want to manage content (for example, RPMs, Kickstart trees, ISO and KVM images, OSTree content, and more) with Foreman or you are a new user unfamiliar with all of Foreman's capabilities, please follow the [Katello]({{ site.baseurl }}plugins/katello) installation instructions. Installing Katello on top of an existing Foreman is unsupported.**
+
+</div>
 
 #### Select operating system
 
--- a/_includes/manuals/2.0/2.1_quickstart_installation.md
+++ b/_includes/manuals/2.0/2.1_quickstart_installation.md
@@ -2,9 +2,10 @@
 
 <div class="alert alert-info">
 
-If you want to manage content (for example, RPMs, Kickstart trees, ISO and KVM images, OSTree content, and more) with Foreman or you are a new user unfamiliar with all of Foreman's capabilities, please follow the [Katello]({{ site.baseurl }}plugins/katello) installation instructions. Installing Katello on top of an existing Foreman is unsupported.
+If you want to manage content (for example, RPMs, Kickstart trees, ISO and KVM images, OSTree content, and more) with Foreman please follow the  <a href="https://theforeman.org/plugins/katello/">Katello</a> installation instructions. If you are a new user, ensure that you familiarize yourself with  <a href="https://theforeman.org/plugins/katello/">Katello</a> and decide whether you want to use its features because installing Katello on top of an existing Foreman is unsupported. Note that you cannot install Katello on Debian systems.
 
 </div>
+
 
 #### Select operating system
 
--- a/_includes/manuals/2.0/2.1_quickstart_installation.md
+++ b/_includes/manuals/2.0/2.1_quickstart_installation.md
@@ -2,7 +2,7 @@
 
 <div class="alert alert-info">
 
-**If you want to manage content (for example, RPMs, Kickstart trees, ISO and KVM images, OSTree content, and more) with Foreman or you are a new user unfamiliar with all of Foreman's capabilities, please follow the [Katello]({{ site.baseurl }}plugins/katello) installation instructions. Installing Katello on top of an existing Foreman is unsupported.**
+If you want to manage content (for example, RPMs, Kickstart trees, ISO and KVM images, OSTree content, and more) with Foreman or you are a new user unfamiliar with all of Foreman's capabilities, please follow the [Katello]({{ site.baseurl }}plugins/katello) installation instructions. Installing Katello on top of an existing Foreman is unsupported.
 
 </div>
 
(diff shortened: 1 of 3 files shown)


_includes/manuals/2.1/2.1_quickstart_installation.md
```diff
@@ -1,6 +1,6 @@
 [The Foreman installer](/manuals/{{page.version}}/index.html#3.2ForemanInstaller) uses Puppet **(4 or later required)** to install Foreman. This guide assumes that you have a newly installed operating system, on which the installer will setup Foreman, a Puppet master, and the [Smart Proxy](/manuals/{{page.version}}/index.html#4.3SmartProxies) by default. It's **not advisable** to follow the steps below on an existing system, since the installer will affect the configuration of several components.
 
-**If you're looking for a special scenario such as Katello content management, please follow the [Katello]({{ site.baseurl }}plugins/katello) installation instructions**
+**If you want to manage content with Foreman or you're a new user unfamiliar with all of Foreman's capabilities, please follow the [Katello]({{ site.baseurl }}plugins/katello) installation instructions. If you do not install Katello with Foreman, you cannot add it at a later stage**
```
tbrisker: Perhaps we should wrap it in a colorful box like we use in other places to make sure it isn't missed: 
```suggestion
<div class="alert alert-info">
If you want to manage content with Foreman or you're a new user unfamiliar with all of Foreman's capabilities, please follow the [Katello]({{ site.baseurl }}plugins/katello) installation instructions instead. If you do not install Katello with Foreman, you cannot add it at a later stage.
</div>
```

tbrisker (commented):
Thanks! see inline, and also we'll need to apply same to nightly & 2.0 manuals

_includes/manuals/2.1/2.1_quickstart_installation.md
ekohl: I think content is too vague. If you don't know this means RPMs or other things, it means nothing. Other forms of content management can also be just doing a yum update on hosts (which REX without Katello can do).

I'd also say "Installing Katello on top of an existing Foreman is unsupported". It has been done by users. https://community.theforeman.org/t/rfc-adding-katello-to-a-foreman-install/18060 is a proposal to make this supported by adding checks for the caveats, but advanced users know how to work around that already. Of course, you wouldn't list all of that, just say "unsupported" instead of "cannot".

melcorr: Thanks!

_includes/manuals/2.0/2.1_quickstart_installation.md
```diff
@@ -1,6 +1,10 @@
 [The Foreman installer](/manuals/{{page.version}}/index.html#3.2ForemanInstaller) uses Puppet **(4 or later required)** to install Foreman. This guide assumes that you have a newly installed operating system, on which the installer will setup Foreman, a Puppet master, and the [Smart Proxy](/manuals/{{page.version}}/index.html#4.3SmartProxies) by default. It's **not advisable** to follow the steps below on an existing system, since the installer will affect the configuration of several components.
 
-**If you're looking for a special scenario such as Katello content management, please follow the [Katello]({{ site.baseurl }}plugins/katello) installation instructions**
+<div class="alert alert-info">
+
+If you want to manage content (for example, RPMs, Kickstart trees, ISO and KVM images, OSTree content, and more) with Foreman or you are a new user unfamiliar with all of Foreman's capabilities, please follow the [Katello]({{ site.baseurl }}plugins/katello) installation instructions. Installing Katello on top of an existing Foreman is unsupported.
```
ekohl: This suggests new users to always install Katello. I'm opposed to that since it's a very heavy install and in many cases no real value. Perhaps we should encourage users to familiarize themselves with Katello so they can make the consideration?

melcorr: In discussions about the new-user experience with regards to Foreman, which is versatile but complex, it was said countless times that the ideal would be that a new user installs, figures out the thing they installed Foreman to do, then starts to expand their deployment. 
When myself and @lzap were migrating docs from Satellite, we discussed this several times

Aren't we moving in the direction where Katello  eventually will be default part of the install?

 https://community.theforeman.org/t/rfc-adding-katello-to-a-foreman-install/18060
. 

I can add what you suggest.

ekohl: > Aren't we moving in the direction where Katello eventually will be default part of the install?

No, I'm IMHO Katello will never be part of the default install. Just as a comparison, you can run a vanilla Foreman with 4GB RAM. Katello + Candlepin + Pulp currently recommends 12GB I think? There is no plan to support Katello on Debian/Ubuntu but we have a decent user base who runs on that.

> https://community.theforeman.org/t/rfc-adding-katello-to-a-foreman-install/18060

That RFC talks about supporting installing Katello as a simple plugin on top of an existing Foreman install. In other words, Katello is reduced from a special case to "just" another plugin.

From the platform team we're working on removing all special cases that Katello currently has, such as different proxy port (8443 vs 9090) and certificate handling.

melcorr: OK thanks! I'll amend the note.

lzap: While Ewoud comment makes sense technically, I do not agree that we should not be encouraging users from installing it. It makes a lot of sense - if you don't know, install full project. If you know what features are you missing and you know that there is no upgrade path, go ahead and install Foreman.

One comment tho, you simply link Katello docs, but can we make sure the reader who does not click is told enough info what Katello does and what is the difference? We can simply copy and paste the basic information from Katello docs.

melcorr: I was going to link to this: https://theforeman.org/plugins/katello/ page as it has a nice overview.

ekohl: > While Ewoud comment makes sense technically, I do not agree that we should not be encouraging users from installing it. It makes a lot of sense - if you don't know, install full project. If you know what features are you missing and you know that there is no upgrade path, go ahead and install Foreman.

I can imagine a lot of users are put off by the large requirements of Katello and its workflows. If we encourage it too much and give the impression it's the recommended approach, you may lose users. I've used Foreman for 8+ years and not once have I installed Katello and actually attached any client to it. One time (about 2 or 3 years ago) I tried to set up content and actually gave up because it was too frustrating. If you don't use RHEL the whole concept of creating Products really don't resonate with anything you know. I'd want to avoid new users doing a proof of concept, trying out content and giving up when they really just wanted REX for example. On the other hand, getting familiar with Katello can be useful so you can make an informed choice.

Long term we should remove the need to make the informed choice by allowing a seamless upgrade.

So in short: :-1: for recommending to install Katello, :+1: for recommending to familiarizing and making an informed choice.

tbrisker: I agree, not everyone will want katello, let's just give the users the information and let them make an informed decision, while longer-term work on making adding katello possible.

melcorr: Thank you!

lzap:
Might worth mentioning at this point that Katello is not available for Debian systems.

lzap (approved):

tbrisker (approved):
Thanks @melcorr!

PR #1625, opened 2020-06-23
Add Salt 13.0 to plugin manual

sbernhard:


plugins/foreman_salt/13.0/index.md
```diff
@@ -0,0 +1,444 @@
...
+
+"Run Salt" button on the hosts page will trigger a `state.highstate` run.
+
+![](/static/images/plugins/foreman_salt/runsalt.png)
+
+## 4.5 Salt with Foreman Remote Execution Plugin
+
+Currenly, there are 2 ways to run salt via the remote execution plugin:
+
+  - Via a SSH tunnel: in this case SSH authentication needs to work. In the end, 'salt-call' on the smart proxy will be used to run the salt command.
```
adamruzicka: From [salt docs](https://docs.saltstack.com/en/latest/ref/cli/salt-call.html):
> The salt-call command is used to run module functions locally on a minion instead of executing them from the master.

I think proxy connects to the target host over ssh and runs `salt-call` there, no?

sbernhard: right. thanks. to many options.

sbernhard: fixed @adamruzicka

adamruzicka (approved):
LGTM

plugins/foreman_salt/13.0/index.md
```diff
@@ -0,0 +1,444 @@
...
+<pre>
+hammer salt-minion update --name minion.example.com --salt-states foo.bar
+</pre>
+
+
+# 5 Help
+
+## 5.1 Troubleshooting
+
+If you find a bug, please file it in [Redmine](http://projects.theforeman.org/projects/salt/issues/new).  You can find us on freenode on #theforeman as well.
```
ekohl: I'm always a fan of linking to https when available to avoid the http redirect
```suggestion
If you find a bug, please file it in [Redmine](https://projects.theforeman.org/projects/salt/issues/new).  You can find us on freenode on #theforeman as well.
```

ekohl (approved):

PR #1571, opened 2020-04-22
Drop foreman-release-scl and tfm-ror52

ehelms:

PR #1630, opened 2020-06-24
fix foreman/katello nightly installation instructions

wbclark:
`foreman-release-scl` has been dropped in favor of `centos-release-scl-rh`


ekohl:
This should be fixed via https://github.com/theforeman/theforeman.org/pull/1571. I have one more question for @ehelms there, but I'd prefer to merge that instead.

PR #1633, opened 2020-07-01
Fix tuning guide table

ehelms:


evgeni:
How about fixing the markdown instead? :)

https://github.com/theforeman/theforeman.org/pull/1634

ekohl:
Closed in favor of #1634

ehelms:
Well, y'all were a bit quick here. I can see the argument for using the
markdown style. I made the switch in part because up above in the file an
html table is used to list supported OSes. So I aimed for consistency.

evgeni:
@ehelms I'd argue the list of supported OSes should be that, a list, not a table?

evgeni:
see https://github.com/theforeman/theforeman.org/pull/1635

PR #1644, opened 2020-07-08
Use the full CC BY-SA 3.0 license text

ekohl:
This uses the full license text so Github can properly detect it. Copied from https://creativecommons.org/licenses/by-sa/3.0/legalcode.txt


evgeni (approved):

ekohl:
Hmm, looks like it still isn't detected. Never mind then.

PR #1654, opened 2020-07-29
Add CVE-2020-14334 to security page

ezr-ondrej:


security.md
```diff
@@ -80,6 +81,15 @@ All security advisories made for Foreman are listed below with their correspondi
 
 ### Disclosure details
 
+#### <a id="2020-14334"></a>CVE-2020-14334: World readable cache could expose sensitive settigs
+Even encrypted settings have their raw values cached. Too permissive mode on cache dir caused, that anyone with access to the hosting system could read this encrypted settings.
+
+*Mitigation:* change cache directory mode `o-r`.
+
+* Affects RPM installations
+* Fix released in Foreman 2.1
```
tbrisker: the fix will be in Foreman 2.2.0, 2.1.1 and 2.0.2 and higher

security.md
```diff
@@ -80,6 +81,15 @@ All security advisories made for Foreman are listed below with their correspondi
 
 ### Disclosure details
 
+#### <a id="2020-14334"></a>CVE-2020-14334: World readable cache could expose sensitive settigs
+Even encrypted settings have their raw values cached. Too permissive mode on cache dir caused, that anyone with access to the hosting system could read this encrypted settings.
+
+*Mitigation:* change cache directory mode `o-r`.
```
tbrisker: please specify which directories

security.md
```diff
@@ -80,6 +81,15 @@ All security advisories made for Foreman are listed below with their correspondi
...
+
+* Affects RPM installations
```
tbrisker: since which foreman version?

tbrisker (changes requested):

security.md
ekohl: It was part of https://github.com/theforeman/foreman-packaging/commit/847450fab873380fc50af885c67735f8f7abf05e which is titled `Import foreman`. Prior to that RPM specs were in foreman.git and was introduced in https://github.com/theforeman/foreman/commit/d30ac557d1d260bfdae13164cd83deca27c371cc. So my guess is Foreman 1.3 when using RPMs and the file cache backend. That means users of memcached are unaffected by this particular bug, though that does listen on a TCP port which is usually also accessible by local users.

ezr-ondrej: Thanks :heart: I was not able to find it out, I did not know about the switch to packaging repo :)

ezr-ondrej: Altough back then there might not be any sensitive information in the cache... :thinking:

security.md
```diff
@@ -80,6 +81,15 @@ All security advisories made for Foreman are listed below with their correspondi
 
 ### Disclosure details
 
+#### <a id="2020-14334"></a>CVE-2020-14334: World readable cache could expose sensitive settigs
+Even encrypted settings have their raw values cached. Too permissive mode on cache dir caused, that anyone with access to the hosting system could read this encrypted settings.
+
+*Mitigation:* change cache directory mode `o-r` on `/run/foreman`.
```
ekohl: This is not a correct mitigation because it's lost at reboot. You must manually change the tmpfiles as well.

security.md
ekohl: OAuth was added in https://github.com/theforeman/foreman/commit/fceb1f895c20019ad71863d23316a446c216845e, released in version 1.1 and already used settings. Caching of settings was added in https://github.com/theforeman/foreman/commit/d9a2ebac6e9ec0082e60265b794f54a29a1f7e65, also in 1.1.

ekohl:
Do we want to suggest to rotate OAuth keys, or at least provide instructions? Are there other secrets there that might have leaked?

security.md
ezr-ondrej: There are some passwords, those should be considered compromised as well if there was a possibility to use this vulnerability.
https://github.com/search?q=org%3Atheforeman+%22encrypted+%3D%3E+true%22&type=Code

ezr-ondrej:
I've tried to suggest the rotation of OAuth keys and passwords, but I'm not sure if we should provide more information.

List of settings that might have been affected: https://github.com/search?q=org%3Atheforeman+%22encrypted+%3D%3E+true%22&type=Code

security.md
ezr-ondrej: Change that to advise the temfile rewrite from `/etc/tmpfiles.d`

security.md
```diff
@@ -80,6 +81,21 @@ All security advisories made for Foreman are listed below with their correspondi
...
+#### <a id="2020-14334"></a>CVE-2020-14334: World readable cache could expose sensitive settigs
+Even encrypted settings have their raw values cached. Too permissive mode on cache dir caused, that anyone with access to the hosting system could read this encrypted settings.
+
+*Mitigation:* override directory mode to `750`.
+To do so, create a file `/etc/tmpfiles.d/foreman.conf` with following content and reboot the system.
+```
+d /run/foreman 0750 foreman foreman -
+```
+
+Additionaly it's better to rotate the OAuth keys and Remote Execution passwords
```
tbrisker:
```suggestion
In case the system may have been accessed locally by an un-trusted user, it may be prudent to change and secrets stored in the settings, such as OAuth keys or remote execution passwords.
```

security.md
```diff
@@ -80,6 +81,21 @@ All security advisories made for Foreman are listed below with their correspondi
 
 ### Disclosure details
 
+#### <a id="2020-14334"></a>CVE-2020-14334: World readable cache could expose sensitive settigs
+Even encrypted settings have their raw values cached. Too permissive mode on cache dir caused, that anyone with access to the hosting system could read this encrypted settings.
+
+*Mitigation:* override directory mode to `750`.
+To do so, create a file `/etc/tmpfiles.d/foreman.conf` with following content and reboot the system.
```
tbrisker: This file should already be present if i'm not mistaken? Also, iiuc this modification is only needed so the change survives reboots, the user can also manually `chmod` the folder for the currently running instance without requiring a full reboot.

ekohl: No, we ship to `/usr/lib/tmpfiles.d/foreman.conf`. The downside of creating `/etc/tmpfiles.d/foreman.conf` is that if we want to make further changes in the future, then users won't get them. However, users probably also shouldn't change `/usr/lib`.

Changing in `/usr/lib/tmpfiles.d` means they get updated in the future, but if they update from unsupported-x.y to unsupported-x.z then they lose the protection. OTOH, they should just update to a supported version.

I'm conflicted on the best advise, but slightly leaning to `/usr/lib/tmpfiles.d`.

@theforeman/packaging what would you advise?

evgeni: etc is for overrides if package defaults, and thats exactly whats needed here. So IMHO etc and the hint to remove that file after upgrading to a fixed release.

dgoetz: systemd-tmpfiles.d does not allow for drop-ins like with unit files, but it explicitly mentions splitting the configuration for easier overwriting of configuration.

```
Each configuration file shall be named in the style of package.conf or package-part.conf. The second variant should be used when it is desirable to make it easy to override just this part of configuration. 
```

So perhaps it would make sense to split it up and store values in `/usr/lib/tmpfiles.d` where packages should install the files? Not sure if this would mess things up, but perhaps an idea for the future. But I am also pro `/usr/lib/tmpfiles.d`.

For those already having created local configuration in `/etc/tmpfiles.d`, it makes sense to advice them to adjust configuration, but I would leave this to documentation and perhaps it is possible to have a migration in the foreman-installer which checks for the CVE, links to it and provides a simple sed for mitigation?

ezr-ondrej: I updated the suggestion to update of the `/usr/lib/tmpfile.d` as @ekohl is mentioning, the next upgrade should be to a supported version, so this solution will leave the least traces in the system.

security.md
```diff
@@ -80,6 +81,21 @@ All security advisories made for Foreman are listed below with their correspondi
...
 
+#### <a id="2020-14334"></a>CVE-2020-14334: World readable cache could expose sensitive settigs
+Even encrypted settings have their raw values cached. Too permissive mode on cache dir caused, that anyone with access to the hosting system could read this encrypted settings.
+
+*Mitigation:* override `/run/foreman` directory mode to `750`.
+To do so in a manner that survives reboot, update a file `/usr/lib/tmpfiles.d/foreman.conf`.
+```
+d /run/foreman 0750 foreman foreman -
+```
+
```
lzap, line 93: This is not enough, the permission will be fixed on next reboot or after 24 hours when the systemd timer fires off. I suggest to also recommend changing the permissions immediately on the directory itself using `chmod`. Look:

```
[root@box lzap]# ll -ad /tmp/test
drwx------. 2 root root 40 10. srp 09.17 /tmp/test
[root@box lzap]# cat /usr/lib/tmpfiles.d/test.conf
d /tmp/test 0700 root root -
[root@box lzap]# vim /usr/lib/tmpfiles.d/test.conf
[root@box lzap]# ll -ad /tmp/test
drwx------. 2 root root 40 10. srp 09.17 /tmp/test
[root@box lzap]# systemd-tmpfiles --create
[root@box lzap]# ll -ad /tmp/test
drwxrwxrwx. 2 root root 40 10. srp 09.17 /tmp/test
```

security.md
```diff
@@ -80,6 +81,22 @@ All security advisories made for Foreman are listed below with their correspondi
...
 
+#### <a id="2020-14334"></a>CVE-2020-14334: World readable cache could expose sensitive settigs
+Even encrypted settings have their raw values cached. Too permissive mode on cache dir caused, that anyone with access to the hosting system could read this encrypted settings.
+
+*Mitigation:* override `/run/foreman` directory mode to `0750`.
+To do so in a manner that survives reboot, update a file `/usr/lib/tmpfiles.d/foreman.conf`.
+```
+d /run/foreman 0750 foreman foreman -
+```
+For the change to have effect immediatelly run `systemd-tmpfiles --create` or change the directory mode manually by running `chmod 755 /run/foreman`.
```
ekohl: I like the `systemd-tmpfiles --create` suggestion since you verify the state post-reboot. I'd leave off the chmod command. Also note you have `755` in the current command

ezr-ondrej: Dropping the `chmod` :+1:

lzap (approved):

security.md
```diff
@@ -80,6 +81,22 @@ All security advisories made for Foreman are listed below with their correspondi
 
 ### Disclosure details
 
+#### <a id="2020-14334"></a>CVE-2020-14334: World readable cache could expose sensitive settigs
+Even encrypted settings have their raw values cached. Too permissive mode on cache dir caused, that anyone with access to the hosting system could read this encrypted settings.
+
+*Mitigation:* override `/run/foreman` directory mode to `0750`.
+To do so in a manner that survives reboot, update a file `/usr/lib/tmpfiles.d/foreman.conf`.
```
ekohl: It feels more natural to use the definitive article here since we know what we're talking about.
```suggestion
To do so in a manner that survives reboot, update the file `/usr/lib/tmpfiles.d/foreman.conf`.
```

ekohl (approved):
Small language nit, otherwise :+1:

security.md
ezr-ondrej: :+1:  thanks, commited.

ezr-ondrej:
@tbrisker, any more comments? :)

security.md
```diff
@@ -80,6 +81,22 @@ All security advisories made for Foreman are listed below with their correspondi
...
+Even encrypted settings have their raw values cached. Too permissive mode on cache dir caused, that anyone with access to the hosting system could read this encrypted settings.
+
+*Mitigation:* override `/run/foreman` directory mode to `0750`.
+To do so in a manner that survives reboot, update the file `/usr/lib/tmpfiles.d/foreman.conf`.
+```
+d /run/foreman 0750 foreman foreman -
+```
+For the change to have effect immediatelly run `systemd-tmpfiles --create`.
+
+In case the system may have been accessed locally by an un-trusted user, it may be prudent to change and secrets stored in the settings, such as OAuth keys or remote execution passwords.
```
tbrisker: just a tiny typo:
```suggestion
In case the system may have been accessed locally by an un-trusted user, it may be prudent to change any secrets stored in the settings, such as OAuth keys or remote execution passwords.
```

ezr-ondrej: :+1: Done.

tbrisker (approved):
Thanks @ezr-ondrej !

PR #1659, opened 2020-08-03
Add Foreman 2.2 signing key

ehelms:


security.md
```diff
@@ -932,6 +932,15 @@ Key management is changing at the time of writing to cycle nightly keys every tw
         <td style='white-space:nowrap'></td>
         <td></td>
       </tr>
+      <tr>
+        <td><a href="{{ site.baseurl }}/static/keys/939E2FD66DF9A4104869FD182A63B068C7C509FA.pub">939E2FD66DF9A4104869FD182A63B068C7C509FA</a></td>
+        <td>FIXME: fingerprint</td>
```
ekohl: I thought FIXME's would be clear enough because people would review a diff but I guess that's assumption I shouldn't have made.

ehelms: This would have helped me catch it -- https://github.com/theforeman/theforeman-rel-eng/pull/54

tbrisker (approved):
Thanks @ehelms !

ekohl:
I verified I can also retrieve the same key from the store, meaning it's not just on @ehelms' machine.

PR #1664, opened 2020-08-15
Fix 2.1 release notes spelling errors

alexjfisher:
Other than the spelling of `implementation`, the other errors came from spelling mistakes in ticket titles.  I don't have the privileges to fix these, but perhaps someone else would?


alexjfisher:
FWIW, I also checked the 2.2 release notes.  I assume they'll be regenerated prior to the final release.  As such, could someone fix the spelling of `component` in https://projects.theforeman.org/issues/29900 title?

ekohl (approved):
Thanks! I've updated the Redmine issues as well.

PR #1665, opened 2020-08-15
Fix more manual spelling mistakes

alexjfisher:


ekohl (approved):
Thanks! I can tell you that the GH UI doesn't make it easy to review so many files ;)

PR #1668, opened 2020-08-26
Update Katello broken SCL rpm name

chris1984:
When testing Katello 3.17 and an upgrade from Katello 3.16 manually, it looks like `foreman-release-scl` no longer exists. I looked on the Foreman nightly docs and it has been replaced with `epel-release centos-release-scl-rh` which allows the install of of rpms to complete.


plugins/katello/3.16/installation/index.md
```diff
@@ -79,7 +79,7 @@ yum -y localinstall https://yum.theforeman.org/releases/{{ page.foreman_version
 yum -y localinstall https://fedorapeople.org/groups/katello/releases/yum/{{ page.version }}/katello/el7/x86_64/katello-repos-latest.rpm
 yum -y localinstall https://yum.puppet.com/puppet6-release-el-7.noarch.rpm
 yum -y localinstall https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
```
ekohl: This can be dropped if you already have `epel-release` below.

chris1984:
@ekohl fixed and tested, thank you for the review.

ekohl (approved):

PR #1672, opened 2020-09-04
removing puppetrun configuration

domitea:


ekohl (changes requested):
Could you add a release note to this? Also pending merging of https://github.com/theforeman/smart-proxy/pull/762

On a related note, looks like the puppetrun setting that was removed in 2.2 is still documented. Also left a comment @ https://github.com/theforeman/foreman/pull/7719#issuecomment-687111669.

tbrisker:
(oops, ewoud disregard the review request, clicked on the wrong PR)

domitea:
Docs updated, where is best place to write about removing docs in release notes? My suggest is Deprecations but I'm not sure. What do you think @tbrisker,  @ekohl ?

tbrisker:
We already mentioned removal of puppetrun in 2.2 upgrade warnings - https://github.com/theforeman/theforeman.org/blob/gh-pages/_includes/manuals/2.2/1.2_release_notes.md#upgrade-warnings, so I'm not sure we need another warning about removing it from the smart proxy as well. what do you think @ekohl ?

ekohl (approved):

PR #1685, opened 2020-09-29
Update Katello 3.17 display to RC2

chris1984:
When 3.17 RC2 get published by the release team


chris1984:
@tbrisker can this get a merge or do you want it to say rc2.2 which is the gem version?

tbrisker (approved):
I'm fine with just rc2

PR #1699, opened 2020-10-14
Prepare manual for 2.1.4 release

tbrisker:


_includes/manuals/2.1/1.2_release_notes.md
```diff
@@ -28,6 +28,21 @@ By default it is deployed using [systemd socket activation](http://0pointer.de/b
 ### Known Issues
 * The version of `rubygem-net-ssh` package required by Foreman is incompatible with the version provided EPEL8. Users who wish to install Foreman on EL8 should ensure that EPEL repository is disabled or that the conflicting package is excluded.
 * During upgrade testing, there have been a few cases where a 502 error was encountered when the installer tried to register the integrated Smart Proxy to Foreman. While we believe we have likely resolved the root cause, if this error is encountered, try re-running the installer again.
+* External authentication does not work properly with the Puma web server. This will be [fixed](https://projects.theforeman.org/issues/30535) in Foreman 2.2. In the mean time, users requiring external authentication should continue using the Passenger web server (which can be done by running the installer with the `--foreman-passenger` flag).
```
melcorr:
```suggestion
* External authentication does not work properly with the Puma web server. This will be [fixed](https://projects.theforeman.org/issues/30535) in Foreman 2.2. In the meantime, users requiring external authentication should continue using the Passenger web server (which can be done by running the installer with the `--foreman-passenger` flag).
```

melcorr (commented):
just spotted one small thing

_includes/manuals/2.1/1.2_release_notes.md
ekohl: Is Puma a webserver or an application server? I think at least Passenger should be considered as an application server.

Another thing to note is that when you just run `--foreman-passenger true` that you need to replace `/etc/httpd/conf.d/passenger.conf`. This file is managed by the `mod_passenger` package so `yum reinstall mod_passenger` may be needed. I think that's only true for RPM based installations, not sure about Debian based.

tbrisker: TBH I don't think the distinction between web server and app server is very clear; puma website calls it "A Fast, Concurrent Web Server for Ruby & Rack". Passenger website states "Passenger® is an app server that runs and automanages your web apps with ease.", but on github it's listed as "A fast and robust web server and application server for Ruby, Python and Node.js "
regarding `/etc/httpd/conf.d/passenger.conf`, do we have that process described anywhere? isn't it done by the installer?

ekohl: >  regarding `/etc/httpd/conf.d/passenger.conf`, do we have that process described anywhere? isn't it done by the installer?

No, it contains exact version numbers so it broke all the time. That's why the file content is unmanaged. However, the file is managed so it's not purged. When the Passenger module is not enabled, it is purged. Then when you enable Passenger again, it can't be created again.

A pre hook could solve this, but nobody has taken time to write one.

evgeni:
FWIW, Puma and Passenger are IMHO application servers, not web servers, but even the puma homepage disagrees with me ;)

_includes/manuals/2.1/1.2_release_notes.md
tbrisker: @ekohl Updated, I hope the wording is clear enough. also dropped `web` and went with `puma/passenger server` instead

_includes/manuals/2.1/1.2_release_notes.md
```diff
@@ -28,6 +28,21 @@ By default it is deployed using [systemd socket activation](http://0pointer.de/b
...
+#### Foreman - Packaging
+* Foreman doesn't start with Puma 5.0 ([#30916](https://projects.theforeman.org/issues/30916))
+
+#### Foreman - Rake tasks
+* installation fails at execution of '/usr/sbin/foreman-rake -- config -k 'remote_execution_cockpit_url' -v '/webcon/=%{host}'' ([#30932](https://projects.theforeman.org/issues/30932))
+
+#### Installer - Foreman modules
+* Pulpcore PostgreSQL database is not enforced to UTF-8 ([#31018](https://projects.theforeman.org/issues/31018))
+* dhcpd.conf file-ACLs are reset ([#30962](https://projects.theforeman.org/issues/30962))
+* Possible connection pool exhaustion ([#30789](https://projects.theforeman.org/issues/30789))
```
ekohl: Updated the Redmine
```suggestion
* Possible database connection pool exhaustion ([#30789](https://projects.theforeman.org/issues/30789))
```

ekohl (commented):
Other than the 1 inline comment, :+1:

zjhuntin (approved):

PR #1700, opened 2020-10-14
Update manual for puma as default web server

tbrisker:


tbrisker:
There's probably more that can be cleaned up, but this is a first step.

_includes/manuals/2.1/3.6_upgrade.md
```diff
@@ -165,7 +165,7 @@ To apply these changes, run the installer again without options
 
 #### Step 4 - Restart
 
-Restart the application server (e.g. mongrel, thin, passenger).
+Restart the application server (e.g. puma, passenger).
```
lzap, line 168: Can you remove passenger from here to align with the rest of the changes? We shouldn't be advertising what we don't recommend.

_includes/manuals/2.2/3.6_upgrade.md
```diff
@@ -165,7 +165,7 @@ To apply these changes, run the installer again without options
 
 #### Step 4 - Restart
 
-Restart the application server (e.g. mongrel, thin, passenger).
+Restart the application server (e.g. puma, passenger).
```
lzap, line 168: Ditto.

_includes/manuals/nightly/3.6_upgrade.md
```diff
@@ -165,7 +165,7 @@ To apply these changes, run the installer again without options
 
 #### Step 4 - Restart
 
-Restart the application server (e.g. mongrel, thin, passenger).
+Restart the application server (e.g. puma, passenger).
```
lzap, line 168: And here as well.

lzap (changes requested):
Minor comment.

_includes/manuals/2.1/2_quickstart_guide.md
```diff
@@ -1,12 +1,14 @@
 
 The Foreman installer is a collection of Puppet modules that installs everything required for a full working Foreman setup.  It uses native OS packaging (e.g. RPM and .deb packages) and adds necessary configuration for the complete installation.
 
-Components include the Foreman web UI, Smart Proxy, Passenger, a Puppet master (either Puppet Server or under Passenger), and optionally TFTP, DNS and DHCP servers.  It is configurable and the Puppet modules can be read or run in "no-op" mode to see what changes it will make.
+Components include the Foreman web UI, Smart Proxy, Puma, a Puppet Server, and optionally TFTP, DNS and DHCP servers.  It is configurable and the Puppet modules can be read or run in "no-op" mode to see what changes it will make.
 
 #### Supported platforms
 * CentOS 7 x86_64
+* CentOS 8 x86_64 (except for Katello plugin)
```
ekohl, line 8: Normally we don't list support for plugins. Katello users shouldn't even read this manual since they have their own installation instructions. :-1: for doing this.

_includes/manuals/2.1/3.2.2_installer_options.md
```diff
@@ -293,7 +293,7 @@ kafo-export-params -c /etc/foreman-installer/scenarios.d/foreman.yaml -f html
     <tbody>
       <tr>
         <td style='white-space:nowrap'>--foreman-apache</td>
-        <td>Configure Foreman via Apache. By default via passenger but otherwise as a reverse proxy.</td>
+        <td>Configure Foreman via Apache. By default via puma with reverse proxy.</td>
```
ekohl, line 296: Do we want to divert from the actual parameters as part of the installer?

_includes/manuals/2.1/3.2_foreman_installer.md
```diff
@@ -10,7 +10,7 @@ By default it will configure:
 * Apache HTTP with SSL (using a Puppet-signed certificate)
 * Foreman running under Puma (using systemd socket activation)
 * Smart Proxy with SSL (using a Puppet-signed certificate) configured for Puppet and TFTP
-* Puppetserver
+* Puppet server
```
ekohl, line 13: I still find it difficult to follow. First it was Puppet master, Puppetserver and now https://puppet.com/docs/puppet/6.18/server.html refers to Puppet Server. I think here we should follow the capitalization
```suggestion
* Puppet Server
```

_includes/manuals/2.1/3.2.2_installer_options.md
```diff
@@ -617,7 +617,7 @@ kafo-export-params -c /etc/foreman-installer/scenarios.d/foreman.yaml -f html
       </tr>
       <tr>
         <td style='white-space:nowrap'>--foreman-ssl</td>
-        <td>Enable and set require_ssl in Foreman settings (note: requires passenger, SSL does not apply to kickstarts)</td>
+        <td>Enable and set require_ssl in Foreman settings (note: requires apache, SSL does not apply to kickstarts)</td>
```
ekohl, line 620: I it's typically Apache, but it hasn't been updated yet in the actual parameter.
https://github.com/theforeman/puppet-foreman/blob/495727bdfe4a60df28c542cf5b771a91e20433b3/manifests/init.pp#L64

_includes/manuals/2.1/3.3.1_rpm_packages.md
```diff
@@ -6,7 +6,7 @@ Foreman is packaged for the following RPM based distributions:
 
 **Note:** The RPM packages are not tested on Scientific Linux or Oracle Linux. The Foreman installation on Scientific Linux or Oracle Linux may or may not work.
 
-For most users, it's highly recommended to use the [installer](manuals/{{page.version}}/index.html#3.2ForemanInstaller) as the packages only provide the software and a standalone Foreman service.  The installer installs these packages, then additionally configures Foreman to run under Apache and Passenger with PostgreSQL, plus can configure a complete Puppet setup integrated with Foreman.
+For most users, it's highly recommended to use the [installer](manuals/{{page.version}}/index.html#3.2ForemanInstaller) as the packages only provide the software and a standalone Foreman service.  The installer installs these packages, then additionally configures Foreman to run under Apache and Puma with PostgreSQL, plus can configure a complete Puppet setup integrated with Foreman.
```
ekohl, line 9: To me 'run under Apache' feels unnatural since Puma is the application server that actually runs it. Does something like this make sense?
```suggestion
For most users, it's highly recommended to use the [installer](manuals/{{page.version}}/index.html#3.2ForemanInstaller) as the packages only provide the software and a standalone Foreman service.  The installer installs these packages, then additionally configures Foreman to run under Puma with Apache as a reverse proxy and PostgreSQL, plus can configure a complete Puppet setup integrated with Foreman.
```

_includes/manuals/2.1/3.5.1_initial_setup.md
```diff
@@ -22,21 +22,6 @@ foreman-rake db:seed</pre>
...
 [here](manuals/{{page.version}}/index.html#3.5.3DatabaseSetup)
 
-#### Import Data from Puppet
-
-At this point, you might want to go through the [[FAQ]] to see how can you import your data into Foreman.
-
-#### Start The Web Server
-
-if you installed via rpm, just start the foreman service, or start the builtin web server by typing:
-<pre>RAILS_ENV=production rails server</pre>
```
ekohl, line 32: For what it's worth, `systemctl start foreman` would be the modern equivalent if you installed the `foreman-service` package.

_includes/manuals/2.1/3.6_upgrade.md
ekohl: Passenger is still supported and there is an installer option to use it.

My bigger issue is that the instructions below it (to restart Apache) are incorrect for Puma. It should really be `systemctl restart foreman`.

ekohl:
I also just noticed we use `foreman-maintain` in the upgrade manual (since 1.23 / 7a5e16e0f416f529162c0c6f8de7ee9034462992). However, that's unavailable on Debian. I still strongly feel we should not mention foreman-maintain at all until Debian is supported. I also work hard to keep it completely optional in the installer.

PR #1704, opened 2020-10-19
drop rackspace from 2.1+ manuals

evgeni:


ekohl (commented):
If you're renaming sections (5.2.9 -> 5.2.8), the code that refers to it should also be updated:
https://github.com/theforeman/foreman/blob/d00fc37a68839f8bf823969880339c9343eacd4c/app/views/compute_resources/form/_vmware.html.erb#L1

evgeni:
damn.

ekohl:
You can cheat and move 5.2.10 to 5.2.8 but that gives a weird order. For existing releases you can also keep 5.2.8 with `The Rackspace compute resource has been removed` and only fix it for nightly (and possibly 2.2).

evgeni:
Yeah, using "was removed" for 2.1 sounds good to me.

@tbrisker opinions?

tbrisker:
Sounds good to me

evgeni:
Updated to not break 2.1 docs. develop renumbering is in https://github.com/theforeman/foreman/pull/8081

tbrisker (approved):
lgtm. @ekohl ?

ekohl (approved):

PR #1709, opened 2020-10-21
Update RHEL installation docs

ekohl:
511397791 Mention the requirement to enable RHSCL for RHEL. This is already part of the quickstart but not the real installation section.
e90f34d1b Drop mention of RH Satellite 5. This is EOL and should no longer be used.


tbrisker (approved):
Thanks @ekohl !

PR #1719, opened 2020-11-04
update katello client installation to include el8 and not include f29

evgeni:
also drop el5 from nightly

### DIFF
--- a/plugins/katello/nightly/installation/clients.md
+++ b/plugins/katello/nightly/installation/clients.md
@@ -14,21 +14,13 @@ Install the appropriate Katello client release packages.
 <p>
   Select your Operating System:
   <select id="operatingSystems">
-     <option value="el5">Enterprise Linux 5 (CentOS, etc.)</option>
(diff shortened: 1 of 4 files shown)


plugins/katello/nightly/installation/clients.md
```diff
@@ -68,7 +60,7 @@ Now you are ready to install the client package:
 
 The `katello-host-tools` package reports errata & package profile information, but does not allow you to run remote actions on the clients.
 
-<div class="el5 el6 el7 f29">
+<div class="el6 el7 el8">
```
evgeni, line 63: @jturel is `katello-host-tools` even required on el8 for report upload, or is sub-man alone enough these days?

ekohl (approved):

plugins/katello/nightly/installation/clients.md
jturel: Subman covers it all. However host-tools will get installed by katello-host-tools-tracer if folks are using that (not relevant here, just sharing)

jturel (approved):

PR #1721, opened 2020-11-04
switch to GH actions

evgeni:


evgeni:
Example run: https://github.com/evgeni/theforeman.org/runs/1351821471?check_suite_focus=true

tbrisker (approved):
Thanks @evgeni !

PR #1725, opened 2020-11-05
foreman 2.2: Fix service name in upgrade docs

bastelfreak:


_includes/manuals/2.2/3.6_upgrade.md
```diff
@@ -32,7 +32,7 @@ For more information about how to backup your instance head over to
 #### Step 2 - Perform the upgrade
 
 Before proceeding, it is necessary to shutdown the Foreman instance (e.g.
-`service httpd stop` or `service apache2 stop` usually).
+`service foreman stop` or `service foreman stop` usually).
```
bastelfreak: I'm not sure if it's still supported to run foreman through apache. If that's the case I can update the wording.

ekohl: It is. Note that we're also working on systemd socket activation in which case `foreman.socket` should also be stopped. Perhaps it should be changed to `systemctl stop httpd foreman.service foreman.socket`?

bastelfreak: updated it. It now assumes systemctl is available. I'm not sure if foreman still supports distributions so old that they don't have systemd, but I would also assume listing `service` isn't required for those rare cases.

ekohl: No, we're moving away from those.

_includes/manuals/2.2/3.6_upgrade.md
```diff
@@ -32,7 +32,8 @@ For more information about how to backup your instance head over to
 #### Step 2 - Perform the upgrade
 
 Before proceeding, it is necessary to shutdown the Foreman instance (e.g.
-`service httpd stop` or `service apache2 stop` usually).
+`systemctl stop httpd foreman.service foreman.socket` or
+`systemctl apache foreman.service foreman.socket` usually).
```
tbrisker: missed the stop here
```suggestion
`systemctl stop apache foreman.service foreman.socket` usually).
```

tbrisker (changes requested):
Thanks @bastelfreak, could you also apply the same change to the 2.3 and nightly manuals?

bastelfreak:
@tbrisker I hope I hit all occurrences.

tbrisker (approved):
Thanks @bastelfreak !

PR #1726, opened 2020-11-06
Katello 3.18 RC1 documentation

jturel:


chris1984 (approved):
ACK, all images are there in the user guide, 0 broken links and can see RC1 next to the version. API doc is generated as well.

PR #1727, opened 2020-11-09
Bump 3.17 version to GA

chris1984:


chris1984:
@jturel for when 3.17.0 gem gets pushed

jturel:
also need to set 3.17 as latest version : https://github.com/theforeman/theforeman.org/blob/256f7da0f1a1ee04312e4c44b7083e60d0bb9e70/_config.yml#L114

chris1984:
@jturel nice catch, updated.

jturel (approved):

chris1984:
@jturel finished testing 3.17.0 and everything looks good. Can this get a merge and I will make the announcement.

PR #1742, opened 2020-12-07
add el8 to osmenu.js list

evgeni:
otherwise it will never show the el8-specific entries

### DIFF
--- a/static/js/osmenu.js
+++ b/static/js/osmenu.js
@@ -3,9 +3,11 @@ $(document).ready(function(){
         'rhel5': [],
         'rhel6': ['el6'],
         'rhel7': ['el7'],
+        'rhel8': ['el8'],
         'el5': [],
         'el6': [],
         'el7': [],
+        'el8': []
         'f22': [],
         'f23': [],
         'f25': [],


static/js/osmenu.js
```diff
@@ -6,6 +6,7 @@ $(document).ready(function(){
         'el5': [],
```
ekohl, line 7: The line above this, should it also add `rhel8`? Haven't looked at the actual code.

evgeni: no, nothing uses "rhel8" via osmenu.js

the only place that uses "rhel8" is quickstart_installation.md -- and that has own JS.

I'll add it nevertheless, so that we can use it later

pcreech (approved):
ack

static/js/osmenu.js
```diff
@@ -3,9 +3,11 @@ $(document).ready(function(){
         'rhel5': [],
         'rhel6': ['el6'],
         'rhel7': ['el7'],
+        'rhel8': ['el8'],
         'el5': [],
         'el6': [],
         'el7': [],
+        'el8': []
```
evgeni, line 10: damn

PR #1743, opened 2020-12-08
Update method for retrieving 'oauth_consumer_key' and 'oauth_consumer…

illumino:
…_secret'

https://projects.theforeman.org/issues/31472
https://community.theforeman.org/t/web-site-manual-2-3-0-out-of-date-instructions-on-obtaining-oauth-consumer-key-and-oauth-consumer-secret/21586

### DIFF
--- a/_includes/manuals/nightly/3.2.3_installation_scenarios.md
+++ b/_includes/manuals/nightly/3.2.3_installation_scenarios.md
@@ -70,7 +70,7 @@ foreman-installer \
   --foreman-proxy-oauth-consumer-secret=<secret here>
 {% endhighlight %}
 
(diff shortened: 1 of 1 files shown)


ekohl (changes requested):
I would still prefer to get them via `foreman-rake config | grep oauth` or `foreman-rake -- config -k oauth_consumer_key`. Those are official interfaces that I would support. They use the application logic. When you don't use the installer, you can leave them out of `settings.yaml` and only set them in the database. Following the application logic prepares us better for a future where we might want to support containers.

illumino:
@ekohl did the update not meet your requirements?
Or did the changes just get missed?

ekohl:
>  Or did the changes just get missed?

This. My apologies for keeping this open so long.

ekohl (approved):

PR #1755, opened 2021-01-06
Make rpmconf remark more clear

lzap:
SSIA

https://community.theforeman.org/t/running-rpmconf-during-foreman-upgrades/21776

### DIFF
--- a/_includes/manuals/nightly/3.6_upgrade.md
+++ b/_includes/manuals/nightly/3.6_upgrade.md
@@ -67,10 +67,12 @@ Next upgrade all Foreman packages:
     yum upgrade tfm\* ruby\* foreman\*
 
 
-In order to catch all configuration changes and manage them properly you should install and run
-rpmconf from the EPEL repository along with vim-enhanced (for vimdiff):
+Optionally, configuration files should be reviewed. A good tool to find
+`.rpmnew` and `.rpmsave` files is `rpmconf`, keep in mind that some
+configuration files are managed by puppet therefore all changes made by the
+tool will be overwritten with the next execution of the `foreman-installer`:
 
-    rpmconf -a --frontend=vimdiff
+    rpmconf -a
 
 You can skip to Step 3.
 


_includes/manuals/nightly/3.6_upgrade.md
```diff
@@ -67,10 +67,13 @@ Next upgrade all Foreman packages:
     yum upgrade tfm\* ruby\* foreman\*
 
 
-In order to catch all configuration changes and manage them properly you should install and run
-rpmconf from the EPEL repository along with vim-enhanced (for vimdiff):
+Optionally, configuration files which are not managed by foreman-installer
+should be reviewed. A good tool to find `.rpmnew` and `.rpmold` files is
```
melcorr: I have no idea what you're trying to say here. Using optionally and should in the same sentence confuses things a lot. The rest makes sense.

lzap: Let me try to explain if you can find better text: When you install an update, RPM will create either `.rpmnew` or `.rpmold` depending on config type on those files you have changed. If you search your `/etc` I am sure you will find many of these on your laptop :-) It's up to you to review these and merge those changes with your files, just like you do in git.

This can be a lenghty process, thus Mirek Suchy from Red Hat wrote this nice tool which will find those and show you differences so you can quickly merge or zap changes.

Now, this still applies for Foreman upgrade, but only for configuration files which are not managed by our installer. And this was unclear.

What you should do (and I am trying to say) is to use this tool to merge configuration changes for those files. But the hard thing is to tell which files are managed by our installer and which are not. Even I don't remember them all.

Maybe we can rewrite this to: Optionally, you can use `rpmconf` to merge configuration file changes from packages, however keep in mind that re-running foreman-installer will overwrite changes for all managed configuration files.

@ekohl ?

ekohl: Is it `.rpmold`? I thought it was `.rpmsave`.

However, the suggestion `configuration files which are not managed by foreman-installer` is misleading (as you also write). That's because the installer doesn't remove those files. That means you'll have `/etc/foreman/settings.yaml.rpmnew` even though `/etc/foreman/settings.yaml` is managed. Your suggestion will mean users start to look at it and merge it manually.

Perhaps at a technical level the installer should ensure those `.rpmnew` files are removed. Then the instructions would become (roughly):
* Run `yum update`
* Optionally run `foreman-installer` (or do it manually)
* Run `rpmconf -a`

lzap: Isn't `rpmconf` first then `foreman-installer` the safer workflow? User can make a mess with the tool, while installer will always end up with working state.

I rebased, hopefully it's shorter with less confusion now.

PR #1759, opened 2021-01-12
Announce setting timestamps deprecation

ezr-ondrej:


_includes/manuals/nightly/1.2_release_notes.md
```diff
@@ -8,6 +8,8 @@ This section will be updated prior to the next release.
 
 ### Deprecations
 
+* Timestamps are deprecated in settings API, that means you should stop realying on `created_at` and `updated_at` attributes as these will change behaviour in Foreman 2.25. `created_at` will be dropped and `updated_at` will be set only for settings that doesn't have default value.
```
tbrisker: I really hope we never have a Foreman 2.25 😆 
```suggestion
* Timestamps are deprecated in settings API, that means you should stop realying on `created_at` and `updated_at` attributes as these will change behaviour in Foreman 2.5. `created_at` will be dropped and `updated_at` will be set only for settings that doesn't have default value.
```

ezr-ondrej: Done :)

tbrisker (approved):
Thanks @ezr-ondrej !

PR #1764, opened 2021-01-16
corrected latest version from 2.2.1 to 2.2.2

mzima:
Latest 2.2 version on the theforeman.org front page seems to be wrong. Should be 2.2.2 not 2.2.1.


tbrisker (approved):
Thanks! looks like we missed this one on the last update.

PR #1775, opened 2021-02-05
Katello 4.0 RC documentation updates

jjeffers:


_config.yml
```diff
@@ -109,7 +109,7 @@ defaults:
       path: "plugins/katello"
     values:
       layout: "plugins/katello/documentation"
-      latest: "3.18"
+      latest: "4.0"
```
jturel: Don't bump latest until 4.0 is GA :)

jjeffers: @jturel should we still merge this PR, assuming we don't bump latest as above? Or should we wait until GA to merge?

jturel: We can merge this as soon as RC1 hits the repos. Recommend sharing the draft release notes in IRC so others can have a look (kind of a tall order for one person to generate them imo)

jjeffers: Good idea!

jjeffers:
Closing in favor of new docs site.

PR #1778, opened 2021-02-16
Foreman 2.4 apidocs

upadhyeammit:
This pr does not include examples for all api as apipie:cache task is throwing many errors while generating docs. https://projects.theforeman.org/issues/31889


upadhyeammit:
This is based on RC1 packages, I will repush this once we have the RC2 packages available.

ekohl (approved):

PR #1781, opened 2021-02-18
Fix inline code for gem uninstall

maximiliankolb:


ekohl (commented):
Why is code better than pre here?

maximiliankolb:
Pre does not render correctly:
pre
I did not test code, but this seems to work on other occurrences in the same file.

"code" is for -well- monospaced code etc, "pre" is preformatted text, i.e. newlines are shown as newlines.

ekohl (approved):
Thanks! Adding that to the initial PR does help reviewers so please keep that in mind in the future.

PR #1782, opened 2021-02-24
Release notes for Foreman 2.3.3

upadhyeammit:


ekohl (changes requested):
I updated https://projects.theforeman.org/issues/31430 to set Fixed In now.

ekohl (approved):

PR #1792, opened 2021-03-17
CVE-2021-20259: Proxmox compute resource password leak

m-bucher:
See https://github.com/theforeman/foreman_fog_proxmox/pull/184


tbrisker (approved):
Thanks @m-bucher !

PR #1806, opened 2021-04-07
Update nightly instructions to include mention of rh-ruby27 and enabl…

ehelms:
…ing the Ruby 2.7 module on EL8

### DIFF
--- a/_includes/manuals/nightly/3.6_upgrade.md
+++ b/_includes/manuals/nightly/3.6_upgrade.md
@@ -103,7 +103,7 @@ yum upgrade tfm\* ruby\* foreman\*
 Optionally, consider removing unused SCL packages:
 
 {% highlight bash %}
-yum erase rh-ruby22\* rh-ruby24\* rh-ror42\* tfm-ror52\*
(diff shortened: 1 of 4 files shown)


_includes/manuals/nightly/3.6_upgrade.md
```diff
@@ -121,6 +121,13 @@ Clean up the yum metadata cache:
 dnf clean metadata
 {% endhighlight %}
 
+Enable the Ruby 2.7 module:
```
ekohl, line 124: On line 106 there's a yum erase line. Perhaps that should be updated to remove `rh-ruby25\*` as well?

ekohl (approved):

_includes/manuals/nightly/3.6_upgrade.md
ehelms: Updated.

ekohl (approved):
Just writing my thoughts down: this is not something we can solve in the installer since the installer already depends on it being done. It's conceptually similar to setting up yum repositories and this really is a bootstrap problem.

PR #1807, opened 2021-04-08
Drop unused i18n in apidocs

lzap:
I've realized that we carry 672 MB of i18n API docs content while we actually only link `v2.html`. All other languages can be dropped, we do not link these at all. This brings the api directory down to 52 MB.

I am going to move the content to a different git repository anyway, but I think it is worth making this repository slimmer right away, what you think?

Do not click on Files changed it will not load here, checkout the branch. But basically, this is what I did:

```shell
for L in ca cs_CZ de en en_GB es fr gl it ja ko nl_NL pl pt_BR ru sv_SE zh_CN zh_TW
do
  find . -name "*.$L.html" -exec rm -f {} \;
done
```


ekohl:
I think someone saw the Katello patch and wanted to beat the number of lines removed.

You should know that since Foreman 2.1 we only generate EN. Compare https://theforeman.org/api/2.1/ to https://theforeman.org/api/2.0/

> ```shell
>   find . -name "*.$L.html" -exec rm -f {} \;
> ```

You can use `-exec rm -f {} +` to have xargs style invocation. Can be faster :)

lzap:
> You should know that since Foreman 2.1 we only generate EN. Compare https://theforeman.org/api/2.1/ to https://theforeman.org/api/2.0/

I noticed, but the thing is - even `v2.en.html` is not used, we actually only use `v2.html` which then links all the versions without any language in the filenames.

Also many old versions ship not just en, but actually all the languages. These were never linked from our site, these should be safe to delete I guess.

lzap:
> You can use `-exec rm -f {} +` to have xargs style invocation. Can be faster :)

I actually made a script in the new repo which should keep things clean from now on, fixed.

ekohl:
> Also many old versions ship not just en, but actually all the languages. These were never linked from our site, these should be safe to delete I guess.

No, I do see them [here](https://theforeman.org/api/2.0/):

image

That's where they're linked in 2.0 and older.

lzap:
Ah good catch. I am gonna remove those links then, no reason to carry 700MB of garbage here I think.

ekohl:
Before making changes, let's decide on the path forward. If we decide to rip everything out anyway, then there's no point in cleaning it up. That only makes the repository bigger.

lzap:
> Before making changes, let's decide on the path forward. If we decide to rip everything out anyway, then there's no point in cleaning it up. That only makes the repository bigger.

Well if we decide to move, I will clean and drop the non-english languages anyway. There is very little advantage, the whole API documentation is actually language-less, it never made any sense to me. So I thought I would run the scripts which I will work on here as well.

This looks like a GNU find + sed - a single command cleanup.

lzap:
Let me rebase it, this is the script I used:

```shell

# Remove the i18n files
for L in ca cs_CZ de en en_GB es fr gl it ja ko nl_NL pl pt_BR ru sv_SE zh_CN zh_TW
do
  find . -name "*.$L.html" -exec rm -f {} +
done

# Remove the links at the top of each file to i18n files
find . -name "*.html" -exec sed -i 's/&nbsp;\[.*\]//' {} \;

```

lzap:
It works fine on my system. No links are shown. I actually used linkchecker to verify:

```
$ linkchecker index.html
INFO linkcheck.cmdline 2021-04-08 11:31:51,580 MainThread Checking intern URLs only; use --check-extern to check extern URLs.
LinkChecker 9.4.0              Copyright (C) 2000-2014 Bastian Kleineidam
LinkChecker comes with ABSOLUTELY NO WARRANTY!
This is free software, and you are welcome to redistribute it
under certain conditions. Look at the file `LICENSE' within this
distribution.
Get the newest version at https://linkchecker.github.io/linkchecker/
Write comments and bugs to https://github.com/linkchecker/linkchecker/issues

Start checking at 2021-04-08 11:31:51+002

Statistics:
Downloaded: 6KB.
Content types: 0 image, 1 text, 0 video, 0 audio, 0 application, 0 mail and 42 other.
URL lengths: min=8, max=74, avg=32.

That's it. 43 links in 43 URLs checked. 0 warnings found. 0 errors found.
Stopped checking at 2021-04-08 11:31:52+002 (1 seconds)
```

lzap:
For the record, the relevant discussion is here: https://community.theforeman.org/t/proposal-move-apidocs-from-git-to-infra-web/23059

We might actually skip this cleanup and simply remove all content from here as it will live in a different github repository: https://github.com/theforeman/apidocs

ekohl:
I'm :-1: on this. Let's not introduce more noise in the git log. Just remove it once we have the new apidoc up and running. And no, I can't post a review because this PR is too big.

lzap:
Yes, I agree. I just wanted to hear from you guys. Closing, lets drop the content completely: https://github.com/theforeman/theforeman.org/pull/1808

PR #1808, opened 2021-04-08
Remove apidocs completely

lzap:
This removes whole /api directory, also it updates the URL in the main menu and also drops unused api_v2 layout template.


lzap:
For the record, rewrite rules PR is here: https://github.com/theforeman/foreman-infra/pull/1600

lzap:
For the record, the discussion is at https://community.theforeman.org/t/proposal-move-apidocs-from-git-to-infra-web/23059

The infra (redirects) change is at: https://github.com/theforeman/foreman-infra/pull/1600

For easier review, the only two changes at the moment are:

* Menu URL change in `_config.yaml`
* Removal of the `layouts/api_v2.html`
* Removal of the whole `/api/` directories

You also want to review this github repo which hosts the `apidocs.theforeman.org` domain prior merging this: https://github.com/theforeman/apidocs

lzap:
For easier review I can remove deletion of "apidocs" directory, there is a change in layout. Anyone? @melcorr ?

melcorr:
I have no objections to this. Is everyone else OK with this?

tbrisker:
While i'd be happy to get rid of this, It doesn't feel like the new api website is quite ready.
A few issues I noticed:
- jquery isn't linked correctly, causing some interactivity to be broken (for example, click on "architectures" header and then on one of the actions, instead of expanding to show the action nothing happens)
- the TEMPLATE directory shouldn't be displayed on the index
- clicking on a version in the main index points to the `index.html` in that version which redirects to the `v2.html` index after a couple of seconds, instead of linking directly to `v2.html`.
- the infra redirects still haven't been merged

lzap:
> * jquery isn't linked correctly, causing some interactivity to be broken (for example, click on "architectures" header and then on one of the actions, instead of expanding to show the action nothing happens)

Fixed.

> * the TEMPLATE directory shouldn't be displayed on the index

Fixed in https://github.com/theforeman/apidocs/commit/c1041493ac2dc2a81aa6724afe8505b029257f7b

> * clicking on a version in the main index points to the `index.html` in that version which redirects to the `v2.html` index after a couple of seconds, instead of linking directly to `v2.html`.

This PR actually fixes this:

```
      - title: API docs
        url: https://apidocs.theforeman.org/foreman/latest/apidoc/v2.html
```

> * the infra redirects still haven't been merged

Let me check if there are some objections.

lzap:
> * clicking on a version in the main index

So the concern was that from the index page (generated by the tree tool) its redirect.

Explanation: I do not want to create another place we would need to manually maintain, our main site is the place that will link everything. The generated index is just for web spiders, and those who would manually enter the address. But if you insist I can create some landing site, but we would need to edit it on each release and also we will never link it from the main site.

I asked in the infra PR for green light.

lzap:
After a short IRC chat, I've updated the index.html to be more friendly in https://github.com/theforeman/apidocs/commit/cceb426967fd33ad99fb5ba2fd24381d9086a6dd

lzap:
Fixed also an extra LI HTML tag (bullet), if you can give this a review so we are all set and can wait for the infra PR merge. Thanks for the review!

ekohl:
When viewing a page I miss what version it's for. I'm leaning to keeping it simple and fall back to classic HTML frames. Thoughts?

lzap:
I am not sure what exactly are you missing, I do not see any version on the current site:

https://www.theforeman.org/api/2.4/index.html

Compare to:

https://apidocs.theforeman.org/foreman/2.4/apidoc/v2.html

The version is in the URL itself, if you expect anything on the apidocs pages, I think we should update apidocs itself.

ekohl:
You're right that the current location also has no indication which version you're browsing. Not a blocker to merge

lzap:
I believe I solved all the concerns, @tbrisker, I updated apidocs index page too. Can we merge this before 2.5 apidoc branching? This hasn't been done yet. FYI @upadhyeammit

I rebased but there were no changes, I wanted to make sure it applies cleanly and everything is up to date.

tbrisker:
GH won't let me approve because of the size of this change, but let's do this, if anything got missed we can follow-up on it. Great cleanup @lzap !

PR #1809, opened 2021-04-09
add el8 instructions to plugins

evgeni:
also fix smart proxy plugin rpm names

### DIFF
--- a/plugins/index.md
+++ b/plugins/index.md
@@ -239,13 +239,14 @@ To provide specific installation instructions, please select your operating syst
   <option value="none">-- select operating system --</option>
   <option value="debian">Debian or Ubuntu</option>
   <option value="el7">Enterprise Linux 7</option>
+  <option value="el8">Enterprise Linux 8</option>
 </select>
 
 <div class="installation_os installation_os_none">
   <i>No operating system selected.</i>
 </div>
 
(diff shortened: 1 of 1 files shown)


ekohl (commented):
Just an idea, but on EL we have provides, so you can do `yum install foreman-plugin-discovery` on both EL7 and EL8. We do do the same on Debian now so it can't fully unify it.

The only downside is that `yum search foreman-plugin` doesn't do anything.

evgeni:
I didn't want to start refactoring the whole page, and yes, the non-working search is rather annoying.

plugins/index.md
```diff
@@ -265,6 +266,16 @@ gpgcheck=0
 </pre>
 </div>
 
+<div class="installation_os installation_os_el8">
+<pre>
+[foreman-plugins]
+name=Foreman plugins
+baseurl=https://yum.theforeman.org/plugins/{{page.version}}/el8/x86_64/
+enabled=1
+gpgcheck=0
```
ehelms, line 275: Ought we to put in `module_hotfixes=1` like our release RPM does to be on the safe side for plugins to work?

evgeni: We can, yeah

ehelms (approved):

PR #1812, opened 2021-04-14
Move Release Notes to after Upgrade warnings

ekohl:
Releases notes can be long and cause users to miss the important bits. It is more important that users read Headline features, deprecations and upgrade warnings than the full release notes. This new order should make that easier.


lzap (approved):
Yeah! Thanks.

PR #1816, opened 2021-04-15
Fixing links to puppet ENC and report scripts

Learloj:
In https://github.com/theforeman/puppet-foreman/commit/0b3507b9286d966233fdf782ebb5d55eb90e4530 the ENC and report scripts were removed as those were split off to theforeman/puppetserver_foreman, causing all links in all docs to be dead.


ekohl (approved):
Thanks!

PR #1821, opened 2021-04-26
Add Foreman 2.5 release GPG key

zjhuntin:


lzap (approved):
I haven't tested it but I trust you. Thanks :-)

ekohl:
> I haven't tested it but I trust you. Thanks :-)

Then please don't merge it. I've always used this step to confirm that the GPG key is indeed correctly available and the procedure has been followed. In the past people have forgotten steps causing problems later on.

I have now verified it:
```console
$ VERSION=2.5 ./import_gpg_private 
✔ Copied theforeman/releases/foreman/2.5-gpg to clipboard. Will clear in 45 seconds.
gpg: keybox '/home/ekohl/dev/theforeman-rel-eng/releases/foreman/2.5/gnupg/pubring.kbx' created
gpg: /home/ekohl/dev/theforeman-rel-eng/releases/foreman/2.5/gnupg/trustdb.gpg: trustdb created
gpg: key 955C283E7522CC3F: public key "Foreman Automatic Signing Key (2.5) <packages@theforeman.org>" imported
gpg: key 955C283E7522CC3F: secret key imported
gpg: Total number processed: 1
gpg:               imported: 1
gpg:       secret keys read: 1
gpg:   secret keys imported: 1
```

lzap:
On my defence, I have used the internal Red Hat verify system to make sure that it is the real @zjhuntin

ekohl:
That's not really the point. The point is that in previous releases we've had the problem that the key wasn't shared internally. Then the next release manger in the rotation wasn't able to release packages. What you could also have done is to run `gpg2 --keyserver pgp.mit.edu --recv-keys HASH` to see if you got the same key. GPG keys must IMHO always be verified.

PR #1825, opened 2021-05-06
Release notes for Foreman 2.5.0 RC1

upadhyeammit:

PR #1826, opened 2021-05-06
Add a upgrade warning about new kind of facts from Discovery

domitea:
After merging https://github.com/theforeman/foreman_discovery/pull/524 it will be good to inform user about new kind of facts in Discovery. Thanks @lzap for mention that!

### DIFF
--- a/_includes/manuals/2.5/1.2_release_notes.md
+++ b/_includes/manuals/2.5/1.2_release_notes.md
@@ -26,6 +26,10 @@ Additionally, a new API endpoint `/api/registration_commands` makes it easy to g
 
 After defaulting to Puma in Foreman 2.1 and deprecation in Foreman 2.4, support for [mod_passenger is dropped](https://community.theforeman.org/t/drop-passenger-support/22787). The packages have been dropped and the installer can no longer configure it. Users who upgrade using the installer will be migrated automatically but users who upgrade without the installer must take care of this themselves.
 
+#### New kind of facts in Discovery plugin
+
+Discovered hosts have a new fact origin (discovery) so facts will no longer appear on the Monitor - Facts page. To search for discovery hosts, use the search box on the Hosts - Discovered page.
+
 ### Deprecations
 
 #### Puppet host methods in templates


lzap:
@upadhyeammit just added release notes here: https://github.com/theforeman/theforeman.org/pull/1825

Where you want this? @upadhyeammit

_includes/manuals/nightly/1.2_release_notes.md
```diff
@@ -6,6 +6,10 @@ This section will be updated prior to the next release.
 
 ### Upgrade warnings
 
+#### New kind of facts in Discovery plugin
```
ekohl: Is this in nightly (2.6/3.0 to be) or in 2.5? If it's 2.5 then you need to put it in the correct file.

_includes/manuals/nightly/1.2_release_notes.md
```diff
@@ -6,6 +6,10 @@ This section will be updated prior to the next release.
...
+
+Discovered hosts have a new kind of fact (discovery) so facts will no longer appear on the Monitor - Facts page. To search for discovery hosts, use the search box on the Hosts - Discovered page.
```
ekohl: Do you mean that it's a new (dedicated) fact origin?

ekohl:
> @upadhyeammit just added release notes here: #1825

That's a commit explicitly for the autogenerated releases notes from issues. A separate PR for the headline is good IMHO

_includes/manuals/nightly/1.2_release_notes.md
domitea: Yes, there a new kind of fact name (`DiscoveryFactName`) in Discovery.

_includes/manuals/nightly/1.2_release_notes.md
domitea: Good point! I will update the PR after release notes will be merged. This PR with new feature is already merged so that means it should be in 2.5 notes, right?

ekohl: Yes, if this is part of 2.5 then 2.5 the correct manual.

_includes/manuals/nightly/1.2_release_notes.md
ekohl: Then I'd use "fact origin" instead of "kind of facts". I think original also comes back in the UI and API so that's consistent. When I read "kind of facts" I thought it was an actual fact that provided more data.

_includes/manuals/nightly/1.2_release_notes.md
lzap: What Ewoud said, this needs to be done in the 2.5 branch and not nightly.

domitea:
Warning moved under version 2.5 but also remains in nightly because releasing of new version of Discovery plugin could be little different.

lzap:
Upgrade warning process has always been a nightmare, I am not sure if I am right, but you need to only change 2.5 branch. In nighty directory, we keep "a template".

ekohl:
Correct. Nightly will be 2.6 once we branch.

lzap:
@domitea please update only 2.5 thanks.

domitea:
@lzap PR updated.

lzap (approved):
Thanks, merging for 2.5.

PR #1827, opened 2021-05-11
Update installer options for Foreman 2.5 and nightly

upadhyeammit:


ekohl (approved):

ekohl:
Should we update the procedure to do this with nightly just prior to branching?

upadhyeammit:
> Should we update the procedure to do this with nightly just prior to branching?

Agree! Will send the change for same!

PR #1829, opened 2021-05-13
Add blog about Job Invocation Reports

melcorr:
@domitea and @ares - here it is. 
All feedback welcome!


domitea:
From my point of view, it's great article. Thanks @melcorr!

PR #1830, opened 2021-05-17
Remove `use_gravatar` setting documentation

alexjfisher:
The setting was removed in
https://github.com/theforeman/foreman/pull/5066 :(


ekohl (approved):
Thanks! In https://community.theforeman.org/t/do-you-have-setting-improvements-ideas/23547/13 I raised the idea of generating this reference documentation from code.

PR #1836, opened 2021-06-01
Fix "Populate users and attributes" section to use correct lookup_identity.conf

emersonford:
With the changes made to HTTP headers in [PR 7909](https://github.com/theforeman/foreman/pull/7909), a different lookup_identity.conf must be used as shown in puppet-foreman's [PR 872](https://github.com/theforeman/puppet-foreman/pull/872). 

As it stands, what is currently in the manual does not work (tested with Foreman 2.4). This new lookup_identity.conf does work as intended.


ekohl (approved):
Thanks!

PR #1838, opened 2021-06-04
Update latest version and latest news

upadhyeammit:


_includes/version.html
```diff
@@ -1,6 +1,8 @@
 <div class="version-marker">
   <p>
     Latest versions:
+      <a href="{{ site.baseurl }}manuals/2.5/index.html#Releasenotesfor2.4">2.5.0</a>
```
ekohl:
```suggestion
      <a href="{{ site.baseurl }}manuals/2.5/index.html#Releasenotesfor2.5">2.5.0</a>
```

ekohl (changes requested):
I think you have missed one instance.

_includes/version.html
```diff
@@ -1,6 +1,8 @@
...
+      /
       <a href="{{ site.baseurl }}manuals/2.4/index.html#Releasenotesfor2.4">2.4.0</a>
       /
       <a href="{{ site.baseurl }}manuals/2.3/index.html#Releasenotesfor2.3">2.3.4</a>
```
upadhyeammit, line 8: Not sure if I should remove 2.3.4 as we decided to do 2.3.5

ekohl: Since we're going to do a 2.3.5, I'd keep it in for now.

ekohl (approved):
Thanks!

PR #1853, opened 2021-06-30
Fix broken links

maximiliankolb:
This PR fixes a couple of broken links :)

### DIFF
--- a/privacy.md
+++ b/privacy.md
@@ -177,7 +177,7 @@ rights, subject to some limitations, against the Foreman Project:
 * The right to lodge a complaint with your local data protection authority
 
 If you would like to exercise any of these rights, you may do so by contacting
-the Foreman Project at [contact us](mailto:privacy AT community.theforeman.org)
+the Foreman Project at [contact us](mailto:privacy@community.theforeman.org)
 or on [IRC](/support.html#IRClivechat). Please understand, however, the rights
 enumerated above are not absolute in all cases, especially in regards to the
 right to erase your data:
@@ -193,7 +193,7 @@ right to erase your data:
 # Contacting us
 
 If you have any questions about this policy or our use of your data, please
-feel free to [contact us](privacy@community.theforeman.org) or on
+feel free to [contact us](mailto:privacy@community.theforeman.org) or on
 [IRC](/support.html#IRClivechat).
 
 # Changes to this Privacy Statement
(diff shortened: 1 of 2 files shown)


privacy.md
```diff
@@ -177,8 +177,8 @@ rights, subject to some limitations, against the Foreman Project:
 * The right to lodge a complaint with your local data protection authority
 
 If you would like to exercise any of these rights, you may do so by contacting
-the Foreman Project at [contact us](mailto:privacy AT community.theforeman.org)
```
ekohl, line 180: I think the `AT` part was mostly there to avoid spammers from harvesting it. Even though it's really easy to avoid, it's a common thing to do.

privacy.md
```diff
@@ -193,8 +193,8 @@ right to erase your data:
 # Contacting us
 
 If you have any questions about this policy or our use of your data, please
-feel free to [contact us](privacy@community.theforeman.org) or on
-[IRC](/support.html#IRClivechat).
+feel free to [contact us](mailto:privacy@community.theforeman.org) or on
+[IRC](/support.html#liberachat).
```
ekohl: Perhaps instead we should revert the header back to not break links?

ekohl (commented):
Thanks for looking at this!

privacy.md
maximiliankolb: Yes, that was probably the original idea. However, the email address was there in plaintext on the same page. So no need to "obfuscate" it in this line. See line 197.

privacy.md
maximiliankolb: There's only two more instances of `IRClivechat`:

```
_includes/manuals/2.4/1.2_release_notes.md:62:* Update all freenode references to point to https://theforeman.org/support.html#IRClivechat or libera.chat ([#32809](https://projects.theforeman.org/issues/32809))
_includes/manuals/2.5/1.2_release_notes.md:93:* Update all freenode references to point to https://theforeman.org/support.html#IRClivechat or libera.chat ([#32809](https://projects.theforeman.org/issues/32809))
```
I've changed those as well.

privacy.md
ekohl: Fair enough.

privacy.md
ekohl: What I meant (and still think) we should do is to align on `IRClivechat` and change the header on the support page to match. My reasoning is that we use IRC which happens to be libera.chat now but as we've seen, we can switch.

maximiliankolb: OK, I've reverted using `liberachat' . However, I am unsure if it works using `{#IRClivechat}` as an additional anchor, as it should be based on the level 2 heading in `support.md`.

maximiliankolb:
Rebased to `gh-pages`. Ready to be merged.

ekohl (approved):
Thanks!

PR #1857, opened 2021-07-07
mod auth gssapi documentation update

evgeni:


ekohl (approved):

tbrisker (commented):
I was hoping you would update https://github.com/theforeman/foreman-documentation/blob/master/guides/doc-Administering_Red_Hat_Satellite/topics/configuring-idm-authentication-on-satellite-server.adoc and modify the manual section to point to it, should have been clearer 😆

evgeni:
Well, those are different docs.

Here it's documented how to do manually, while the linked one uses the installer.

tbrisker:
I think the end goal is to have the new docs replace the existing manual, and ideally we shouldn't be documenting how to manually do things that the installer is doing?

evgeni:
That'd mean we'd stop supporting non-installer setups. That's something @ehelms was aiming at, but I don't think we ever formalized?

lzap:
The new documentation is more opinionated, one of these opinions is that we prefer our installer to do everything. Manual configs are there only for those setups which installer do not support. But I agree that generally we should be rather moving content from the old docs.

If you find useful, we can do "if foreman" in the manual and only show manual configuration for upstream version. I am not sure in this case tho.

evgeni:
> The new documentation is more opinionated, one of these opinions is that we prefer our installer to do everything. Manual configs are there only for those setups which installer do not support. But I agree that generally we should be rather moving content from the old docs.

I don't disagree. But today we do support non-installer scenarios (or we pretend to do so), and then ripping out perfectly good docs seems bad to me?

> If you find useful, we can do "if foreman" in the manual and only show manual configuration for upstream version. I am not sure in this case tho.

Nah, the existing "idm auth" in docs-ng is perfectly fine for all kinds of downstreams too.

And I'd probably not move "do it manually" steps there.

ekohl:
I still like to have some reference documentation somewhere. I agree it shouldn't be in the main manual, but I still question where to store it. In https://github.com/theforeman/puppet-pulpcore#readme I did try to document one system. We can do the same for others we support ([puppet-foreman](https://github.com/theforeman/puppet-foreman) for Foreman itself, [puppet-foreman_proxy](https://github.com/theforeman/puppet-foreman_proxy) for Foreman Proxy, etc) but then it's hard to gather all the combined information.

PR #1865, opened 2021-07-23
add upgrade note for host_param_true/false macros

evgeni:


evgeni:
FTR: I probably won't have time in the next weeks to revisit this, if there are any requested changes, so feel free to just edit as you like :)

ekohl (approved):
Thanks!

PR #1879, opened 2021-08-17
Add links to docs.theforeman.org where docs are better

melcorr:
I am working through https://community.theforeman.org/t/foreman-manual-reboot/22606
There are places where docs.theforeman.org links could really help a user today. 
If there are no objections, I would like to include links to docs.theforeman.org where the docs coverage is equal to or better than what we have in the Foreman manual. 

This first commit targets here: https://theforeman.org/manuals/2.3/index.html#4.1.6UserManagement

And tells the user about here: https://docs.theforeman.org/nightly/Administering_Red_Hat_Satellite/index-foreman-el.html#user-management_admin

I think this is a good example of when we can make an impact.

### DIFF
--- a/_includes/manuals/nightly/4.1.6_user_management.md
+++ b/_includes/manuals/nightly/4.1.6_user_management.md
@@ -1,7 +1,12 @@
 
 Foreman is all about hosts and users interacting with these hosts.
 
+You can find comprehensive user management instructions over on our
+[next generation docs site](https://docs.theforeman.org/nightly/Administering_Red_Hat_Satellite/index-foreman-el.html#user-management_admin).
+
 #### SSH Keys
+
+You can find more comprehensive documentation on SSH Keys in our [next generation docs site](https://docs.theforeman.org/nightly/Administering_Red_Hat_Satellite/index-foreman-el.html#ssh-keys_admin)
 
 Each Foreman user can have multiple SSH keys assigned when editing a user. These keys alone do not serve any purpose, but are available for use in provisioning templates and can be accessed via ENC data. They provide an easy way to manage users and login ssh keys on hosts without the need for LDAP.
 If you want users to be able to login to a host using the data provided in Foreman, you need to include the `create_users` snippet in your provisioning template. There is a [puppet module](https://github.com/ekohl/puppet-foreman_simple_user) available to keep user data in sync with Foreman and your hosts.


_includes/manuals/nightly/4.1.6_user_management.md
```diff
@@ -1,7 +1,12 @@
 
 Foreman is all about hosts and users interacting with these hosts.
 
+You can find comprehensive user management instructions over on our
+[next generation docs site](https://docs.theforeman.org/nightly/Administering_Red_Hat_Satellite/index-foreman-el.html#user-management_admin).
```
ekohl, line 5: We have a variable for the version so that on branching we don't link to nightly but actually to a stable version.

Now more in detail: I don't like this. It's so very wrong for many users. First of all, it's a Katello document. That also means it's a RPM-based document.

For starters, the whole first chapter (1.1) is wrong for non-Katello users.

Chapter 2 uses foreman-maintain which is not available on Debian and not installed by default on Foreman installations.

Chapter 3 only works on RHEL 7, not even CentOS (not 7 nor 8).

Chapter 5.4.5. why in gods name do we recommend users to use foreman-rake console? If anything, we should create a rake task in upstream and recommend something like `foreman-rake permissions:table`. Also, `yum install tfm-rubygem-foreman*` is **dangerous**. It installs **all** plugins which are available, which is a lot! You are extremely likely to break your installation with no way back.

Chapter 7 is RPM-only, which I guess is OK for the EL-guide but we should really link to both the RPM and Debian/Ubuntu guides here.

Chapter 8: again, RPM-only due to foreman-maintain usage. Mention of `/root/ssl-build` and various other directories makes it Katello-only. `/var/opt/rh/rh-postgresql12` makes it EL7 only.

Chapter 10: Also 100% Katello-only.

Chapter 11.4: Katello-only

Chapter 12:

> You can use the log files and other information described in this chapter to do your own troubleshooting, or you can capture these and many more files, as well as diagnostic and configuration information, to send to Red Hat Support if you need further assistance.

I don't think RH support is happy with this recommendation upstream :)

Chapter 12.4: contains various Katello-only paths

Chapter 13.3.3: I think step 9 is wrong and you should create a drop in. Using .include is deprecated and issues a warning on current systemd (included in EL8 at least). We should check that the installer doesn't clear it though.

Chapter 13.7: If you swap steps 3 and 4 step 5 becomes redundant. The installer (from current step 3) should handle registration for you and restarts if needed.

ekohl (changes requested):
Let's create some issues to get that document into a user consumable resource.

melcorr:
OK

_includes/manuals/nightly/4.1.6_user_management.md
lzap: Well I think this is a good review of the document itself rather than what Mel was trying to do. She proposed to link to chapter 5 which is correct, perhaps except the mentioned rake task.

We should not block this, unless its a link to a chapter that is completely irrelevant (e.g. chapter 10 - renaming an instance).

I suggest to reopen this PR and take the conversation about improving the admin guide elsewhere (I see there is another PR which I am going to comment in).

ekohl: There are 2 paragraphs. I'm a strong :-1: on this paragraph because the admin guide for non-Katello users is garbage for now.

PR #1886, opened 2021-08-26
Fix noVNC Link in 7.1_novnc.md

chrystt:
- http://kanaka.github.com/noVNC/noVNC/vnc.html throws 404
- replacing with https://novnc.com/info.html


ekohl (approved):
Thanks!

PR #1889, opened 2021-08-27
Installer options for Foreman 3.0.0-RC2

upadhyeammit:


ekohl (commented):
This all looks correct. Could you update nightly as well so that 3.0 isn't newer than nightly?

Thinking more about it, should we also update the installer options during stabilization week so that they are automatically copied when we branch?

upadhyeammit:
> This all looks correct. Could you update nightly as well so that 3.0 isn't newer than nightly?
> 
> Thinking more about it, should we also update the installer options during stabilization week so that they are automatically copied when we branch?

Agree, though we will still need to re-run if we do changes in installer in later stage. However major changes will be there in nightly and upcoming release: https://github.com/theforeman/tool_belt/pull/393

ekohl (approved):
Thanks!

PR #1890, opened 2021-08-27
Use CentOS 8 for installer options script

upadhyeammit:


upadhyeammit:
@ekohl on CentOS 7 foreman-installer was failing was dependency issues,
~~~
Error: Package: 1:foreman-installer-3.0.0-0.1.rc2.el7.noarch (foreman)
           Requires: puppet-agent >= 6.15.0
           Available: puppet-agent-5.0.0-1.el7.x86_64 (puppet5)
               puppet-agent = 5.0.0-1.el7

~~~

Instead of fixing this I thought to move to EL8. Request to review, thank you!

ekohl (approved):
:+1: for this. As a project we should start considering EL7 deprecated and moving things over to EL8.

PR #1891, opened 2021-08-31
fixed noVNC Link in 7.1_novnc.md in all manuals

chrystt:
https://github.com/theforeman/theforeman.org/pull/1886 fixed only in 1.4 manual
this fixes it now in all manuals


ekohl (approved):
Thanks!

PR #1897, opened 2021-09-13
Update demo listing with demo 100

melcorr:
@ianballou  - your demo wasn't labelled katello by Brad but looks to me like I should include it with the Katello demos, what do you think?


ianballou:
@melcorr I'd say we can mark it Katello, sure.

PR #1903, opened 2021-09-27
Fixes #20977 - Document that rubygem-rubyipmi might cause issues

nbarrientos:
Unfortunately, there isn't still any version of rubyipmi containing a fix for what's described in #20977 [0]. We've hit this issue in production and having this described in the documentation would have saved us some time debugging.

I think that given there hasn't been a release upstream since 2015 and that the bug is still present in the wild (for instance, the version of rubyipmi shipped in the Foreman repositories is affected) it's now
time to add something to the manual.

Not sure if I'm adding the new entry to the documentation to the correct place.

[0] https://projects.theforeman.org/issues/20977

### DIFF
--- a/_includes/manuals/nightly/4.3.3.2_smartproxy_bmc_ipmitool.md
+++ b/_includes/manuals/nightly/4.3.3.2_smartproxy_bmc_ipmitool.md
@@ -1,0 +1,14 @@
+
+If you select `ipmitool` as your default provider, there's an issue
+that might bite you. The Gem that's used to interact with `ipmitool`
+has a bug for which there's a fix but no stable release has been
+released for years now containing it.
+
+When Foreman calls `/bmc/X.X.X.X/lan/foo` (typically when a host with
+BMC interface is browsed), the proxy might enter an infinite loop if
+the BMC is not working, provoking a huge process creation/destruction
+rate, high system CPU use and high system load as consequence.
+
+See Foreman issue
+[#20977](https://projects.theforeman.org/issues/20977) for more
+details.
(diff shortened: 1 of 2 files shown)


ekohl (commented):
I wonder if we should instead work around it in packaging. It's always easier to not have to tell users about a known issue if there is no issue.

nbarrientos:
Yes, indeed, getting the Gem fixed would be perfect but given how long we've been waiting for a release perhaps this is the best option we have at the moment. Happy to wait for https://github.com/logicminds/rubyipmi/issues/50, though. Let's see :crossed_fingers:

ekohl:
There is now a release and https://github.com/theforeman/foreman-packaging/pull/7148 (RPM) and https://github.com/theforeman/foreman-packaging/pull/7149 (deb) should update our packaging.

nbarrientos:
That's great, that's much better indeed. What versions of Foreman the new package will make an appearance on? Only 3.0? This can be closed, btw.

ekohl:
Let's first get it into nightly, then see how well that works before cherry picking.

nbarrientos:
Thanks. For us it'd fine if it does make it down to `2.4`, just asking to be prepared to remove our local patching or not. Thanks for working on this.

ekohl:
Technically 2.4 is unsupported now that 3.0 is out. IIRC Foreman 2.4 was also Ruby 2.5 for RPMs so there you can't simply drop in the RPM. That means you probably need to keep your local patching for now.

ekohl:
Given we've merged fixed packages, shall we close this?

nbarrientos:
Aye, we'll keep the local patching for 2.4 :)

PR #1914, opened 2021-11-05
Add historic Foreman demos & prepare demos for 2022

ekohl:
This moves /demos-2021.html to /demos/2021.html. This creates a nice directory for all demos. It also adds an index with a listing. While nothing links to it yet, it prepares for it. There's also a page with historic demos as copied from the Redmine wiki (https://projects.theforeman.org/projects/foreman/wiki/Current_Sprint_Information).


lzap:
Other than that, it renders fine. @stejskalleos ?

stejskalleos:
LGTM, with this we could remove the old wiki page!

ekohl:
@melcorr since you're maintaining the demo pages, I'd like to get your thoughts on this too.

PR #1918, opened 2021-11-16
Use more markdown

ekohl:
This replaces various HTML tags and uses native markdown syntax. This also brings in more syntax highlighting.

In `3.5.5_facts_and_the_enc.md` the phrasing is changed a bit. Otherwise the numbering was off. It would have been `1. ... 1. ...` instead of `1. ... 2. ...`.


lzap (approved):
Nice, I have always wondered why we had these  pre element in the fist place.

ekohl:
Let's call them historical reasons ;)

PR #1921, opened 2021-11-19
Show Downloading the installer data to centos/el 7

MariaAga:
was shown as empty
Screenshot from 2021-11-20 00-09-23


evgeni:
Thanks @MariaAga 

Could you apply the same change also for the nightly, 3.0 and 2.5 (and any older, if you desire to, of course) manuals?

MariaAga:
Yup, hope I found them all

evgeni (approved):

PR #1933, opened 2021-12-09
Update 3.1.4_browser_compatibility.md

sharvit:
Since the foreman started using the patternfly-v4 library, it is not compatible with IE anymore.
Also, it requires the latest versions of the supported browsers:
https://www.patternfly.org/v4/get-started/about#supported-browsers

### DIFF
--- a/_includes/manuals/nightly/3.1.4_browser_compatibility.md
+++ b/_includes/manuals/nightly/3.1.4_browser_compatibility.md
@@ -3,9 +3,8 @@ Using the most recent version of a major browser is highly recommended, as Forem
 
 The recommended requirements are as follows for major browsers:
 
-* Google Chrome 54 or higher
-* Microsoft Edge
-* Microsoft Internet Explorer 10 or higher
-* Mozilla Firefox 49 or higher
+* Google Chrome latest
+* Microsoft Edge latest
+* Mozilla Firefox latest
 
 Other browsers may work unpredictably.


_includes/manuals/nightly/3.1.4_browser_compatibility.md
```diff
@@ -3,9 +3,8 @@ Using the most recent version of a major browser is highly recommended, as Forem
 
 The recommended requirements are as follows for major browsers:
 
-* Google Chrome 54 or higher
-* Microsoft Edge
-* Microsoft Internet Explorer 10 or higher
-* Mozilla Firefox 49 or higher
+* Google Chrome latest
+* Microsoft Edge latest
+* Mozilla Firefox latest
```
ekohl, line 8: I don't think this is acceptable. Browsers also have LTS versions and we should always require the latest version.

ekohl (changes requested):
How far back do we need to cherry pick this?

_includes/manuals/nightly/3.1.4_browser_compatibility.md
sharvit: I agree it shouldn't be acceptable but this is what pf4 gives us to work with.
The most we can do is to manually set up the current latest LTS release every time we release foreman. It also means that we will need to be careful with cherry-picks that upgrade the pf package.

ekohl: We should take this as feedback to the PF4 team. Just as a context: Firefox on RHEL is an LTS release. That means a team of Red Hat people don't support their own platform.

sharvit:
> How far back do we need to cherry pick this?

I'm not sure, we added the pf4 to the vendor at v4.2.0:
https://github.com/theforeman/foreman-js/blob/master/CHANGELOG.md#420-2020-03-15

foreman v2.0 asks for `^4.0.7` so it depends on packaging:
https://github.com/theforeman/foreman/blob/2.0-stable/package.json#L23

foreman v2.1 uses vendor v4.3.0 so it's must contain pf4:
https://github.com/theforeman/foreman/blob/2.1-stable/package.json#L23

lzap:
What is the status please?

ekohl:
Superseded by https://github.com/theforeman/theforeman.org/pull/1959

PR #1938, opened 2022-01-11
Add drop require_ssl_smart_proxies upgrade warning

ekohl:


ezr-ondrej (approved):
The PR was merged and it definitelly deserves a warning. The wording looks good to me :+1:

ekohl:
@melcorr could I ask you to take?

lzap:
I am hoping to migrate manual to the new site this month, I suggest to start collecting upgrade warnings and release notes on Discourse, we could turn it into a wiki and then move to the final destination during release process.

ekohl:
> I am hoping to migrate manual to the new site this month

I think that's very optimistic. Last time I also had similar thoughts and it ended up depressing me.

> I suggest to start collecting upgrade warnings and release notes on Discourse, we could turn it into a wiki and then move to the final destination during release process.

:-1: on that. We still have a single place that everything points to. Why do we need to find an intermediate location. If your migration doesn't happen, we'll have regressed. If it does, we still know exactly where to get it. Here we have reviews, on Discourse we don't.

lzap:
The release/upgrade notes process is a pure disaster, modifying nightly document, conflicts, branching problems, then we have to delete it during new release. I see nothing but burden, few sentences do not need deep reviews. We do not support upgrading to nightlies, why do you think it is worth any effort to keep nightly documentation up to date with upgrade/release notes?

Frankly, I intentionally avoid some upgrade/feature notes to avoid this terrible process. Unless someone insists or its very very important note, I skip it. If I only could edit a single document to add a sentence that would be different.

Sometimes, it is better to think what's better for devs, not users.

lzap (approved):

PR #1939, opened 2022-01-14
Rephrase unattended deprecation message

ekohl:
This adds a bit of context to the deprecation which makes it easier to understand. It also removes the duplicate entry for Leos in a (Last + First) notation since there's already a (First + Last) notation.


lzap (approved):
TY

PR #1942, opened 2022-01-21
Add 2022 demo page to website

melcorr:
* Add Demo 105
* Add link to 2022 to index page
* Add link to index to 2022 page

I previewed the build and all links working etc.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -64,7 +64,7 @@ navigation:
       - title: Events
         url: https://theforeman.org/events
       - title: Community Demos
-        url: /demos/2021.html
+        url: /demos/2022.html
       - title: Videos
         url: /media.html
       - title: Security
--- a/demos/2022.md
+++ b/demos/2022.md
@@ -1,0 +1,45 @@
+---
+layout: demos
+title: Community Demos
+---
+
+# Community Demos (2022)
+------------------------------
+
+Every few weeks we host a Community Demo to showcase new & interesting developments from the Foreman
+community.
+We encourage participation from any member of the community (although you do need a
+Google account), so if you’ve been working on something cool, please do come show it off.
+If you'd like to watch the community demo, they're usually on a Thursday and we always list them in
+advance on our [Community Events Calendar](https://community.theforeman.org/c/events/l/calendar).
+
+Here is a list of every demo that has taken place this year.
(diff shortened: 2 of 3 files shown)


_config.yml
```diff
@@ -64,7 +64,7 @@ navigation:
       - title: Events
         url: https://theforeman.org/events
       - title: Community Demos
-        url: /demos/2021.html
+        url: /demos/2022.html
```
ekohl, line 67: I wonder if this should start linking to the demos overview
```suggestion
        url: /demos/
```
That way users can also look back at the 2021 and earlier demos.

demos/2022.md
```diff
@@ -0,0 +1,45 @@
...
+advance on our [Community Events Calendar](https://community.theforeman.org/c/events/l/calendar).
+
+Here is a list of every demo that has taken place this year.
+Our demos are split into two sections: a user-focused section and a developer-focused section.
+
+If you've any questions or feedback, we welcome any feedback in our
+[community](https://community.theforeman.org)
+
+You can find older demos here:
+* [Foreman Community Demos](/demos/index.html)
```
ekohl, line 23: You don't need to include `index.html`
```suggestion
* [Foreman Community Demos](/demos/)
```

demos/2022.md
```diff
@@ -0,0 +1,45 @@
+---
+layout: demos
+title: Community Demos
```
ekohl, line 3: Should include the year to differentiate pages? If so, we should do the same with the 2021 page.
```suggestion
title: Community Demos 2022
```

ekohl (commented):
Overall :+1: with some possible improvements inline.

_config.yml
melcorr: In reality, why would anyone want to watch old demos? 
On YouTube, there is a stagnation in views after about a month.
QE etc are interested for a max of about six months. 
I don't want to fetishize this process. This suggestion adds an extra step for users.

ekohl: I did see (after I placed this comment) you added a link on the 2022 page itself. That is something we previously didn't have and may be sufficient to cover the use case.

ekohl:
Did you close this accidentally? Because I'm ok with merging it even without changes.

PR #1946, opened 2022-01-31
Updated command to setup foreman-proxy

aheath1992:
Per conversation with ewoud:
<ewoud> anheath: that's odd, somehow --no-enable-foreman-puppet isn't respected?
<ewoud> anheath: oh, it's --no-enable-foreman-plugin-puppet but it silently ignores the invalid option?
<anheath> ewoud, So the commanline aruments are incorrect?
<ewoud> anheath: line 3 (--no-enable-foreman-puppet) should be --no-enable-foreman-plugin-puppet


ekohl (changes requested):
Could you also update nightly? Other than that :+1:

aheath1992:
Done

ekohl (approved):
Thanks!

PR #1951, opened 2022-02-04
Use Stream 8 image as CentOS 8 is EOL

upadhyeammit:


upadhyeammit:
Considering we are going to build on RHEL now, I think we can use ubi here: https://www.redhat.com/en/blog/introducing-red-hat-universal-base-image .

@ekohl request to review

evgeni:
Why not the Stream 8 image?

upadhyeammit:
> Why not the Stream 8 image?

I thought ubi is close to building on RHEL, however Stream 8 works as well.

evgeni:
UBI is close to RHEL, yes, but I don't see any benefit for building the installer docs on RHEL

scripts/installer/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM centos:8
+FROM centos/centos:stream8
```
evgeni: Can you please use quay.io/centos/centos:stream8 so it always pulls from quay and not whatever is the default on the machine it's executed on?

upadhyeammit: done!

evgeni:
I'm cool with that change, leaving final review to Ewoud

ekohl (approved):
Thanks!

PR #1968, opened 2022-03-08
add my talks from fosdem 2021 and 2022

evgeni:


evgeni:
I do wonder if we want to re-structure the media page a bit:
* there is a community demos section, with *really* old stuff. should be probably merged into `/demos/` and removed here?
* the "case studies" is rather old and a single video, probably shouldn't be the first thing people see?
* deep dives and screencasts probably could use a bit of date-stamping, so people know it's not the freshest? (or should we create an "archive" page?)

ekohl (approved):
Restructuring sounds like a good idea. For now I'm going to merge this.

PR #1969, opened 2022-03-14
Update release notes - Host Reports

lzap:
SSIA

FYI the plugin is not in 3.2 yet we are finishing migration rake task today and then releasing a new version that is the final one.

### DIFF
--- a/_includes/manuals/3.2/1.2_release_notes.md
+++ b/_includes/manuals/3.2/1.2_release_notes.md
@@ -6,6 +6,12 @@
 
 It is now possible to run Foreman on Debian 11. Users are encouraged to upgrade.
 
+#### Host reports (preview)
+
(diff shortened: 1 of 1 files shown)


_includes/manuals/3.2/1.2_release_notes.md
```diff
@@ -2,12 +2,24 @@
 
 ### Headline features
 
+#### Host reports
```
ekohl: I don't think this should be a headline feature if there is no installer support. I also think we shouldn't introduce headline features in a bugfix release.

IMHO we should either postpone the release (and make sure everything gets in) if we want to make it a headline feature or we release it in 3.2.1 as an optional new feature. I'm leaning to the latter.

lzap: Ok how about Preview feature? I will link the known bug as well.

lzap: I moved it down and marked them as a preview feature.

ares: I think the "preview" makes sense if we provide instructions how to install. We want to let users know they can start playing with that. Also that prepares the ground for the release with 3.3.

ekohl:
By now the reports plugin has been retired so I'm closing this.

PR #1970, opened 2022-03-14
Updating docs to inform users not to enable EPEL8 repo

aheath1992:
Updating docs to inform users not to enable EPEL8 repo as it will cause dependency issues with foreman


ekohl (approved):
Thanks!

PR #1975, opened 2022-04-06
Fix link to Upgrade section

dreua:
This answer helped me to apply the change to all affected versions: [filenames - How to apply a Git patch to a file with a different name and path? - Stack Overflow](https://stackoverflow.com/questions/16526321/how-to-apply-a-git-patch-to-a-file-with-a-different-name-and-path/29888775#29888775)


ekohl (approved):
Thanks, both for the fix and the easy git way. I always used `cd $directory && patch -p1 < $patch`.

PR #1977, opened 2022-04-28
add upgrade el7 to el8 post

evgeni:


_posts/2022-04-28-upgrading-foremankatello-from-el7-to-el8.md
```diff
@@ -0,0 +1,140 @@
...
+   ```console
+   # yum remove epel-release
+   ```
+5. Remove `centos-release-scl` and `centos-release-scl-rh` repositories, as we don't need them anymore on EL8:
+   ```console
+   # yum remove centos-release-scl centos-release-scl-rh
+   ```
+6. Let LEAPP analyze your system:
+   ```console
+   # leapp preupgrade
```
ehelms, line 116: In the even this finds harder to fix issues, should it happen before users go and remove a bunch of repositories off their systems?

ehelms:
Should we note in here that it will take a CentOS 7 installed machine to CentOS 8 Stream ?

_posts/2022-04-28-upgrading-foremankatello-from-el7-to-el8.md
evgeni: what kind of issues do you envision (that would be solved by having repos)?

ehelms: I have no idea what issues all this step finds, but if its something big enough that I want to back off and keep my system as is while I investigate the changes it requires then it would be nice to have the repos? I'd be happiest if you said there is nothing that big it finds, I just know from a few times trying upgrades and scratching my head at the output.

evgeni:
https://github.com/theforeman/foreman-documentation/pull/1248 has been merged and superseedes this PR

PR #2005, opened 2022-08-02
Update the dockerfile to generate installer options

upadhyeammit:


upadhyeammit:
The missing piece was `dnf -y module enable foreman:el8` however as I am at it updated commands to dnf too

ekohl (approved):
Thanks!

PR #2007, opened 2022-08-02
Update some documentation.

aheath1992:
Updated some documentation from some of my testing and conversations with memebers


ekohl:
@adamruzicka could you have a look? My biggest issue is that this documents version 1.7 but we're at a much higher version now.

adamruzicka:
The rename in installer is not tied to any particular REX version, IIRC it was renamed with Foreman 3.2. The change here is correct, however shouldn't the new docs be considered the source of truth?

ekohl:
We're in a weird place where https://docs.theforeman.org is production quality documentation for Katello, but it's incomplete for vanilla Foreman. This is because so many places assumed Katello was present that many guides were simply unusable.

aheath1992:
@ekohl @adamruzicka I am flexible, if this request needs to be updated moved to a different location, or not needed at all please let me know.

PR #2014, opened 2022-09-13
Fix stop/start commands of upgrade manual

timdeluxe:
This bugged me a long time already. I don't know in which world "apache" might work without the "2", but in ours it doesn't. I hope i raised the PR for the right document.


ekohl (approved):
Thanks!

PR #2018, opened 2022-09-29
Use syntax highlighting where available

ekohl:


ekohl:
I forgot about this PR and now I'm sad I didn't do this before branching. @evgeni mind taking a look?

PR #2025, opened 2022-10-12
Add a warning that Katello docs have moved to docs.theforeman.org

ianballou:
theforeman.org is still the top search result on Google, so I figured we should make it more obvious where the official Katello docs live.

### DIFF
--- a/plugins/katello/index.md
+++ b/plugins/katello/index.md
@@ -11,6 +11,12 @@ title: Katello Documentation
     </h1>
 
     <br/>
+	<b style="font-size: 26px">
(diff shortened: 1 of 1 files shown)


ianballou:
image

chris1984 (approved):
ACK looks good. @ekohl can this get a review/merge when you get a free moment please?

evgeni (approved):

ekohl:
I think we should also reduce the versions on the right side. 4.0 and nightly should be dropped. Perhaps even remove it altogether and have an inline link to "older releases" (which should include 3.18 and 3.17).

ianballou:
Yeah good point, I missed the nightly bit. I can fix that up in another PR.

PR #2029, opened 2022-11-03
drop chef from the nightly manual

evgeni:


manuals/nightly/index.md
```diff
@@ -192,27 +192,24 @@ previous_version: "3.0"
...
-## 4.10 Monitoring
-{%include manuals/{{ page.version }}/4.10_monitoring.md %}
-### 4.10.1 Global status
-{%include manuals/{{ page.version }}/4.10.1_global_status.md %}
-### 4.10.2 Sub-statuses
-{%include manuals/{{ page.version }}/4.10.2_substatuses.md %}
-### 4.10.3 Searching by statuses
-{%include manuals/{{ page.version }}/4.10.3_searching_by_statuses.md %}
-
-## 4.11 Reports
```
ekohl: You'll hate me for this, but this breaks anchors in Foreman:
https://github.com/theforeman/foreman/blob/d559505b68f5461593ecb945f602a7925f260179/app/controllers/report_templates_controller.rb#L10
https://github.com/theforeman/foreman/blob/d559505b68f5461593ecb945f602a7925f260179/app/views/report_templates/welcome.html.erb#L9
https://github.com/theforeman/foreman/blob/d559505b68f5461593ecb945f602a7925f260179/app/views/report_templates/index.html.erb#L4

evgeni: I have no idea how to answer that :smiling_imp:

ekohl: It means we have to patch Foreman if we reorder this (change 411 to 410).

evgeni: Or we just replace the section with "the chef plugin was removed"? Then there is no reordering

ekohl: I'd be OK with that in the short term. Long term we'd ideally migrate to the new docs.theforeman.org anyway.

evgeni: done now

ekohl (approved):

PR #2030, opened 2022-11-03
add upgrade warning about chef and abrt

evgeni:


ekohl (approved):

ekohl:
Should this reflect that abrt was already gone from packaging?

evgeni:
> Should this reflect that abrt was already gone from packaging?

I wouldn't know how to phrase it without making it overly complicated and less understandable.

If you have a suggestion, sure.

ekohl:
I think we can drop abrt since it was already uninstallable. The installer didn't expose it and there was no package. I even documented its removal in `_includes/manuals/1.22/1.2_release_notes.md` already (and made https://projects.theforeman.org/issues/26665). Keeping it in puppet-foreman(_proxy) was an oversight.

evgeni:
Fair.

ekohl (approved):

PR #2031, opened 2022-11-04
Ansible.cfg release note

adamruzicka:
The change that prompted this https://github.com/theforeman/puppet-foreman_proxy/pull/777

### DIFF
--- a/_includes/manuals/nightly/1.2_release_notes.md
+++ b/_includes/manuals/nightly/1.2_release_notes.md
@@ -27,6 +27,12 @@ The installer now ensures Redis 6 is used on Enterprise Linux 8.
 The Redis 5 AppStream [went EOL in May 2022](https://access.redhat.com/support/policy/updates/rhel-app-streams-life-cycle).
 Previously it was up to the user to switch streams, but now the installer enforces the `redis:6` stream and updates the package.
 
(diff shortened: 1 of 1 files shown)


_includes/manuals/nightly/1.2_release_notes.md
```diff
@@ -27,6 +27,12 @@ The installer now ensures Redis 6 is used on Enterprise Linux 8.
 The Redis 5 AppStream [went EOL in May 2022](https://access.redhat.com/support/policy/updates/rhel-app-streams-life-cycle).
 Previously it was up to the user to switch streams, but now the installer enforces the `redis:6` stream and updates the package.
 
+#### Allow manual modifications of ansible.cfg
+
+The installer no longer manages ansible.cfg and thus allows users to put in their own modifications.
+This file was previously used to configure the options that Foreman required to have set.
+This is no longer the case as the necessary options are set using a different mechanism.
```
ekohl, line 34: Should we add a note that the old file remains?
```suggestion
This is no longer the case as the necessary options are set using a different mechanism.
During upgrades the file is not touched which means some cruft remains.
It is safe to remove the file or remove all content.
```

ekohl (approved):

PR #2035, opened 2022-11-08
Fix `Foreman installer` puppet version reference

alexjfisher:
At least some of the installer modules have required Puppet 6 since Foreman 3.0


ekohl (approved):
Thanks!

PR #2038, opened 2022-11-09
Update the repository list for committers

ekohl:


evgeni (approved):
There are more repos that could be added, but the existing two commits are absolutely good as they are

PR #2045, opened 2022-12-01
Update statements around EL support

ekohl:
Scientific Linux never released version 8 so it can be dropped starting Foreman 3.4. It also generalizes the descriptions and mentions AlmaLinux which is used in our pipelines. CentOS Linux 8 is EOL and can be removed.

If this is approved I'll also apply the changes to 3.4 & 3.5.

### DIFF
--- a/_includes/manuals/nightly/2.1_quickstart_installation.md
+++ b/_includes/manuals/nightly/2.1_quickstart_installation.md
@@ -24,10 +24,8 @@ function update_quickstart_os(select) {
 To provide specific installation instructions, please select your operating system:
 <select onChange="update_quickstart_os(this);">
   <option value="none">-- select operating system --</option>
-  <option value="el8">CentOS 8</option>
-  <option value="el8">CentOS 8 Stream</option>
+  <option value="el8">Enterprise Linux 8</option>
   <option value="debian11">Debian 11 (Bullseye)</option>
-  <option value="rhel8">Red Hat Enterprise Linux 8</option>
   <option value="ubuntu2004">Ubuntu 20.04 (Focal)</option>
 </select>
 
@@ -37,7 +35,7 @@ To provide specific installation instructions, please select your operating syst
   <i>No operating system selected.</i>
 </div>
 
-<div class="quickstart_os quickstart_os_rhel8 quickstart_os_el8">
+<div class="quickstart_os quickstart_os_el8">
   <p>
     Enable Puppet's 7.x repository:
   </p>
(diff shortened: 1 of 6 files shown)


_includes/manuals/nightly/2.1_quickstart_installation.md
```diff
@@ -24,7 +24,6 @@ function update_quickstart_os(select) {
 To provide specific installation instructions, please select your operating system:
 <select onChange="update_quickstart_os(this);">
   <option value="none">-- select operating system --</option>
-  <option value="el8">CentOS 8</option>
   <option value="el8">CentOS 8 Stream</option>
   <option value="debian11">Debian 11 (Bullseye)</option>
   <option value="rhel8">Red Hat Enterprise Linux 8</option>
```
evgeni: This list doesn't mention AlmaLinux or any other EL-flavors.

_includes/manuals/nightly/2_quickstart_guide.md
```diff
@@ -3,19 +3,15 @@ The Foreman installer is a collection of Puppet modules that installs everything
...
 
+The installation will require 4GB of memory, see [System Requirements](manuals/{{page.version}}/index.html#3.1SystemRequirements) for more information.
+
 #### Supported platforms
-* CentOS 8 x86_64
-* CentOS 8 Stream x86_64
+* AlmaLinux 8, x86_64
+* CentOS 8 Stream, x86_64
 * Debian 11 (Bullseye), amd64
 * Red Hat Enterprise Linux 8, x86_64
```
evgeni: Technically, this is also untested ;)

_includes/manuals/nightly/3.1.1_supported_platforms.md
```diff
@@ -25,7 +23,7 @@ Other operating systems will need to use alternative installation methods, such
 
 The following operating systems are known to install successfully from Foreman:
 
-* RHEL and derivatives (CentOS, Scientific Linux, Oracle Linux) 3+
+* RHEL and derivatives (AlmaLinux, CentOS, Oracle Linux) 3+
```
evgeni:
```suggestion
* RHEL and derivatives (AlmaLinux, CentOS, CentOS Stream, Oracle Linux, Rocky Linux) 3+
```
(this is about clients, and I think we do support all of these)

evgeni (commented):
I wonder if for the target deployment platforms, we should switch to the "Enterprise Linux" lingo we have been using in `foreman-documentation` and explain what we test in a "separate" section (like we already do for the "but it also works on other RHEL rebuilds" -- just needs a bit of rewording I guess?)

evgeni (approved):

ekohl:
Now also applied to the 3.5 branch using `cd _includes/manuals/3.5 && git show | patch -p4`

PR #2026, opened 2022-10-14
Cleanup Katello versions, convert remaining legacy ones to PDF, update some wording

ianballou:
I tore out the Katello 3.17 and 3.18 docs and turned them into PDFs.  I also updated a little of the Katello landing page wording to match modern Katello.  Links to the nightly docs have been removed, along with the out of date nightly docs that are currently there.

I also need two PDF files (the Katello 3.17 and 3.18 docs) uploaded to downloads.theforeman.org.  I can't post them here because they're too big, so let me know how I can get those files uploaded.

PR #2053, opened 2023-01-12
Reflect new Candlepin changes on Katello docs

adamlazik1:
The target Katello version is 4.7, I am not sure when is the right time to merge this.
CC @chris1984


ekohl:
I think we're moving away from the Katello docs in this repo. https://github.com/theforeman/theforeman.org/pull/2026 still needs some work, but foreman-documentation should be the primary source for Katello docs.

adamlazik1:
Should I just close this, then?

ekohl:
I think so yes.

PR #2075, opened 2023-03-28
Foreman 3.6.1 release notes

ekohl:


_includes/manuals/3.6/1.2_release_notes.md
```diff
@@ -43,6 +43,15 @@ Starting 2023-03-01 Puppet 6 is end of life. Users should upgrade to Puppet 7.
 
 ### Release Notes
 
+### Release notes for 3.6.1
```
ekohl, line 46: I noticed this is the same header as above, but I think that's been this way for a long time so I didn't address it.

_includes/version.html
```diff
@@ -1,8 +1,8 @@
 <div class="version-marker">
   <p>
     Latest versions:
-      <a href="{{ site.baseurl }}manuals/3.4/index.html#Releasenotesfor3.6">3.6.0</a>
-      /
```
ekohl, line 5: This `/` is now placed using CSS

PR #2077, opened 2023-04-19
Update release notes for Foreman 3.5.3

ianballou:
I didn't append anything to hammer because the release is back to 3.5.1 -- should I have? I didn't see any hammer-related notes to base it off of in the 3.5 changelog.

### DIFF
--- a/_includes/manuals/3.5/1.2_release_notes.md
+++ b/_includes/manuals/3.5/1.2_release_notes.md
@@ -59,6 +59,24 @@ It is safe to remove the file or remove all content.
   This data used to be part of Foreman, but was extracted to a plugin in Foreman 2.3.
   Until now this data was kept in the database, but now this data will be purged unless the plugin is installed.
 
+### Release notes for 3.5.3
+#### Foreman - Unattended Installations
+* preseed_netplan_generic_interface with DHCP interface - ([#35578](https://projects.theforeman.org/issues/35578))
+
+#### Foreman - API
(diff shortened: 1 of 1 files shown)


_includes/manuals/3.5/1.2_release_notes.md
```diff
@@ -59,6 +59,24 @@ It is safe to remove the file or remove all content.
   This data used to be part of Foreman, but was extracted to a plugin in Foreman 2.3.
   Until now this data was kept in the database, but now this data will be purged unless the plugin is installed.
 
+### Release notes for 3.5.3
+#### Foreman - Unattended Installations
+* preseed_netplan_generic_interface with DHCP interface - ([#35578](https://projects.theforeman.org/issues/35578))
```
ianballou, line 64: I had to add this by hand since the PR was added to an older Redmine.

ekohl: I added it as Fixed in version, so it should generate properly now. Normally we would have created a new bug and used `Fixes` on that new bug, relating it to the old one. However, that wasn't done so this is the next best thing.

ekohl:
> I didn't append anything to hammer because the release is back to 3.5.1 -- should I have? I didn't see any hammer-related notes to base it off of in the 3.5 changelog.

That's fine. I've been struggling with that part of the changelog. It's been in the procedure for ages but I don't think we've done that either for ages. I've postponed it, but perhaps it's time to put it on the roadmap again.

ekohl (approved):

PR #2079, opened 2023-04-25
Update 3.5.3 latest release

ianballou:
Bumping the patch (that's z-stream right?) to match 3.5.3.


ekohl (approved):
Correct. I'm starting to wonder if this should be part of the same PR as the release notes. This is still an area I'm improving  so I'd welcome your feedback.

evgeni:
The release notes is what technically announces the release on the site, so I think updating the "latest version" in the same PR/commit makes total sense

PR #2092, opened 2023-08-22
Update release notes to include Foreman 3.6.2

Griffin-Sullivan:


ekohl (changes requested):
Sorry for the slow review. Could you also regenerate these like you did with the foreman-documentation PR?

ekohl (approved):

PR #2105, opened 2023-10-10
Update releases notes for Foreman 3.8 GA

Griffin-Sullivan:


ekohl (changes requested):
This also needs to update `_config.yml`. From the latest procedure:
  - Update [`_config.yml`](https://github.com/theforeman/theforeman.org/blob/gh-pages/_config.yml)
    - Update the `foreman_version` to 3.8
    - Update `foreman_latest_releases`

And the release note you added in foreman-documentation also needs to be done here.

_config.yml
```diff
@@ -9,11 +9,11 @@ url: https://theforeman.org
...
 # latest *stable* version
-foreman_version: '3.7'
+foreman_version: '3.8'
 
 # Latest releases
 foreman_latest_releases:
-  - minmaj: '3.7'
+  - minmaj: '3.8'
     patch: '0'
   - minmaj: '3.6'
```
ekohl: You need to modify this to 3.7. The result should be on https://theforeman.org/ under "Latest versions".

ekohl (approved):

PR #2106, opened 2023-10-18
Update the chat information

ares:


support.md
```diff
@@ -69,61 +69,41 @@ In addition to the [general guidelines](/support.html#CommunityGuidelines) above
...
-bug](http://projects.theforeman.org/projects/foreman/issues), the docs may be
-lacking, or it's possible Foreman doesn't yet support your use case. It just
-means you need to keep troubleshooting to find out which.
+The Matrix rooms are no longer bridged to the IRC channels (see below).
 
-### IRC Guidelines
+### Chat Guidelines
 
-In addition to the [general guidelines](/support.html#CommunityGuidelines) above, we also have a few IRC-specific guidelines:
+In addition to the [general guidelines](/support.html#CommunityGuidelines) above, we also have a few Matrix-specific guidelines:
```
evgeni: I'd s/Matrix/chat/ here, as the protocol does not influence the guidelines

support.md
```diff
@@ -69,61 +69,41 @@ In addition to the [general guidelines](/support.html#CommunityGuidelines) above
 
 (_Credit to the [Fedora Mailing List Guidelines](http://fedoraproject.org/wiki/Communicate/MailingListGuidelines) as a source of ideas, under the Creative Commons Attribution-ShareAlike 3.0 Unported license_)
 
-## [libera.chat](https://libera.chat) {#IRClivechat}
-A great place to get real-time help with The Foreman is our IRC channel:
-
-* **#theforeman** on [libera.chat](https://libera.chat/)
+## ![Matrix icon](static/images/matrix.png) [Matrix.org live chat](https://riot.im/app/#/room/#theforeman:matrix.org)
```
omenos: Might be better to use the more generic URL for the matrix rooms?

```
https://matrix.to/#/#theforeman:matrix.org
```

Or updating to use the Element domain:

```
https://app.element.io/#/room/#theforeman:matrix.org
```

support.md
```diff
@@ -69,61 +69,41 @@ In addition to the [general guidelines](/support.html#CommunityGuidelines) above
...
 
-Please read and understand [this fantastic guide](http://workaround.org/getting-help-on-irc)
-to getting help for open-source projects on IRC before diving in. All of the
-points there apply to #theforeman, especially "Don't repeat yourself", "Don't
-ask to ask", and "Stick around".  #theforeman in particular has concentrations
-of people in Europe (CEST) and East-coast US (EST), so asking your question
-when those time zones are in business hours is more likely to get a good
-result.
+* https://riot.im/app/#/room/#theforeman:matrix.org
+* https://riot.im/app/#/room/#theforeman-dev:matrix.org
```
omenos: Same as L72:

```
https://matrix.to/#/#theforeman:matrix.org
https://matrix.to/#/#theforeman-dev:matrix.org
```

support.md
ares: These takes me through 2 more screen though, right? I like it mentions more matrix clients. Any other benefit? (I'm not opposed to it)

omenos: The simple change would just be to use the new Element URL rather than the legacy Riot domain. The matrix.to share link lets the user access the room with the client of their choice, but it does involve a few extra clicks if they don't already have an app or web session ready to go.

Either way, a user can't preview the room/discussion until they have an account on some server. While linking directly to the room makes sense, using matrix.to makes sense from the Matrix community perspective of showcasing the decentralized nature of the platform. Whether or not that will just be more confusing for new non-Matrix users... I'll leave that to someone else to determine as I am a biased source.

ares: makes sense now, thanks, I've updated the domain but left the direct link to the room in element, as that's the fastest way to get involved, people can find better client or get an account on another server later.

ares:
Updated, ready for re-review

omenos (approved):
lgtm

ekohl (approved):

PR #2103, opened 2023-09-27
Update Puppet compatibility matrix

Areyus:
The Puppet compatibility matrix has not been updated regularly and is currently outdated, stating that Puppet 6 is still supported in Foreman 3.7.

Updated the matrix and relevant text passages found according to what was in the changelog since Foreman 3.0

PR #2107, opened 2023-10-19
Update Puppet Smart-Proxy compatibility

Areyus:
Smart-Proxy compatibility changed in Foreman 3.4 to only support Puppet>=6.
I also copied #2103 to the docs for 3.8 and nightly, which were missing in the original PR.

### DIFF
--- a/_includes/manuals/3.4/3.1.3_puppet_versions.md
+++ b/_includes/manuals/3.4/3.1.3_puppet_versions.md
@@ -8,22 +8,10 @@ Foreman integrates with Puppet and Facter in a few places, but generally using a
     <th>Smart Proxy</th>
   </tr>
   <tr>
-    <td>0.x-4.3</td>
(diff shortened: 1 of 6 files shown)


ekohl:
This looks to be mostly the same as https://github.com/theforeman/theforeman.org/pull/2108 but that's a bit more complete. I also provided references there. Would you mind checking if I made any mistakes there?

Areyus:
@ekohl as already mentioned on IRC, the only main difference I spotted between the PRs is that #2108 marks Puppet 6.x as supported for Foreman 3.7+, while this PR marks them as deprecated (since they are EOL).
I'm fine with both, depending on what the project considers "supported" for this situation.

ekohl:
I went back on forth on that myself and have had both versions. I think I'll end up with going for deprecated since it's EOL. I'll squash in my notes into the commit message of yours so they're kept in `git log`.

PR #2111, opened 2023-11-07
Add 3.7.1 release notes

Griffin-Sullivan:


ekohl (changes requested):
There is a merge conflict in the committers script now. I'd also think the list of contributors would be updated, just like in foreman-documentation. Of course, if we postpone that to a separate step then you'd probably do the same here.

_config.yml
```diff
@@ -16,7 +16,7 @@ foreman_latest_releases:
   - minmaj: '3.8'
     patch: '0'
   - minmaj: '3.7'
-    patch: '0'
+    patch: '1'
```
pcreech: With 3.9 now out, there is no '3.7' section.  This will need to be removed.  https://github.com/theforeman/theforeman.org/blob/gh-pages/_config.yml#L15

scripts/committers.rb
```diff
@@ -69,6 +69,103 @@
   puppetdb_foreman
 )
 
+@author_map = {
```
pcreech: Based on the contents of this file today in gh-pages, author_map is no longer needed.

ekohl (approved):

PR #2126, opened 2024-02-14
Replace the archtitecture image with a newer one

dgoetz:


maximiliankolb (approved):
TFTP is kinda missing compared to the old graphic, but :+1: for updating it! I like it a lot!

Any other input?

ekohl:
@maximiliankolb in https://community.theforeman.org/t/rfc-github-profile-for-foreman-and-katello/34432 there was already discussion.

Regarding TFTP: we also have HTTPBoot so I imagine it's captured in the ... block.

ekohl (approved):

PR #2129, opened 2024-02-16
Update plugin overview page

ekohl:
See individual commits for details.


ekohl:
I just noticed we still package graphite and list it on https://theforeman.github.io/foreman-plugin-overview/ so now I wonder if we should still drop it here. Or drop it completely.

evgeni (approved):

PR #2144, opened 2024-05-14
Add 3.11 folder and update nightly

Griffin-Sullivan:
1. Copied nightly content to 3.11
2. Cleaned deprecation and upgrade warnings in nightly manual
3. Ran `get-params` for nightly


zjhuntin:
Info looks right to me, still would be interested to hear @ekohl's review. I tried to run it locally, but ran into issues with my environment, but I didn't see anything that looked like an error when generating data.

ekohl (approved):

PR #2148, opened 2024-06-11
drop CentOS Stream 8 from "supported platforms"

evgeni:


evgeni:
Should be applied to at least 3.9-3.11, IMHO

ehelms (approved):

PR #2150, opened 2024-06-23
Add faraday-retry to Gemfile

bastelfreak:
It's an optional dependency of octokit. Without it we always get a warning on the CLI:

```
To use retry middleware with Faraday v2.0+, install `faraday-retry` gem
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,3 +3,4 @@ source "https://rubygems.org"
 gem 'github-pages', '>= 147'
 gem 'rake'
 gem 'webrick'
+gem 'faraday-retry'


evgeni:
But we're not using octokit, so why pulling in another dep?

bastelfreak:
octokit is a dependency of github-pages-health-check, jekyll-gist and jekyll-github-metadata. octokit by default uses faraday which in turn wants to use faraday-retry and complains when it's missing. It's a bit of a rabbithole :( At Vox Pupuli we've the same issue with the github_changelog_generator that also pulls in octokit => faraday => faraday-retry. That's why we added faraday-retry there as well.

evgeni:
I know. But we're still not actively using any of that, so I see no reason to add another dependency just to silence that warning.

PR #2156, opened 2024-06-29
fix 3.11 and nightly manual rendering

evgeni:
<List> is not a valid HTML tag, but breaks rendering of the page


evgeni:
@ekohl I did those changes manually, but whatever generates this table needs a fix ;)

PR #2159, opened 2024-07-02
Update 2021-02-17-introduction-to-the-remote-execution-plugin.md

voom:
Fix typo in installation command


ekohl (changes requested):
This is a historic post and at that point the correct command was `--enable-foreman-proxy-plugin-remote-execution-ssh` and I doubt we should change it in this way. There's a good chance other things have also changed.

@ares since it's your blog, what do you think?

PR #2160, opened 2024-07-02
Update index.md

voom:
Fix typo in installation command


ekohl (commented):
This isn't really a typo, but more a matter of it not having been updated to reflect the rename. Overall this whole page hasn't been maintained for a long time.

@adamruzicka I still have https://github.com/theforeman/foreman_remote_execution/pull/820 open, but perhaps we should migrate away from these pages altogether and retire them?

adamruzicka:
> perhaps we should migrate away from these pages altogether and retire them?

That would be the best course of action

PR #2181, opened 2024-09-23
less libera

evgeni:
- **link to support page on "getting help" page**
- **drop last remains of libera.chat in pages**


evgeni:
@ares places you forgot :)

ehelms (approved):

PR #2182, opened 2024-09-23
[3.12 GA] update config file for versions 

ColeHiggins2:


evgeni (changes requested):
This will break the rendering of the page :(
That's a bug in the script generating the options. It doesn't properly convert html special chars.

Given this patch only changes the < and > I think we can safely ignore it for now (and fix the script for the next time)

evgeni (approved):
I've dropped the installer params commit and merging the config one

evgeni:
installer params fix in https://github.com/theforeman/kafo/pull/377

PR #2196, opened 2024-11-07
Add release notes for 3.13

ColeHiggins2:


ColeHiggins2:
@ekohl Do we want to add anything to the headline features, deprecations, and upgrade warnings?

_includes/manuals/3.13/1.2_release_notes.md
```diff
@@ -23,6 +23,89 @@ Clients will remain supported.
 
 ### Release Notes
```
ekohl: I think we can drop this header
```suggestion

```

_includes/manuals/3.13/1.2_release_notes.md
ekohl, line 1: I think line 3 (`This section will be updated prior to the next release.`) should be removed.

ekohl (changes requested):
I think we should add a note that Debian 11 is deprecated and will be dropped with Foreman 3.14, but I can do that tomorrow if you'd like. Some inline notes to improve readability.

ekohl:
If you agree, could you merge & include https://github.com/theforeman/foreman-documentation/pull/3452 here? Note this site doesn't have ifdef support so just include it unconditionally.

_includes/manuals/3.13/1.2_release_notes.md
ekohl: To be explicit, that's this line:

https://github.com/theforeman/theforeman.org/blob/035e0eab0b9115cfc8b818ad9f85325f44d593d9/_includes/manuals/3.13/1.2_release_notes.md?plain=1#L3

ekohl (approved):
I'd like to see https://github.com/theforeman/theforeman.org/pull/2196#discussion_r1838770476 resolved before we announce, but the PR in itself is correct.

ColeHiggins2:
@ekohl Just updated. Must have missed that comment

PR #2201, opened 2024-11-25
Update installer options 3.13

ColeHiggins2:


ColeHiggins2:
@ekohl When viewing the file itself. we do see some text cut off. BUT this is occurring on previous versions too. We should fix this

ekohl (approved):

PR #2206, opened 2024-11-28
Add GitHub Action to deploy

ekohl:
Today we use GitHub Actions to build and deploy using Jenkins. This implements a deployment using GitHub Actions so we use a single environment.

The design is to build the website just like we do in CI, but then set up SSH for rsync followed by the actual rsync.

It introduces 3 secrets that are intended to be set as environment secrets:

* `SSH_PRIVATE_KEY`
* `SSH_KNOWN_HOSTS`
* `DEPLOY_TARGET`

Link: https://docs.github.com/en/actions/use-cases-and-examples/deploying/deploying-with-github-actions

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,35 @@
+name: Deployment
+
+concurrency:
+  group: production
+
+on:
+  push:
+    branches:
+      - gh-pages
+
+jobs:
+  deployment:
+    runs-on: ubuntu-latest
+    environment: production
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '3'
+          bundler-cache: true
+      - name: Build website
+        run: bundle exec rake build
+      - name: Set up SSH key
+        run: |
+          mkdir -p ~/.ssh
+          echo "${{ secrets.SSH_PRIVATE_KEY }}" > ~/.ssh/id_rsa
+          chmod 600 ~/.ssh/id_rsa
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,35 @@
...
+          mkdir -p ~/.ssh
+          echo "${{ secrets.SSH_PRIVATE_KEY }}" > ~/.ssh/id_rsa
+          chmod 600 ~/.ssh/id_rsa
+          echo "${{ secrets.SSH_KNOWN_HOSTS }}" > ~/.ssh/known_hosts
+      - name: Create a log file
+        run: echo "rsync_log=$(mktemp)" >> "$GITHUB_ENV"
+      - name: Copy files
+        run: rsync --log-file "${rsync_log}" --log-file-format 'CHANGED %f' --archive --checksum --verbose --one-file-system --compress --stats --delete-after ./_site/ ${{ secrets.DEPLOY_TARGET }}
+      - name: Purge CDN
+        run: awk '/ CHANGED /{print $5}' "${rsync_log}" | xargs --no-run-if-empty fastly-purge 'https://theforeman.org/'
```
ekohl, line 35: `fastly-purge` is not present now so moving to draft.

evgeni:
Today this won't work, as we limit who is allowed to deploy the website:
https://github.com/theforeman/foreman-infra/blob/7c342ff6b6c9cb3462b21b3c8747cfe80f2de325/puppet/modules/web/manifests/vhost/web.pp#L92-L96

Not saying we should not change that, just that this requires more work, which I personally don't have the time for right now.

PR #2214, opened 2025-02-08
fix: typos in documentation files

maximevtush:


ekohl (changes requested):
Thank you for these, but version 1.14 has been long unsupported so you should really fix things in nightly and then see about the stable versions.

As for the installer option: it's generated from the source and I submitted https://github.com/theforeman/puppet-puppet/pull/927 to fix it there.

The other is no longer present in recent versions.

PR #2221, opened 2025-03-12
Add 3.14.0 headline features

ColeHiggins2:
Adding 3.14.0 Headline Features to release notes

### DIFF
--- a/_includes/manuals/3.14/1.2_release_notes.md
+++ b/_includes/manuals/3.14/1.2_release_notes.md
@@ -188,7 +188,6 @@ For more details, see the [removal RFC](https://community.theforeman.org/t/drop-
 
 *A full list of changes in 3.14.0 is available via [Redmine](https://projects.theforeman.org/issues?set_filter=1&sort=id%3Adesc&status_id=closed&f[]=cf_12&op[cf_12]=%3D&v[cf_12]=1907)*
 
-
 ### Contributors
 
 We'd like to thank the following people who contributed to the Foreman {{page.version}} release:
--- a/_includes/manuals/3.14/1.2_release_notes.md
+++ b/_includes/manuals/3.14/1.2_release_notes.md
@@ -39,6 +39,11 @@ For more details, see the [removal RFC](https://community.theforeman.org/t/drop-
 ### Release Notes
 
 ### Release notes for 3.14.0
+#### Foreman
+* Failure occurs after selecting the correct value while creating an Ansible variable. ([#38281](https://projects.theforeman.org/issues/38281))
+* Can't provision using image + SSH ([#38272](https://projects.theforeman.org/issues/38272))
+* AAP 2.5 support in Ansible Callback ([#38219](https://projects.theforeman.org/issues/38219))
+
 #### Foreman - API
 * Hide organization-id and location-id options from api documentation of unscoped resources ([#37824](https://projects.theforeman.org/issues/37824))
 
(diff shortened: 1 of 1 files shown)


ColeHiggins2:
@ekohl can you review/tag anyone else you think might be useful to get feedback from :)

_includes/manuals/3.14/1.2_release_notes.md
```diff
@@ -4,6 +4,17 @@ This section will be updated prior to the next release.
 
 ### Headline features
 
+#### IPV6 Support
```
ekohl: Typically it's written as
```suggestion
#### IPv6 Support
```

_includes/manuals/3.14/1.2_release_notes.md
```diff
@@ -4,6 +4,17 @@ This section will be updated prior to the next release.
...
+Satellite deployment and operation in IPv6 networks is now fully supported. Dual stack deployments have not been tested and are not supported.
```
ekohl: Please fix the branding. In upstream we also have a different definition of supported. Red Hat interprets supported as you can call support if it's broken while in upstream you can only ask around in the community. They don't care about supported vs not supported so probably leave it out and emphasize IPv6-only.

Something like this perhaps?

```suggestion
Deployment and operation in IPv6-only networks is now fully supported.
```

_includes/manuals/3.14/1.2_release_notes.md
```diff
@@ -4,6 +4,17 @@ This section will be updated prior to the next release.
...
+
+Provisioning over IPv6 is supported on bare metal hosts.
+For compute resources, you can define the machine outside Satellite on the compute resource and then you can provision the machine as bare metal in Satellite.
+
+#### Support for Secure Boot provisioning
```
ekohl: I'd keep the header short
```suggestion
#### Secure Boot provisioning
```

_includes/manuals/3.14/1.2_release_notes.md
```diff
@@ -4,6 +4,17 @@ This section will be updated prior to the next release.
 
 ### Headline features
 
+#### IPV6 Support
+Satellite deployment and operation in IPv6 networks is now fully supported. Dual stack deployments have not been tested and are not supported.
+
+Provisioning over IPv6 is supported on bare metal hosts.
+For compute resources, you can define the machine outside Satellite on the compute resource and then you can provision the machine as bare metal in Satellite.
```
ekohl: Why would image mode not work on IPv6-only? If that isn't supported I wouldn't call IPv6-only fully supported.

Also, please fix the branding.

_includes/manuals/3.14/1.2_release_notes.md
```diff
@@ -4,6 +4,17 @@ This section will be updated prior to the next release.
...
+
+#### Support for Secure Boot provisioning
+Provide support for Secure Boot provisioning on bare-metal, VMware vSphere and Libvirt across multiple operating systems.
```
ekohl: I don't think the word `Provide` adds a lot.

_includes/manuals/3.14/1.2_release_notes.md
```diff
@@ -4,6 +4,17 @@ This section will be updated prior to the next release.
...
+
+#### Invalidating JWT for global registration
```
ekohl: Looking at the writers, but I think this reads better as a title.
```suggestion
#### Invalidate JWT for global registration
```

ekohl (commented):
This may be just me, but I always like the visual separation between sections. Something like
```
# header 1

paragraph 1

# header 2

paragraph 2
```

ekohl (approved):

PR #2222, opened 2025-03-21
Update config file 3.14.0 GA

ColeHiggins2:
the get-params script made no additional changes to the manuals/3.14/3.2.2_installer_options-help.md and manuals/3.14/3.2.2_installer_options-params.md files. only change here is the config !

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -9,13 +9,13 @@ url: https://theforeman.org
 baseurl: "" #keep at "" for the Atom feed
 
 # latest *stable* version
-foreman_version: '3.13'
+foreman_version: '3.14'
 
 # Latest releases
 foreman_latest_releases:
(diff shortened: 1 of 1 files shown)


_config.yml
```diff
@@ -9,13 +9,13 @@ url: https://theforeman.org
...
+foreman_version: '3.14'
 
 # Latest releases
 foreman_latest_releases:
-  - minmaj: '3.13'
+  - minmaj: '3.14'
     patch: '0'
-  - minmaj: '3.12'
+  - minmaj: '3.13'
     patch: '1'
```
ekohl: There is no 3.13.1
```suggestion
    patch: '0'
```

ekohl (changes requested):

PR #2225, opened 2025-04-17
Use correct Markdown link format

nachtjasmin:


ogajduse (approved):

ogajduse:
Thank you for your patch.
I've rebased your patch to trigger the "jeykyll" check that is now required.

PR #2234, opened 2025-05-28
Guidelines to set breaks-robottelo label

lhellebr:
This documents a process regarding setting `breaks-robottelo` label to PRs as implemented here: https://github.com/theforeman/foreman/pull/10553

### DIFF
--- a/contribute.md
+++ b/contribute.md
@@ -404,6 +404,22 @@ git push origin develop
 
 Now follow step 4 to the end from above.
 
+#### Labelling pull requests for QA
+If your PR makes a change that would cause a Robottelo (not unit) test to fail, add the following comment to the PR:
+
+{% highlight bash %}
+/label breaks-robottelo
+{% endhighlight %}
+
+Changes that can cause Robottelo tests to fail include changes that break interface in a way that a user would notice or changes that affect parts of the WebUI used to navigate our automated tests. For example:
+
+* id, classname or ouia id changes, especially (but not only) on interactable items like links and buttons
+* WebUI DOM change (including package updates leading to it)
+* API changes (endpoint gets removed, parameters change, return value/type changes)
+* adding or adjusting required flags for CLI procedures
+
+Adding the comment marks your PR with the `breaks-robottelo` label. The label helps QA with early detection of changes that would break tests in Robottelo.
+
 ### Projects / mentoring
 Please see [this page](/mentoring.html) for details on our current projects.
 


contribute.md
```diff
@@ -404,6 +404,22 @@ git push origin develop
 
 Now follow step 4 to the end from above.
 
+#### Labelling pull requests
+In order to help QA with early detection of tests that would break tests in Robottelo, please comment exactly
```
adamruzicka:
```suggestion
In order to help QA with early detection of changes that would break tests in Robottelo, please comment exactly
```

contribute.md
```diff
@@ -404,6 +404,22 @@ git push origin develop
...
+
+{% highlight bash %}
+/label breaks-robottelo
+{% endhighlight %}
+
+to your PR if you have a reason to think it makes a change that would cause some Robottelo (not unit) test to fail. Generally, it's changes that break interface in a way that a user would notice or change things in WebUI that are used to navigate our automated tests. It's not always easy to decide, use your best judgement. Major painpoints have been, historically:
+
+* id, classname or ouia id changes, especially (but not only) on interactable items like links and buttons
+* webUI DOM change (including package updates leading to it)
```
adamruzicka:
```suggestion
* WebUI DOM change (including package updates leading to it)
```

adamruzicka (commented):
Since this will go out on tfm.org, we might want to polish it a bit. @aneta-petrova could you please take a look from the language perspective?

contribute.md
```diff
@@ -404,6 +404,22 @@ git push origin develop
 
 Now follow step 4 to the end from above.
 
+#### Labelling pull requests
```
aneta-petrova: Can you try to be more specific in the heading and add the purpose of the labelling? Something like:
```suggestion
#### Labelling pull requests for Robottelo test failure detection
```
or perhaps simply:
```suggestion
#### Labelling pull requests for QA
```
I'm proposing this because there are other labels in use and the update talks about only one of them.

contribute.md
```diff
@@ -404,6 +404,22 @@ git push origin develop
...
+{% endhighlight %}
+
+to your PR if you have a reason to think it makes a change that would cause some Robottelo (not unit) test to fail. Generally, it's changes that break interface in a way that a user would notice or change things in WebUI that are used to navigate our automated tests. It's not always easy to decide, use your best judgement. Major painpoints have been, historically:
+
+* id, classname or ouia id changes, especially (but not only) on interactable items like links and buttons
+* WebUI DOM change (including package updates leading to it)
+* API changes - endpoint gets removed, parameters change, return value/type changes
+* adding or adjusting required flags for cli procedures
+
+The comment marks the PR with `breaks-robottelo` label.
```
aneta-petrova: What would you think of this? Summary of the changes I'm proposing:

* Better to start with *what* needs to be done and *when*. The *why* with the explanation, while important, is secondary and should be more of an FYI at the end.
* Starting with a description of *when* ("If your PR makes a change...") helps users easily skip this part if it doesn't apply to them and saves them from having to read the whole section.
* I made some statements a bit shorter by using more minimalist wording. I think it's still understandable like this and hopefully it gets the point across faster.

```suggestion
If your PR makes a change that would cause a Robottelo (not unit) test to fail, add the following comment:

{% highlight bash %}
/label breaks-robottelo
{% endhighlight %}

Changes that can cause Robottelo tests to fail include changes that break interface in a way that a user would notice or changes that affect parts of the WebUI used to navigate our automated tests. For example:

* id, classname or ouia id changes, especially (but not only) on interactable items like links and buttons
* WebUI DOM change (including package updates leading to it)
* API changes (endpoint gets removed, parameters change, return value/type changes)
* adding or adjusting required flags for CLI procedures

Adding the comment marks your PR with the `breaks-robottelo` label. The label helps QA with early detection of changes that would break tests in Robottelo.
```

When it comes to "add the following comment", it would be nice to specify where to add the comment. The code? The PR description?

lhellebr:
Changed as per Aneta's suggestion, and added ` to the PR:` to the first paragraph

aneta-petrova (approved):

adamruzicka (approved):

PR #2242, opened 2025-08-27
Automate Hammer CLI release notes integration

ogajduse:
## Summary
- Automate Hammer CLI release notes integration in `scripts/release_notes.rb`
- Add configurable category mapping system for maintainable transformations
- Eliminate manual copy-paste work for Hammer CLI components during releases

## Background
Manual integration of Hammer CLI release notes into Foreman documentation is time-consuming and error-prone during releases. This addresses [theforeman/theforeman-rel-eng#514](https://github.com/theforeman/theforeman-rel-eng/issues/514) by automating the process through the existing `release_notes.rb` script.

### Core Functionality
- Extended `release_notes.rb` to fetch both `hammer-cli` and `hammer-cli-foreman` release notes from Redmine
- Added support for version prefixes (e.g., `hammer-cli-3.16.0`, `hammer-cli-foreman-3.16.0`)
- Integrated Hammer CLI sections into the existing release notes workflow

### Configurable Category Mapping System
- Replaced hardcoded category transformations with configurable `CATEGORY_MAPPINGS`
- Added support for:
  - **Merge mappings**: Combine multiple source categories into target categories
  - **Project-specific transformations**: Apply different rules based on component type
  - **Regex-based renaming**: Flexible pattern matching for category names

### Output Format
The script now automatically generates properly formatted Hammer CLI sections:

```diff
+#### Hammer CLI - Hammer core
+* Translated paths for replaced_by fields may differ from actual path ([#38427](https://projects.theforeman.org/issues/38427))
+
+#### Hammer CLI - Foreman - Compute resources
+* Remove Ovirt from Hammer ([#38294](https://projects.theforeman.org/issues/38294))
+
+#### Hammer CLI - Foreman - Users and roles
+* LDAP group membership ([#38611](https://projects.theforeman.org/issues/38611))
```

## Usage
```bash
ruby scripts/release_notes.rb foreman 3.16.0
```

Now automatically includes Hammer CLI and Hammer CLI - Foreman sections alongside Foreman release notes.

## Benefits
- ✅ **Eliminates manual work**: No more copy-paste from Hammer CLI repositories
- ✅ **Maintains consistency**: Automated formatting ensures uniform output
- ✅ **Easily extensible**: New category mappings can be added without code changes

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/scripts/release_notes.rb
+++ b/scripts/release_notes.rb
@@ -19,6 +19,34 @@ end
 @project_name = ARGV[0]
 @current_release_name = ARGV[1]
 
+# Additional projects to include in release notes
+# Both hammer-cli and hammer-cli-foreman versions are stored in the 'hammer-cli' project
+ADDITIONAL_PROJECTS = [
+  { project: 'hammer-cli', version_prefix: 'hammer-cli-' },
+  { project: 'hammer-cli', version_prefix: 'hammer-cli-foreman-' }
+].freeze
+
+# Category mappings and transformations
+CATEGORY_MAPPINGS = {
+  # Merge categories - source categories are merged into target category
+  merge: {
+    'Smart Proxy' => ['Smart Proxy - Core'],
+    'SELinux' => ['SELinux - General Foreman']
+  },
+
+  # Rename categories - simple string replacements using regex patterns
+  rename: {
+    /^Hammer CLI$/ => 'Hammer CLI',  # Keep as-is for base hammer-cli
+  },
+
+  # Project-specific category transformations
+  project_specific: {
+    'hammer-cli-foreman-' => {
+      /^Hammer CLI/ => 'Hammer CLI - Foreman'
+    }
+  }
+}.freeze
+
 def url_to_json(url)
   uri = URI(url)
   response = Net::HTTP.get(uri)
@@ -26,40 +54,42 @@ def url_to_json(url)
 end
 
 # Determine version
-def get_version_id(project)
+def get_version_id(project, release_name = @current_release_name)
   url = "#{URL}/projects/#{project}/versions.json"
   result = url_to_json(url)
   return nil if result['versions'].nil?
-  result = result['versions'].detect{|v| v['name'] == @current_release_name }
+  # Find version with matching name (e.g., "3.16.0" or "hammer-cli-3.16.0")
+  result = result['versions'].detect{|v| v['name'] == release_name }
   return nil if result.nil?
   result['id']
 rescue StandardError => e
   puts "Error getting version list from #{url}: #{e}"
-  exit 1
+  return nil
 end
 
 @current_release_id = get_version_id(@project_name)
 
 if @current_release_id.nil?
-  puts "Release #{@current_release_name} not found in any of the projects #{@project_name}"
+  puts "Release #{@current_release_name} not found in project #{@project_name}"
   exit 1
 end
 
-def gather_issues(offset = 0)
+def gather_issues(project_name, release_id, offset = 0)
   params = {
     'status_id' => 'closed',
     'offset' => offset,
     'limit' => 100,
-    'f[]' => 'cf_12',
-    'op[cf_12]' => '=',
-    'v[cf_12][]' => @current_release_id,
+    'f[]' => 'cf_12',           # Filter by custom field 12 (Fix version)
+    'op[cf_12]' => '=',         # Operator: equals
+    'v[cf_12][]' => release_id, # Value: the release ID we want
   }
   url = "#{URL}/issues.json?#{URI.encode_www_form(params)}"
   result = url_to_json(url)
+  # Recursively fetch all pages of results (Redmine paginates at 100 items)
   if result['total_count'].to_i - offset - 100 <= 0
     result['issues']
   else
-    result['issues'] += gather_issues(offset + 100)
+    result['issues'] += gather_issues(project_name, release_id, offset + 100)
   end
 end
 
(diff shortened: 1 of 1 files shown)


scripts/release_notes.rb
```diff
@@ -19,47 +19,77 @@
...
   return nil if result['versions'].nil?
-  result = result['versions'].detect{|v| v['name'] == @current_release_name }
+  # Find version with matching name (e.g., "3.16.0" or "hammer-cli-3.16.0")
+  result = result['versions'].detect{|v| v['name'] == release_name }
   return nil if result.nil?
   result['id']
 rescue StandardError => e
   puts "Error getting version list from #{url}: #{e}"
-  exit 1
+  return nil
```
ekohl, line 67: I think this should still exit if it fails. It may not be very clean, but it's a simple script and it's good enough error handling.
```suggestion
  exit 1
```

scripts/release_notes.rb
```diff
@@ -19,47 +19,77 @@
...
 
+# Additional projects to include in release notes
+# Both hammer-cli and hammer-cli-foreman versions are stored in the 'hammer-cli' project
+ADDITIONAL_PROJECTS = [
+  { project: 'hammer-cli', version_prefix: 'hammer-cli-' },
+  { project: 'hammer-cli', version_prefix: 'hammer-cli-foreman-' }
+].freeze
+
+# Category mappings and transformations
+CATEGORY_MAPPINGS = {
```
ekohl, line 30: I think this is too complex. Why don't you simplify this to various separate constants?

scripts/release_notes.rb
```diff
@@ -19,47 +19,77 @@
...
+  # Merge categories - source categories are merged into target category
+  merge: {
+    'Smart Proxy' => ['Smart Proxy - Core'],
+    'SELinux' => ['SELinux - General Foreman']
+  },
+
+  # Rename categories - simple string replacements using regex patterns
+  rename: {
+    /^Hammer CLI$/ => 'Hammer CLI',  # Keep as-is for base hammer-cli
+  },
```
ekohl, line 40: What's the point of this noop rename? Feels like this can just be dropped.

scripts/release_notes.rb
```diff
@@ -68,24 +98,87 @@ def gather_issues(offset = 0)
...
+  # Step 1: Handle merge mappings
+  CATEGORY_MAPPINGS[:merge].each do |target_category, source_categories|
+    grouped_issues[target_category] ||= []  # Initialize if doesn't exist
+    source_categories.each do |source_category|
+      if grouped_issues.key?(source_category)
+        # Move issues from source to target category and remove source
+        grouped_issues[target_category] += grouped_issues.delete(source_category)
+      end
+    end
+  end
```
ekohl, line 111: This feels way overengineered. The previous code was easier to understand IMHO. Now you have ~15 lines of code for what used to be 2 lines.

scripts/release_notes.rb
```diff
@@ -68,24 +98,87 @@ def gather_issues(offset = 0)
...
+
+  # Step 3: Apply general rename mappings
+  new_grouped_issues = {}
+  grouped_issues.each do |category, category_issues|
+    new_category = category
+    CATEGORY_MAPPINGS[:rename].each do |pattern, replacement|
+      new_category = new_category.gsub(pattern, replacement)
+    end
+    new_grouped_issues[new_category] = category_issues
+  end
```
ekohl, line 137: The whole renaming isn't used so I think this can be dropped
```suggestion
```

scripts/release_notes.rb
```diff
@@ -19,47 +19,77 @@
...
+  rename: {
+    /^Hammer CLI$/ => 'Hammer CLI',  # Keep as-is for base hammer-cli
+  },
+
+  # Project-specific category transformations
+  project_specific: {
+    'hammer-cli-foreman-' => {
+      /^Hammer CLI/ => 'Hammer CLI - Foreman'
+    }
+  }
```
ekohl, line 47: This took me a long time to figure out and again, IMHO a bit too complex.

scripts/release_notes.rb
```diff
@@ -68,24 +98,87 @@ def gather_issues(offset = 0)
...
+# Generate notes for main project
+generate_release_notes_for_project(@project_name, @current_release_name)
 
-  puts
+# Generate notes for additional projects (Hammer CLI components)
+ADDITIONAL_PROJECTS.each do |project_config|
+  generate_release_notes_for_project(project_config[:project], @current_release_name, project_config[:version_prefix])
 end
 
 puts "*A full list of changes in #{@current_release_name} is available via [Redmine](https://projects.theforeman.org/issues?set_filter=1&sort=id%3Adesc&status_id=closed&f[]=cf_12&op[cf_12]=%3D&v[cf_12]=#{@current_release_id})*"
```
ekohl, line 184: This is no longer true because you also need to include the Hammer releases.

scripts/release_notes.rb
```diff
@@ -19,47 +19,77 @@
...
   params = {
     'status_id' => 'closed',
     'offset' => offset,
     'limit' => 100,
-    'f[]' => 'cf_12',
-    'op[cf_12]' => '=',
-    'v[cf_12][]' => @current_release_id,
+    'f[]' => 'cf_12',           # Filter by custom field 12 (Fix version)
+    'op[cf_12]' => '=',         # Operator: equals
+    'v[cf_12][]' => release_id, # Value: the release ID we want
```
ekohl, line 84: You can repeat this parameter to gather the issues for multiple releases in a single API call. I think you can just change `params` to an array of arrays and repeat `v[cf12][]` for each release you want to include.

scripts/release_notes.rb
```diff
@@ -68,24 +98,87 @@ def gather_issues(offset = 0)
...
+
+  # Group issues by "Project Name" or "Project Name - Category" format
+  grouped_issues = issues.group_by { |issue| "#{issue['project']['name']}#{ ' - ' + issue['category']['name'] if issue.key?('category') }" }
+
+  # Apply configurable category mappings
+  grouped_issues = apply_category_mappings(grouped_issues, version_prefix)
 
-grouped_issues = gather_issues.group_by { |issue| "#{issue['project']['name']}#{ ' - ' + issue['category']['name'] if issue.key?('category') }" }
+  # Sort categories alphabetically
+  grouped_issues = Hash[ grouped_issues.sort_by { |key, val| key } ]
```
ekohl, line 158: I know this existed before, but this can be done easier in current Ruby (probably couldn't when it was written). The version in foreman-documentation already does this.
```suggestion
  grouped_issues = grouped_issues.sort
```

ekohl (commented):
I think this looks very complex and would suggest to start fresh. My recommendation would be to take https://github.com/theforeman/foreman-documentation/blob/master/guides/doc-Release_Notes/redmine_release_notes as a base. Then work on expanding that.

The first thing I would try is the trick I suggest inline with repeating `v[cf_12][]` multiple time, once for each release you want to include.

It also already cleaner where `category_from_issue` is already the method to extract the category from an issue. I think that's a better base then the complex mapping here.

Once you have the version in foreman-documentation working then I'd suggest to replace the script in this repo with a version that spits out markdown instead of asciidoc.

PR #2257, opened 2025-11-13
Simpler execution of committers script

ogajduse:
The script was failing when run directly due to the wrong Ruby interpreter in the shebang.

**Changes:**
- **Change shebang to use Ruby from PATH** - Changed from `#!/usr/bin/ruby` to `#!/usr/bin/env ruby` to respect the user's Ruby installation (rbenv, rvm, etc.)
- **Add usage instructions** - Added comments explaining how to run the script using the existing `scripts/Gemfile`

**Usage:**
```bash
bundle exec --gemfile=scripts/Gemfile ruby scripts/committers.rb TAG_FROM TAG_TO
```

The `git` gem dependency is managed in `scripts/Gemfile` (not the main Gemfile) to keep script dependencies separate from website building dependencies.

### DIFF
--- a/scripts/committers.rb
+++ b/scripts/committers.rb
@@ -1,6 +1,6 @@
 #!/usr/bin/env ruby
-# Run with: bundle exec --gemfile=scripts/Gemfile ruby scripts/committers.rb TAG_FROM TAG_TO
-# Or from scripts directory: bundle exec ruby committers.rb TAG_FROM TAG_TO
+# Run with: bundle exec --gemfile=scripts/Gemfile ruby ./committers.rb TAG_FROM TAG_TO
+# Or from scripts directory: bundle exec ./committers.rb TAG_FROM TAG_TO
 require 'tmpdir'
 require 'git'
 require 'optparse'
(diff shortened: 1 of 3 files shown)


ekohl (commented):
I don't think I ever ran the committers script via `bundle exec` and treated the `Gemfile` as only what's needed to build the website. It adds a fair number of dependencies so I'm not sure I entirely agree with adding it, but I see your point. I don't have a good answer on how to proceed now.

ogajduse:
Good point! I've updated the PR to remove the git gem from the main Gemfile. 

Instead, the script should be run using the existing `scripts/Gemfile` which already has the git gem dependency:

```bash
bundle exec --gemfile=scripts/Gemfile ruby scripts/committers.rb TAG_FROM TAG_TO
```

This way we keep script dependencies separate from website building dependencies. I've also added usage instructions as comments at the top of the script.

scripts/committers.rb
```diff
@@ -1,4 +1,6 @@
-#!/usr/bin/ruby
+#!/usr/bin/env ruby
+# Run with: bundle exec --gemfile=scripts/Gemfile ruby scripts/committers.rb TAG_FROM TAG_TO
+# Or from scripts directory: bundle exec ruby committers.rb TAG_FROM TAG_TO
```
ekohl: Bundler will set the right path so the env shebang works.

```suggestion
# Run with: bundle exec --gemfile=scripts/Gemfile ruby ./committers.rb TAG_FROM TAG_TO
# Or from scripts directory: bundle exec ./committers.rb TAG_FROM TAG_TO
```

PR #2287, opened 2026-03-30
docs: add Foreman Release Cycle

d1nuc0m:
Adding Foreman Release Cycle

It is a WIP - which is Katello schedule? The same of Foreman? Where can I find dates?

Probably style can be improved too, web dev is not my field

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -48,6 +48,8 @@ navigation:
         url: /manuals/latest/index.html
       - title: Plugins
         url: /plugins/
+      - title: Foreman Release Cycle
+        url: /release_cycle.html
       - title: Foreman API docs
         url: https://apidocs.theforeman.org/foreman/latest/apidoc/v2.html
       - title: Foreman+Katello API docs
--- a/release_cycle.md
+++ b/release_cycle.md
@@ -1,0 +1,83 @@
+---
+layout: default
+title: Foreman Release Cycle
+---
+
+# Foreman Release Cycle
+
+## Policy
+
+New Foreman and Katello versions are released roughly every three months and the two latest released minors are supported.
+
+NOTE: version numbers follow the MAJOR.MINOR.PATCH convention.
+
+Eg. when MAJOR.3.PATCH is released, MAJOR.2.PATCH is still supported and MAJOR.1.PATCH goes EOL.
+
+### Upgrades paths
+
+Each minor version can be directly upgraded to any of the following patches released _for the same minor_.
+Eg. both **MAJOR.MINOR.1 -> MAJOR.MINOR.2 -> MAJOR.MINOR.3** and **MAJOR.MINOR.1 -> MAJOR.MINOR.3** are valid upgrade paths.
+
+Each minor version can only be upgraded to the next minor version (regardless of patches versions).
+_Intermediate upgrades can not be skipped_.
+Eg. both **MAJOR.1.0 -> MAJOR.2.0 -> MAJOR.3.0** and **MAJOR.1.0 -> MAJOR.2.2 -> MAJOR.3.1** are valid upgrade paths.
+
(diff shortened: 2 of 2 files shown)


stejskalleos:
cc @jafiala @ogajduse

ogajduse (commented):
Thanks for putting this together, @d1nuc0m!

A few things to update:

3.19.0 GA [shipped June 9](https://community.theforeman.org/t/foreman-3-19-0-is-now-available/46768) - table still shows "EXP." The next release is **5.0** (not 3.20), with GA planned 2026-09-08. This was agreed in the [versioning sync RFC](https://community.theforeman.org/t/rfc-sync-foreman-and-katello-versioning-and-reset-both-to-5-0-with-containerized-deployment/45758) - the major bump reflects the shift to containerized deployments via `foremanctl` and permanently aligns Foreman + Katello versions. See the [5.0 schedule](https://community.theforeman.org/t/foreman-5-0-schedule-and-planning/46514).

The `X.Y.1 -> X.Y.2 -> X.Y.3` notation in the policy section reads like patch versions, but these are actually minor/major bumps (3.17, 3.18, 3.19). The policy is: two latest released versions are supported, where "version" means the last X or Y bump. So once 5.0 ships, 5.0 and 3.19 are supported, 3.18 goes EOL.

On the version history table - I'd question whether maintaining a full release date table on this page adds much value. It uses scheduled dates that differ from actuals (e.g. 3.16 announced 2025-09-10 not 09-09, 3.18 announced 2026-03-11 not 03-10), and it's already drifted (3.19 still marked expected, 5.0 missing). Each release already has its own "Schedule and Planning" post on Discourse, which the table links to anyway. Instead of maintaining this manually, `_config.yml` already has `foreman_latest_releases` with the currently supported versions — it gets updated every release. The page could use Liquid to render the supported versions directly from that data, so no manual page edits are needed — just the config update that already happens at release time. For the full history, a link to the Discourse [Release category](https://community.theforeman.org/c/release/20) would do.

On Katello - historically each Katello release shipped alongside a specific Foreman version (e.g. Katello 4.21 with Foreman 3.19). Starting with 5.0, both projects share the same version number. Worth mentioning this.

On the styling - the `nth-child` CSS breaks whenever rows are added or removed (all colors shift down). If you do keep a table, Bootstrap is already loaded site-wide, so `<span class="label label-success">Supported</span>` / `<span class="label label-important">End Of Life</span>` in the Status column would work better — status-based, not position-based.

d1nuc0m:
@ogajduse thanks for the review!

> 3.19.0 GA [shipped June 9](https://community.theforeman.org/t/foreman-3-19-0-is-now-available/46768) - table still shows "EXP."

Tbh I had not updated the PR not knowing if it may be interesting.

> The X.Y.1 -> X.Y.2 -> X.Y.3 notation in the policy section reads like patch versions, but these are actually minor/major bumps (3.17, 3.18, 3.19). The policy is: two latest released versions are supported, where "version" means the last X or Y bump. So once 5.0 ships, 5.0 and 3.19 are supported, 3.18 goes EOL.

Tried to clarify it - the two latest minors are supported

> On the version history table - I'd question whether maintaining a full release date table on this page adds much value. 

Idk, for history's sake? Switched to Liquid anyway

> On Katello - historically each Katello release shipped alongside a specific Foreman version (e.g. Katello 4.21 with Foreman 3.19). Starting with 5.0, both projects share the same version number. Worth mentioning this.

Done

> Bootstrap is already loaded site-wide, so <span class="label label-success">Supported</span> / <span class="label label-important">End Of Life</span> in the Status column would work better — status-based, not position-based.

Web dev is definitely not my field :D

Used `<span class="label label-danger">` instead of `<span class="label label-important">` as the latter seems not to be working.

PR #2295, opened 2026-05-15
Add Containerfile using official Ruby image

ogajduse:
Adds Containerfile using official Ruby image instead of unmaintained jekyll/jekyll.

Build once, then run:
```bash
podman build -t theforeman-org .
podman run --rm -v "$(pwd)":/srv/jekyll:Z -p 4000:4000 theforeman-org
```

Works with Docker too (replace `podman` with `docker`).

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -1,8 +1,7 @@
-FROM docker.io/library/ruby:3.1-alpine
+FROM docker.io/library/ruby:3.4-alpine
 WORKDIR /srv/jekyll
 
 RUN apk add --no-cache build-base
-RUN gem install bundler
 
 COPY Gemfile ./
 RUN bundle install
--- a/Containerfile
+++ b/Containerfile
@@ -7,7 +7,5 @@ RUN gem install bundler
 COPY Gemfile ./
 RUN bundle install
 
-COPY . .
-
 EXPOSE 4000
 CMD ["bundle", "exec", "jekyll", "serve", "--host", "0.0.0.0", "--watch"]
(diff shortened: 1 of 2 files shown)


ekohl (commented):
I suspect that docker will need the same installation instruction. Would it be better to write a `Containerfile` that just uses the Ruby container? https://hub.docker.com/r/jekyll/jekyll/ looks unmaintained and https://hub.docker.com/_/ruby/#create-a-dockerfile-in-your-ruby-app-project is probably better.

Containerfile
```diff
@@ -0,0 +1,24 @@
+FROM docker.io/library/ruby:3.1-alpine
+
+WORKDIR /srv/jekyll
+
+# Install build dependencies
+RUN apk add --no-cache build-base
+
+# Install bundler
+RUN gem install bundler
```
ekohl: I'd expect this to already be present in Ruby 3.1. Is this really needed?

Containerfile
```diff
@@ -0,0 +1,24 @@
...
+
+# Copy Gemfile for dependency installation
+COPY Gemfile ./
+
+# Install gems
+RUN bundle install
+
+# Copy the rest of the site
+COPY . .
```
ekohl: Is it really wise to copy the whole site into the container? IMHO it needlessly inflates the container and we don't intend to actually serve it but rather mount a volume.

ogajduse: That's what I realized now too. On it ...

Containerfile
ogajduse: Removed and upgraded to Ruby 3.4

Containerfile
ogajduse: Let me say that this was total nonsense. So, removed.

ekohl (approved):

PR #2310, opened 2026-08-18
Change s-p config file name to settings.yml

lhellebr:
While foreman has settings.yaml, s-p has settings.yml as shown in other
parts of this docs and also example file name


ogajduse (approved):
Is this something that needs an update in foreman-documentation too?

lhellebr:
I don't think so. This is probably some leftover in https://theforeman.org/contribute.html , users don't normally do that setup and quick search for the filename doesn't find anything where this would apply.

ogajduse:
Thank you, Lukas!

PR #2320, opened 2026-09-11
Add Noble and Trixie release notes

ogajduse:
Add Foreman 5.0 release-note and changelog entries for Ubuntu 24.04 (Noble) and Debian 13 (Trixie): https://projects.theforeman.org/issues/37520

sibbling PR: https://github.com/theforeman/foreman-documentation/pull/5307

### DIFF
--- a/_includes/manuals/5.0/1.2_release_notes.md
+++ b/_includes/manuals/5.0/1.2_release_notes.md
@@ -16,6 +16,9 @@ Containerized Foreman deployments (managed with `foremanctl`) now use Valkey 8 r
 
 As the container is based on Enterprise Linux 10, Enterprise Linux 10 system requirements apply, even if the host system itself is not running Enterprise Linux 10. In particular, the host needs to support the x86_64-v3 microarchitecture level. See [Exploring x86-64-v3 for Red Hat Enterprise Linux 10](https://developers.redhat.com/articles/2024/01/02/exploring-x86-64-v3-red-hat-enterprise-linux-10) for further information.
 
+#### Ubuntu 24.04 and Debian 13 support
+Foreman can now be installed on Ubuntu 24.04 (Noble) and Debian 13 (Trixie).
+
 #### Ruby 3.3 support
 Foreman can now run on Ruby 3.3.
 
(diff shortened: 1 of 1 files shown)


ekohl (commented):
I'd also like to deprecate Debian 12 and Ubuntu 22.04 to be removed in Foreman 5.1. Is that being tracked somewhere? I can help write something if needed.

ogajduse:
> I'd also like to deprecate Debian 12 and Ubuntu 22.04 to be removed in Foreman 5.1. Is that being tracked somewhere? I can help write something if needed.

I'm not sure if it's tracked, but Evgeni raised the same point in https://github.com/theforeman/foreman-documentation/pull/5307.

I've added a cross-link between the two PRs.

ekohl (approved):
I'd like to drop it with Foreman 5.1. Shall I create an RFC we can refer to? I'd be happy to track that in a follow up.

ogajduse:
> I'd like to drop it with Foreman 5.1. Shall I create an RFC we can refer to? I'd be happy to track that in a follow up.

That would be great if you could do that!
This makes me wonder if we should have an RFC for dropping EL9 too. How was that done in the past?

ekohl:
> This makes me wonder if we should have an RFC for dropping EL9 too.

I've touched on it in https://community.theforeman.org/t/upgrading-foreman-to-a-supported-rails-version/47402 but it deserves its own.

> How was that done in the past?

https://community.theforeman.org/t/deprecation-plans-for-foreman-on-el7-debian-10-and-ubuntu-18-04/25008 and https://community.theforeman.org/t/drop-support-for-running-on-el8-with-foreman-3-13/38083. The [3.11 release notes deprecations](https://theforeman.org/manuals/3.11/index.html#Deprecations) are a good example, though I now see that it wasn't actually a clickable link there.

And we also had posts to add support, like https://community.theforeman.org/t/path-to-ruby-3-0-3-1-el9-and-ubuntu-22-04/30295.

ekohl:
I've created https://community.theforeman.org/t/drop-support-for-running-foreman-on-debian-12-and-ubuntu-22-04/47614.

